PR #42, opened 2026-05-29
Remove OWASP ZAP Scan Request prompt from prompt-library

ravikod:
## Description

Removes the `OWASP ZAP Scan Request` prompt and softens hard counts (277 → 277+, 30 → 29+) across narrative copy so future single-prompt or single-article changes don't cascade across many docs. 
The two index files (prompt-library.md, learn.md) keep the soft count as well, while the per-row table in each remains the source of truth for the actual content. 

## Type of Change

- [ ] New plugin/power/tool
- [ ] Bug fix
- [X] Enhancement to existing content
- [ ] Documentation update
- [ ] Guardrail/CI update

## Team Folder

- [ ] `migrate/`
- [x] Other: `advisor/`

## Checklist


brandondahler (approved):
LGTM, but we should probably adjust the text to say "hundreds" and "dozens" instead of the exact counts.  That way the AI has a concept of how many without having to be accurate and creating the amount of churn we see here in this PR.

ravikod:
> LGTM, but we should probably adjust the text to say "hundreds" and "dozens" instead of the exact counts. That way the AI has a concept of how many without having to be accurate and creating the amount of churn we see here in this PR.

Agreed, thats the reason I introduced `+` suffix. But agreed on your suggestion as well. Next time.

ravikod:
> agree with Brandon's comment above, 'dozens' and 'hundreds' sound more natural in prose. Also, does a content removal warrant a version bump vs. a patch (1.2.1)?

True, "+" conveys this explicitly is what I felt. But will do that if it warrants a frequent update. And regarding version bump, may be but honestly it doesn't matter since we did not launch this yet.

ayn-builds (approved):

PR #93, opened 2026-07-01
Remove dormant heroku test suite (imports uninstalled fast-check)

icarthick:
## What

Removes the 7 tests under `migrate/plugins/migration-to-aws/tests/{property,integration}/heroku/`.

## Why

They `import fast-check`, which is **not a declared or installed dependency** anywhere in the repo. Running any of them fails immediately:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'fast-check'
# pass 0  # fail 1
```

So they don't run, aren't wired into CI or any mise task, and provide no actual coverage — while presenting a broken-looking suite that misleads authors into thinking the behavior is tested.

Nothing references these files (no workflow, mise task, or docs). **Git preserves them** for revival — they can be reintroduced later as real, dependency-declared, type-checked tests when the behavior-test setup is done properly.

No functional change to any skill.


icarthick:
Closing for now — will do the dormant-test removal as a follow-up after the frontmatter/validator PR (#91) merges, so it lands cleanly on its own.

PR #148, opened 2026-07-18
feat(heroku-to-aws): live discovery via the authenticated Heroku CLI

herosjourney:
## Summary

- Adds **live Heroku discovery via the user's authenticated Heroku CLI** as a first-class, recommended discovery source — read-only, consent-gated, config-var key names only.
- Heroku discovery previously required Terraform files with `heroku_*` resources, but most startups on Heroku manage everything through the dashboard and CLI. If `heroku login` works, "migrate my Heroku app to AWS" now just works.
- When Terraform is also present, live data is authoritative for current state and disagreements surface as **drift** (unmanaged resources, plan changes, config conflicts) — never silently resolved.

### Capture / parse split

- `discover-live-capture.md` (new): interactive capture pre-work in the main window — consent gate, **exact-command whitelist** (list/info commands only; `auth:token` explicitly banned), raw CLI output written to `live-capture/` with a manifest. Config vars captured as **key names only**; values never touch disk.
- `discover-live.md` (new fragment): parse-only — maps captures into `heroku-resource-inventory.json` (formations, add-ons with `monthly_price_usd`, domains, config-key lists, pipelines, spaces).

### Merge & drift

- `discover-assemble.md`: merge rules for live+Terraform runs — live wins for current state; Terraform-only resources marked `not_found_live`; live-only marked `unmanaged_by_terraform`; config conflicts recorded per field in `live_metadata.drift`; scaled-to-zero formations gap-filled from Terraform (not drift).

### Failure honesty

- Per-app capture failures (e.g. 403 on team apps) degrade confidence to `reduced` and continue — never a halt. Skipped captures (kafka plugin missing) are warnings.

## Relationship to other PRs

- #149 (GCP live discovery), #150 (live prices in Estimate), and #152 (what-if workshop) are stacked on this branch.

## Test plan

- [x] Fixtures: `fixtures/heroku-live-capture/` — canned CLI captures for a synthetic 3-app account (prod + staging + a 403 team app), deliberately-stale Terraform for the drift scenario
- [x] Fresh-agent replay, scenario A (live-only): inventory `discovery_sources: ["live"]`, 2 apps + 1 failed, no drift key
- [x] Fresh-agent replay, scenario B (live + Terraform): all merge rules exercised; `python3 fixtures/heroku-live-capture/check_expected_drift.py <run-dir>` → PASS
- [x] Secret hygiene asserted mechanically: no config-var values, no token material, no clustering fields anywhere in artifacts
[...]

PR #149, opened 2026-07-18
feat(gcp-to-aws): live discovery via the authenticated gcloud CLI (stacked on #148)

herosjourney:
## Summary

- Companion to #148: adds **live GCP discovery via the user's authenticated `gcloud` CLI** — read-only, consent-gated, env-var names only.
- Discovery previously required Terraform, app code, or billing exports; every GCP startup has an authenticated `gcloud`. The dead-end "No GCP sources detected" STOP becomes a consent prompt.
- Asset-search fast path (one Cloud Asset Inventory call) with per-service fallback when the CAI API is disabled; IaC drift merge when Terraform also exists.

## Changes

- `discover-live.md` (new): consent gate → capture (whitelisted `gcloud` list commands, projected fields only — no env values, no secret payloads, no full-metadata asset dumps) → parse into `gcp-resource-inventory.json` + simplified clustering (Step 3S) → edge inference from resolved config only.
- **Two capture modes:** Cloud Asset Inventory search (project-wide, one call; soft-ask to enable when disabled) or per-service fallback (19-row command table; per-row failures are warnings, never halts). Redis/Vertex walk regions deterministically via a whitelisted `gcloud compute regions list` (row 20) in fallback mode.
- **Drift merge** with Terraform: type-alias lock (legacy `google_cloud_run_service` vs live v2 merged as one resource), console-resize conflicts recorded per field, absence-of-evidence-is-not-drift for failed captures.
- Unmapped asset types are counted and disclosed (`live_metadata.unmapped_asset_types`), never guessed.

## Relationship to other PRs

- **Stacked on #148** (shares the live-discovery pattern; will rebase once it merges).
- #156 (grounded GCP baseline) is stacked on this branch.
- #101 (App Engine → EB): asked the author whether TF-less users are in scope — if yes, App Engine capture rows are a follow-up here.

## Test plan

- [x] Fixtures: `fixtures/gcp-live-capture/` — synthetic `acme-prod` project in per-service mode (CAI declined), covering Cloud Run/SQL/Redis/buckets/secrets + failed pubsub capture + stale Terraform
- [x] Fresh-agent replay of the drift scenario: `python3 fixtures/gcp-live-capture/check_expected_drift.py <run-dir>` → PASS (merge rules, type-alias lock, secret hygiene)
- [x] Real-project validation against the author's `sf-beach-app` GCP project (three review rounds folded back)
- [x] `mise run build` green

## Pre-existing tensions noted (not addressed here)

Live-only compute resources are not graviton-profiled in v1; SQL `dataDiskSizeGb` is provisioned (not actual) size — both documented as caveats in the capture spec.


herosjourney:
Follow-up commit: documented two deliberate design choices inline after reviewing the Cloud Asset Inventory docs — why the capture path avoids full-metadata dumps (env var values would land on disk) and why CAI relationship queries aren't used (the `relationships.*` query syntax works on the standard `search-all-resources` endpoint, but the underlying data requires SCC Premium/Enterprise or Gemini Cloud Assist — easy to be misled there). Possible future enhancement, out of scope here: `search-all-iam-policies` is standard-tier and could strengthen service-account `serves` edges.

herosjourney:
All review items addressed in 1198fb9 — point by point: 1. **Graviton regression — confirmed and fixed.** Root cause: SKILL.md was copied whole-file from a working tree based on a pre-#86 branch. Rebuilt from current main + the five feature edits, so nothing else regressed (verified: 4 graviton mentions restored, conditional-load row, file tree, Defaults bullet). 2. **Live offer widened.** Step 1d now offers live whenever no Terraform was found — billing-only and app-code-only workspaces get the primary offer; the STOP only fires when nothing at all will produce an artifact. 3. **BigQuery fallback added.** New row 17 (`bq ls`, dataset names/locations only, skip-with-warning if the CLI is absent); mapping table references it, so the specialist gate survives the CAI-disabled case. 4. **`--region=-` dropped** (managed Cloud Run lists all regions by default; note added). Redis region source now includes asset-search results for the CAI path. 5. **`--scope=projects/...` adopted.** Kept `--asset-types` unfiltered deliberately — it feeds `live_metadata.unmapped_asset_types`, the 'what else lives here' signal; documented inline. 6. **Rows 8/9/11/12 are now E rows** so the CAI path keeps edge-inference and name-inventory fidelity. 7. **Asserter committed** (`fixtures/gcp-live-capture/check_expected_drift.py`, stdlib-only). Validated by a fresh fresh-agent replay of scenario B against this revision: `HANDOFF_OK`, graviton_profile emitted from the restored path, asserter PASS. 8. **Disk-size caveat documented** (provisioned ≠ actual; follow-up for monitoring-metrics enrichment, along with graviton-profiling live-only compute). 9. **Agreed on process** — merge #148 first, will rebase; happy to bump the version once when both land. Full `mise run build` green on the new commit.

herosjourney:
**Real-world validation** (commit eb01165): ran the capture path against an actual startup GCP project and compared results to a prior IaC-based discovery of the same project. What held up: - CAI fast path returned 270 assets in one call; the >100 scale guard and extraction-script pattern engaged as designed - All six applicable enrichment rows succeeded, including the new `bq ls` row (1 dataset — matching the project's BigQuery specialist-gate state) - **The security contract passed with real stakes**: the project's Cloud Run env list includes names like `DB_PASSWORD` and `GEMINI_API_KEY` — captures contain names only, values verifiably absent. This is exactly what the documented rejection of `--content-type=resource` protects. - Live-only discovery found real click-ops: an unmanaged BQ dataset, an unmanaged bucket, and the default VPC — none in Terraform Three spec bugs the synthetic fixtures couldn't catch, fixed in this commit: 1. **Run v1/v2 false drift** — Terraform declaring `google_cloud_run_service` while live maps to `google_cloud_run_v2_service` flagged the SAME service as both missing and unmanaged. Step 6 now aliases v1/v2 type pairs (Run + Functions) on name match. 2. **Firewall rules unmapped** — `compute.googleapis.com/Firewall` added to the mapping table (network_path secondary, IaC-classification parity). 3. **SA naming** — `name` mandated as email local-part, never display name. No customer data committed; captures stayed local and were deleted.

herosjourney:
Residual nits addressed in 94e7875: (1) capture table reordered so row 16 (vertex) precedes 17 (bq) — mapping references unchanged; (2) SKILL.md prerequisites blurb aligned to Step 1d semantics (offer live whenever no Terraform, not only when no files); (3) took the optional one now — `bq ls` is an E row so CAI-success runs still verify BigQuery presence for the specialist gate. (4) Agreed as follow-up: wire `check_expected_drift.py` into the mise/CI test suite.

herosjourney:
Review round addressed in 84f270a — all seven items taken: 1. **Firewall fallback row** (18) — minimal projection, E-marked; firewalls no longer CAI-only (same fix class as the earlier bq gap). 2. **Connector edge corrected** — the annotation is a CONNECTOR id, not a network. New row 19 resolves `connector.network` via a read-only describe; if unresolvable, the connector id is recorded and NO edge is emitted. Good catch — this was a genuine error the real-project test missed (that service had no connector). 3. **v1/v2 alias locked by fixture** — orders-api is now declared as legacy `google_cloud_run_service` while live maps to v2; `expected-drift.json` asserts one merged resource (IaC address kept, `live_type` recorded, never a false not_found_live/unmanaged pair). Fresh-agent replay confirmed the alias rule executes as designed. 4. **Asserter hardened** — recursive env-value payload detection (name+value/valueFrom, raw env keys, env-name-as-key). 5. **Heroku parity** — asserter ported to #148 (fbbe273 there), validated against its own fresh replay. 6. **Unmapped types surfaced** — Discover summary now reports the count + top types. 7. **ADC fallback** — preflight probes `projects describe` before demanding `gcloud auth login`. Replay + asserter PASS + full build green on the new commit. One replay judgment call folded back in: SA merge matching now explicitly uses `config.account_id`.

herosjourney:
Good catch on the ADC probe ordering — it referenced $GCP_PROJECT one step before Step 0.3 set it. Fixed in c2dce5b by reordering: project resolution (a credential-free local config read + user confirmation) is now step 2, auth check + probe step 3, so the probe always has a resolved project. Took the reorder over the inline `$(gcloud config get-value project)` variant since the user should confirm the target project before any project-scoped call anyway.

herosjourney:
### Suggestion: soft-ask to enable CAI (user-driven), then fall back

For the startup audience this skill targets, enabling an API is usually low-friction — so when the Cloud Asset Inventory fast path fails because the API isn't enabled, consider offering a **user-driven** enable step before silently falling back to per-service rows.

**Proposed flow (agent never mutates):**
1. Detect CAI failure as API-not-enabled (vs. permission denied / other).
2. Ask once, with a short why + how-to, e.g.:

> Cloud Asset Inventory isn't enabled on this project. Enabling it gives a fuller inventory in one call (including resources outside the per-service list). I won't enable it for you.
>
> To turn it on:
> ```
> gcloud services enable cloudasset.googleapis.com --project="$GCP_PROJECT"
> ```
> (Console: APIs & Services → Enable APIs → "Cloud Asset API")
>
> [Y] I've enabled it — retry CAI / [N] Continue with per-service fallback

3. On **Y** → re-run the asset-search fast path.
4. On **N** / still failing → existing `method: "per_service"` path unchanged.

Keeps the read-only contract (user runs the enable), spends the ask only when CAI would help, and still fail-softs if they decline. Happy to take this as a follow-up if you'd rather not expand this PR's scope.

herosjourney:
Implemented in 666fa93.

When the CAI fast path fails with an API-not-enabled signal (`SERVICE_DISABLED` / "has not been used" / "is not enabled"), Discover now soft-asks once with:

```
gcloud services enable cloudasset.googleapis.com --project="$GCP_PROJECT"
```

plus the console path, then retries once on [Y] or falls through to per-service on [N]/still-failing. Permission-denied and other errors skip the ask. Agent never runs `services enable`. Manifest + `live_metadata` record `cai_enable_offered` / `cai_enable_accepted`; fixture models the declined path.

herosjourney:
Follow-up in the tip of the branch: soft-ask now matches the [official CAI setup](https://docs.cloud.google.com/asset-inventory/docs/view-assets) — enable `cloudasset.googleapis.com` **and** `roles/cloudasset.viewer` (with the docs link). Permission-denied (API already on, missing IAM) gets its own soft-ask instead of a silent per-service fallthrough.

herosjourney:
Review follow-up pushed (ee0ebb7): **deterministic region walk for the redis/vertex rows.** The previous "each region seen in asset-search results or rows 1–9" heuristic was fine in asset-search mode (project-wide coverage) but failed both ways in per-service fallback mode: an instance in a region with no other footprint was silently missed on custom-mode VPCs, while auto-mode VPC subnets inflated the seen set to all ~40 regions anyway. Now: new whitelisted **row 20** (`gcloud compute regions list --format=json(name)` — one cheap call, per-service mode only) enumerates the walk; asset-search mode walks the matching assets' locations; and if the regions list itself fails, the old heuristic is the fallback **with a capture warning naming the coverage limit** (surfaces in `live_metadata.capture_warnings`). Fixtures updated to model the walk (`regions.json` + empty-region redis captures = empty results, not errors), plus dropped the undocumented `--region=-` from the canned run.json command that contradicted the row-1 note. Build green. Related: asked on #101 whether TF-less App Engine users are in scope — if yes, App Engine capture rows would be a follow-up to this PR.

PR #150, opened 2026-07-19
feat(heroku-to-aws): live-captured prices for the Estimate baseline (stacked on #148)

herosjourney:
## Summary

- Wires #148's **live-captured add-on prices** (`config.monthly_price_usd`) into the Estimate phase's current-spend baseline: a new rung between billing data and the rate-card cache, with exact add-on math + published dyno rates and a mandatory not-an-invoice caveat.
- Fixes the bigger gap found on the way: the migrate-or-stay **comparison (Part 3) and migration cost considerations (Part 4) were gated on billing data specifically** — cache- or live-derived baselines produced projected AWS costs with no side-by-side at all. Both now run for ANY determined baseline, labeled with `baseline_source`.

**Motivation:** live-discovered runs (the common TF-less case after #148) had real plan prices sitting in the inventory that Estimate could not use without a billing export.

### `estimate-cost-engine.md`

- New baseline rung 2, `live_prices_plus_cache`: add-ons summed exactly from `monthly_price_usd`; dynos from `heroku-pricing-cache.md` × quantity; per-resource fallthrough for unpriced entries; gate requires ≥1 add-on with a live price (a dyno-only capture falls honestly to the cache rung).
- Part 3 comparison unlocked for any baseline source; new `cost_comparison.baseline_source`; derived-source per-app breakdown variant.
- Part 4 keyed off baseline presence (`baseline_available`/`baseline_source` replace `billing_data_available`); dual-run window priced at ~baseline/month with the derived caveat.
- Follow-ups landed on this branch from review: EB root volumes priced via a new `ebs` block in the shared pricing file (kills the permanent "known omission" warning); sign conventions documented and cross-referenced (`roi_analysis` difference vs `financial_summary` savings); pricing-hierarchy row 4 `estimated` documented.

### Schema

- `estimation-infra.schema.json` (+ vendored sync): `cached_stale` added to the `pricing_source.status` enum — both heroku and gcp specs already mandated it; the enum rejected it (caught by the validation replay).

## Relationship to other PRs

- **Stacked on #148** — only the commits after its tip are new here; will rebase once #148 merges.
- #156 is the GCP counterpart. #157 replaces this PR's pricing-hierarchy table with a canonical vendored copy (containing this PR's content) — take the pointer on conflict.

## Test plan

- [x] Fixtures: `fixtures/heroku-live-capture/seed-estimate/` (13-resource merged inventory, no billing profile, hand-assembled preferences + design per the specs)
- [x] `python3 fixtures/heroku-live-capture/check_expected_estimate.py <run-dir>` asserts the exact baseline math ($220 live add-ons + $132 cache dynos = **$352.00**), source label, caveat wording, unlocked comparison arithmetic, Property-16, no fabricated billing fields, secret hygiene
- [x] Fresh-agent Estimate replay: `HANDOFF_OK`, baseline derived exactly as specced, **asserter PASS on the first run**
- [x] `mise run build` green + `lint:frontmatter`

## Follow-up

GCP twin shipped as #156. Pre-existing observations recorded in PR comments (RDS Proxy formula-row placement, `service_count >= 9` complexity cliff).


herosjourney:
Review follow-up pushed (17cf2f1): tightened the rung-2 gate. "Live discovery ran" was ambiguous when the capture contained zero priced add-ons (e.g. a dyno-only app) — the rung would fire, price everything from the cache via per-resource fallthrough, and still label the baseline `live_prices_plus_cache` with "exact for add-ons" accuracy. Right arithmetic, wrong provenance. The gate is now **≥1 add-on resource carrying `config.monthly_price_usd`**; otherwise Part 1 falls to rung 3, whose `pricing_cache` label and ±5% accuracy describe that baseline honestly. Existing fixture/asserter unaffected (the seed scenario has four live-priced add-ons).

herosjourney:
Pushed 24092ac — three review-observation cleanups from the estimate replay, all in this PR's blast radius: 1. **EB root volumes are now priced.** New `ebs` block in the shared pricing file (gp3 $0.08/GB-mo + the 8 GB AL2023 EB root default, overridable by `aws_config.root_volume_gb`), vendored copies synced. The EB cost step's "known minor omission requiring verification" branch is gone — every Heroku EB estimate previously shipped that warning permanently. 2. **Sign conventions documented.** `roi_analysis.monthly_difference_*` (AWS − Heroku, negative = AWS cheaper) and `financial_summary.monthly_savings_*` (Heroku − AWS) are the same fact with opposite signs. Both specs now name their own convention, cross-reference the other, and require presentations to label direction in words — never a bare signed value. 3. **Pricing hierarchy row 4: `estimated`.** The schema and assembler already carried a `services_by_source.estimated` bucket with no hierarchy row defining when it's legitimate. Now defined: a rate this file's own formulas state verbatim may price a service (with a warning naming the rate); `unavailable` is reserved for no-cache + no-MCP + no-formula. No asserter impact: the committed estimate asserter checks the baseline math and structural invariants, not exact AWS-side totals, so the EB root-volume component changes no assertions.

PR #164, opened 2026-07-20
fix(llm-to-bedrock): sweep stale Sonnet 4 example model IDs to Sonnet 4.6

herosjourney:
## Summary

The llm2bedrock agent examples (code-rewriter, code-analyzer, report-generator) and the behavior-delta reference still used `us.anthropic.claude-sonnet-4-20250514-v1:0` — Claude Sonnet 4 (May 2025) — as the example rewrite target. The plugin's own model catalog marks that model **excluded, EOL Oct 14, 2026** (`ai-model-lifecycle.md`), and every recommendation elsewhere points at Sonnet 4.6. An agent pattern-matching the examples would produce rewrites targeting a model the plugin itself bans. This PR sweeps the examples to `claude-sonnet-4-6` and adds a guard note so example IDs can't silently drive rewrites again.

## Changes

- Swept `claude-sonnet-4-20250514-v1:0` → `claude-sonnet-4-6` in:
  - `agents/llm2bedrock-code-rewriter.md` (7 sites: Vercel AI SDK / LangChain / boto3 examples, env-var block)
  - `agents/llm2bedrock-code-analyzer.md` (target_models example)
  - `agents/llm2bedrock-report-generator.md` (model-id list example, pricing comment — Sonnet 4.6 has identical $3/$15 rates so the comment's math is unchanged, dual-ARN prose example)
  - `skills/llm-to-bedrock/references/helpers/behavior-delta-detection/references/openai-to-bedrock.md` (converse example)
  - `skills/llm-to-bedrock/scripts/test_iam_policy.py`, `test_validate_result.py` (example IDs in tests; inference-profile logic is ID-agnostic)
- **Intentionally kept:** the Sonnet 4 rows in `pricing-cache.md` and `ai-model-lifecycle.md` — those are catalog entries whose whole point is recording the model and its EOL/excluded status.
- New guard note in `llm2bedrock-code-rewriter.md` above the rewrite examples: example IDs are illustrative; authoritative IDs come from the migration plan (`aws_model_id`); verify plan IDs resolve via `aws bedrock list-foundation-models` (or the preflight converse-ping); if an ID no longer resolves, STOP and surface it rather than substituting a model. This mirrors the agent-toolkit-for-aws guidance on verifying model IDs at runtime.

## Relationship to other PRs

Independent — based on current `main` (post-#162). No overlap with open PRs (#141/#144/#149/#151/#155/#156/#161/#163).

## Test plan

- [x] Repo-wide sweep verified: zero `claude-sonnet-4-20250514` references outside the two catalog files
- [x] `pytest` on the two touched test scripts: 119 passed
- [x] Full `mise run build` green (160/160 checks)
- [ ] Fresh-agent rewrite replay against the fixture repo (examples-only change; flagging honestly that no replay was run)


herosjourney:
Applied the Cursor review (commit 89eda24) — all five points, with two findings that came out of verifying point 2:

**1. Regression lint — added.** `tools/model-id-lint.py` (stdlib) + `lint:model-ids` mise task in the `lint` group, so it runs in every build. Two pattern classes: the EOL Sonnet 4 ID outside the two catalog files, and fabricated dated forms (`sonnet-4-6-<date>` / `opus-4-8-<date>`). Explicit per-pattern allowlists; extend `BAD_PATTERNS` on future EOLs.

**2. The "dated Sonnet 4.6" IDs are fabricated, and one had a wrong price.** `claude-sonnet-4-6-20250514-v1:0` grafts Sonnet 4.6's name onto Sonnet 4's date stamp — that ID never existed. Swept in the three gcp design refs. Two things Cursor's list didn't surface:
- `bedrock_pricing.py`'s STATIC_FALLBACK carried the fabricated dated forms as **executable lookup keys**, plus a fabricated `claude-opus-4-8-20250610` at **$15/$75 per 1M — a 3x overstatement** (Opus 4.8 is $5/$25 per the plugin's own pricing cache). Table now uses real undated IDs with the correct rate; `_static_fallback` strips date stamps so dated inputs still resolve; two new tests guard the rate and the date-strip (130 pass in the touched files; the 3 pre-existing `test_preflight_bedrock.py` failures reproduce on untouched main — env-only, missing boto3).
- `resolve-bedrock-model-id.md` keeps its dated example **on purpose** (repairing broken IDs is that helper's job) — now labeled intentionally-invalid, and allowlisted in the lint.

Swept in this PR rather than a separate one: the lint has to be green to merge, so lint + sweep are one atomic change.

**3. Analyzer guard — added** as a `_comment_target_models` sibling key in the example JSON.

**4. ID-form note — added** to the rewriter guard: geo-prefixed inference-profile form preferred in rewritten code; bare foundation-model IDs belong in pricing tables (they often reject on-demand invocation — see the IAM known-fix).

**5. Replay** — left unchecked, per the review.

Full `mise run build` green (160/160) including the new lint.

migrate/plugins/migration-to-aws/agents/llm2bedrock-code-analyzer.md
```diff
@@ -352,7 +352,8 @@ Return ONE flat object: the typed fields and `summary` are all top-level sibling
   "ai_framework_version": "langchain==0.1.14",
   "source_provider": "openai",
   "source_models": ["gpt-4o"],
-  "target_models": ["gpt-4o -> us.anthropic.claude-sonnet-4-20250514-v1:0"],
+  "target_models": ["gpt-4o -> us.anthropic.claude-sonnet-4-6"],
+  "_comment_target_models": "example only — the real mapping comes from the migration plan, never from this illustration",
```
leon1418: Currently Bedrock supports GPT-5.6, and Sonnet is positioned in the same tier as GPT-5.6 Terra. So, should we recommend migrating to a model from the same providor?

herosjourney: Good call — GPT-5.6 (Sol/Terra/Luna) went GA on Bedrock July 13, and Terra does sit in the same tier as Sonnet 4.6. Addressed in d06b405:

- `ai-openai-to-bedrock.md` (the design ref where target recommendations actually originate) now leads with a same-provider-first rule: OpenAI-sourced apps map to the corresponding GPT-5.6 tier on Bedrock by default, with the cross-provider option (Claude/Nova) presented alongside. Added the GPT-5.6 mapping table with Bedrock model IDs and post-July-30 pricing (`openai.gpt-5.6-sol` $5.50/$33, `-terra` $2.20/$13.20, `-luna` $0.22/$1.32 per 1M, ≤272K ctx).
- One important constraint documented as a hard gate: GPT-5.6 on Bedrock is **Mantle-only** (OpenAI Responses API path on `bedrock-mantle`) — no Converse/bedrock-runtime, no geo/global inference profiles, no Guardrails/KB. So it's the recommended path when the app keeps the OpenAI SDK surface (env-var swap), but apps being rewritten onto boto3 `converse()` still need a Converse-capable family.
- `pricing-cache.md` gained the GPT-5.6 catalog rows + breakpoint pricing table; the analyzer example (this line) now notes plans may map same-provider and that GPT-5.6 targets are echoed bare (no `us.` prefix).

This example line stays gpt-4o → Sonnet because the surrounding example shows a LangChain → `ChatBedrockConverse` rewrite, which can't target a Mantle-only model — but the mapping comment now makes the same-provider path explicit. Rewriter support for Mantle-target rewrites (keeping the OpenAI SDK, swapping base URL) is probably worth its own PR if we want Track 2 to handle it end-to-end.

leon1418:
Pushed the merge with main plus the `drift:check` unblock. This one had substance beyond the mechanical sync — three commits:

**[f3c80ee] Merge with main (#165/#167 landed).** Six conflicts resolved: the three design-ref examples take main's Sonnet 5 direction, the pricing table takes main's Sonnet 5 defaults, and the `display_name_guess` assertions from both sides are kept (dated Haiku + undated Sonnet 5 coverage).

**[b46b556] Alignment with this PR's own lint — including a correction to merged main content.** Running your `tools/model-id-lint.py` against the merged tree flagged 4 hits, and it was right on all of them:
- The two dated `claude-sonnet-4-6-20250514-v1:0` rows #167 added to `STATIC_FALLBACK` are fabricated hybrids per the lint's ground truth — removed. Dated queries still resolve via the date-strip/bidirectional match, so behavior is unchanged.
- The Opus rate-lock test (from #218's strengthening) asserted dated literals; it keeps its four-shape behavioral coverage but now builds the dated forms by concatenation, borrowing this branch's own technique from `test_static_fallback_dated_id_matches_undated_key`.
- Your intentionally-invalid example in `resolve-bedrock-model-id.md` survives as the lint's single allowlisted exception — it was correct as written.

**[1427242] Mechanical twin sync** (same as #165/#167/#180–#186): 8 advisor copies re-synced + dprint fmt.

Verified: `model-id lint: OK`, llm-to-bedrock 195/195 on both plugin copies, drift 256 identical, vendored-shared, dprint all green; the advisor tree greps clean of fabricated dated IDs.

One note for a follow-up: the lint currently guards `migrate/` only (its PLUGIN root is the migrate plugin). The advisor twins are clean by sync-construction today, but wiring the same lint over `advisor/` would make that permanent.

leon1418 (approved):

az-zhu (approved):

PR #173, opened 2026-07-22
feat(gcp-to-aws): decision outcomes (go/conditional/defer/stay) + executive trust panel

herosjourney:
## Summary

**The problem, in plain English:** This plugin exists to help a startup decide whether moving to AWS is a good idea — but its recommendation vocabulary can only say *how* to migrate, never *whether*. "Stay on GCP" is only reachable when AWS is more expensive AND cost is the customer's sole metric, there is no way to say "we don't have enough evidence to decide yet," and a complex stack silently becomes a migration recommendation (just a phased one). Meanwhile the report's verdict never tells the reader what it rests on: which inputs were measured from their billing, which were assumed by default, and how stale the pricing is. A reader can't judge how much to trust the recommendation, and the recommendation itself can't express honest uncertainty.

Changes (fully additive — `recommendation.path` and its consumers are untouched):

- **`recommendation.outcome`** (`estimate-infra.md` Part 7, `schema-estimate-infra.md`): new field beside `path` with vocabulary `go | conditional_go | defer_for_evidence | stay`. `path` = execution shape; `outcome` = the decision. Derivation is codified in a two-tier trigger table, with an explicit rarity guard: **defer is expected to be rare** — AWS almost always has the services and the AWS-side estimate can almost always be produced, so when in doubt the outcome is `conditional_go` with named conditions.
  - *Hard triggers* (any one forces `defer_for_evidence` — the three genuinely undecidable cases): specialist-gated services (e.g. BigQuery) that are a **material share of spend** and must cut over in the **same window** as app infra (the plugin refuses by policy to design/price that portion); **GovCloud-class** compliance ambiguity (FedRAMP flips regions, service catalog, and pricing wholesale); cost-savings-only motivation with **zero** spend signal (no billing AND Q3 declined).
  - *Soft triggers* (become `conditions[]` on a `conditional_go`, never force defer): missing billing when a Q3 spend band exists; non-GovCloud compliance unknowns (HIPAA explicitly: the decision survives, only controls/architecture change); CUD lock-in without exit plan; unresolved multi-region HA conflict; defaulted availability; stale pricing beyond the accuracy band. The tiering matters: an ordinary run trips 2–3 soft triggers, and if any one forced defer, defer would become the meaningless default outcome.
  - *Defer rendering rule:* a defer must lead with what IS established ("AWS can host this stack; AWS-side estimate $X–$Y/mo") before naming the single missing piece of evidence — defer means "one named thing blocks a responsible verdict," never "we can't tell you anything."
- **`stay` de-biased:** reachable on any decisive factor, no longer gated on "cost is the sole metric." Complexity alone selects `path: migrate_phased` but never moves `outcome`.
- **Decision basis + flip conditions:** `decision_basis` (measured / assumed / unknown, derived from existing `chosen_by` provenance) and `would_flip_if[]` (1–3 changes most likely to alter the outcome).
- **Report Section 0 — typography-first verdict:** `outcome_label` renders as the section's headline statement in display type, with one labeled metadata line ("Execution shape: Phased Migration · Complexity: Complex") and a conditions checklist beneath. No colored pill badges — structure carries the information and meaning never depends on color alone (per frontend-design guidance). Falls back to `path_label` as the headline for pre-extension artifacts. A one-line confidence pointer links to the assumptions panel.
- **"What This Assessment Rests On" panel** (`generate-artifacts-report.md`, new required `<section id="exec-assumptions">`, plain heading — no literal numbered heading, per the validator's readability rule): table of every defaulted constraint with its `design_consequence` verbatim, confidence gloss with the measured/assumed/unknown basis, and pricing provenance. **Placed at the end of the executive summary** (after Top Risks, before the appendices), where assumption/validation sections conventionally sit — the verdict links to it rather than front-loading it. All read from existing artifact fields — nothing invented.
- **Validator + fixtures:** `exec-assumptions` added to required section IDs in `validate-migration-report.py`; reference fixture gains the panel (end-of-summary position), the headline verdict treatment, and TOC entry; `estimation-infra-reference.json` models the v2 fields; test HTML scaffolds updated.
- **Plugin README:** Estimate phase entry documents the outcome vocabulary and defer's rarity; the migration-report section lists the assumptions panel and its placement.

## Why

Usability review conclusion, agreed across two review rounds: the migrate-biased vocabulary and the invisible assumption base are the two biggest gaps for the plugin's core question ("should we move?"). Additive rollout (option B) was chosen deliberately — consumers that only read `path` keep working; `path` deprecation is a follow-up.

## Test plan

- [x] `python3 -m pytest tests/` — 81 passed (includes reference-fixture-passes and stub-fails asserts against the extended validator)
- [x] `mise run lint:md` — 0 errors; `mise run shared:check` — OK
- [x] Reference fixture passes the extended validator with the new required section
- [x] No literal "Section N" heading introduced (validator readability rule would catch it — fixture passes)
- [ ] Manual: run with defaulted Q6 produces `conditional_go` with the availability condition; run with same-window material BigQuery cutover produces `defer_for_evidence` that opens with the AWS-side estimate

## Out of scope

- Replacing or removing `recommendation.path` (follow-up deprecation PR)
- Workshop compare outcome column (next PR consumes `outcome`)
- Heroku/Vercel recommend engines
- No model-selection or Bedrock mapping changes
- Validator does not enforce section *order* (presence + TOC only) — mechanical order check is a possible follow-up

## Revision history

- `262307b` — narrowed defer triggers to the three genuinely undecidable cases; rarity guard; defer rendering rule (review feedback: original triggers were too eager)
- `da128e1` — verdict de-badged to typography-first headline; assumptions panel moved to end of executive summary; README updated


herosjourney:
Amended in 262307b after review feedback: `defer_for_evidence` was too eager as originally scoped — AWS almost always has the services and the AWS-side estimate can almost always be produced, so most missing-evidence cases are honestly a `conditional_go`, not a defer. Changes: - **Retiered triggers.** Hard triggers narrowed to the three genuinely undecidable cases: (1) same-window cutover of specialist-gated services that are a material share of spend (the plugin refuses by policy to design/price the analytics target, so a verdict on the designed slice would misrepresent the migration); (2) GovCloud-class compliance ambiguity (FedRAMP flips regions, service catalog, and pricing wholesale — a commercial-region verdict could be flat-out wrong); (3) cost-savings-only motivation with zero spend signal (no billing AND Q3 declined). Former hard triggers 'no billing baseline' and 'general regulated-domain compliance unknown' are demoted to soft triggers 4–5 (HIPAA explicitly noted: the decision survives, only controls/architecture change). - **Rarity guard.** Derivation rules now open with: defer is expected to be rare; when in doubt, prefer conditional_go with named conditions. - **Defer rendering rule** (estimate summary + report Section 0): a defer must lead with what IS established — 'AWS can host this stack; AWS-side estimate $X–$Y/mo' — before naming the single missing piece of evidence. Defer means 'one named thing blocks a responsible verdict,' never 'we can't tell you anything.' Lint + 81 tests still green.

herosjourney:
Second amendment in da128e1, addressing three review points: 1. **Assumptions panel moved to the end of the executive summary.** `exec-assumptions` now sits after Top Risks, immediately before the appendices — where assumption/validation sections conventionally belong in a migration report. The decision summary keeps only a one-line confidence pointer that links to it (`Confidence: Medium — full basis in What This Assessment Rests On`), so the verdict still carries minimal provenance without front-loading a full assumptions table. Spec, fixture section order, and TOC all updated. 2. **Verdict de-badged (frontend-design skill compliance).** Section 0 no longer renders the outcome as a row of colored pill badges — that treatment made color the meaning-carrier and used badges as decoration rather than structure. The verdict is now a typography-first headline (`outcome_label` in display type) with one labeled metadata line ('Execution shape: Phased Migration · Complexity: Complex'). The hierarchy encodes the actual relationship: the decision is the thesis; how and how-hard are attributes. Words carry the verdict; color is at most an accent — meaning never depends on it (also the accessible choice). 3. **Plugin README updated.** Estimate phase now documents the outcome vocabulary (including that defer is rare by design), and the migration-report section lists the What This Assessment Rests On panel and its placement. Lint + 81 tests green; report validator only asserts section presence/TOC integrity, so the reorder passes without validator changes.

PR #174, opened 2026-07-22
feat(gcp-to-aws): AI clarify detect-confirm cards + AI-only mini assumption sheet

herosjourney:
## Summary

**The problem, in plain English:** The plugin's code scanner usually already knows which AI framework and models a startup uses — but the moment detection is anything less than fully certain, the questions fall back to cold walls of options: a 7-option framework taxonomy and a 19-option model list (Gemini 3.5 Flash Thinking vs Gemini 3.1 Flash-Lite vs GPT-5.4 Nano…). A founder who just watched the plugin correctly find `langchain` in their requirements.txt is then asked to pick their framework from scratch, as if the scan never happened. And in the AI-only flow there's no assumption sheet at all — the full Clarify wizard shows users what was detected before asking anything, but AI-only users go straight into question batches.

Changes:

- **Q14 partial-detection confirm card (`clarify-ai.md`):** when detection produced signals but not full resolution, present "I detected LangChain (requirements.txt) — that's right / edit / just LangChain" instead of the full taxonomy. The full option list only appears cold when detection found nothing.
- **Q19 confirm card + family-first cold path (`clarify-ai.md`):** detected-but-uncertain models are offered as a short pick-list of the actual candidates. When nothing was detected, ask provider family first (Gemini / GPT / o-series / other), then show only the matching subset — never the 19-option list in one message. The full list remains the interpretation catalog.
- **AI-only mini assumption sheet (`clarify-ai-only.md`, new Step 1.75):** detected values (framework, model, capabilities) and to-be-assumed defaults presented as a compact confirm-or-edit sheet before Batch 1, mirroring the full wizard's sheet-before-questions contract. Sheet-resolved questions are skipped in the batches.
- **Q27 separated (`clarify-ai.md`):** Activate-credits question asked as its own short follow-up after the technical essentials ("Last one, and it's about money rather than tech"), never mixed into the technical batch. Existing anti-inference rules (no eligibility from spend, no silent default) untouched.

Multi-workload confirmation already had accept/edit/drop actions — unchanged.

## Why

Usability review: detect-confirm is the natural completion of the extraction machinery this skill already has (`chosen_by: "extracted"`, confidence scores, auto-skip rules). The only missing piece was what to do at partial confidence, which is the common case.

## Test plan

- [x] Q14/Q19 interpretation rules, override hierarchy, and Bedrock baseline tables untouched (diff-verified — wording additions only above the option lists)
- [x] AI-only fast path (Step 1.5) unaffected; mini sheet only runs on the standard/complex path
- [ ] Manual: profile with `models[0].confidence = 0.6` produces a two-candidate confirm card, not the full catalog

## Out of scope

- Bedrock mapping tables / design-ai changes (explicitly none — model recommendations unchanged)
- Full Clarify wizard batching (PR #170)
- Multi-workload merge/split actions (v2 per existing limitation note)


herosjourney:
Pushed 731cdba (after merging current main, which now includes #170): the batch-composition line #170 added to clarify.md still listed Q27 inside the AI/agentic essentials batch, contradicting this PR's Q27 separate-follow-up rule in clarify-ai.md. The batch list now excludes Q27 and points at the presentation rule, so an agent reading both files gets one consistent instruction. Lint green (855 files).

PR #176, opened 2026-07-22
feat(gcp-to-aws): benefit-first workshop offer, outcome flips in compare, feedback decision check

herosjourney:
## Summary

**The problem, in plain English:** The what-if workshop is the plugin's best decision-support feature — reprice your whole migration under different choices without redoing discovery — but it's offered in insider jargon ("change region, availability/HA, compute (EKS/ECS), or Graviton preference") that gives a founder no reason to say yes. Once inside, the compare table shows dollar deltas but never answers the question the user actually brought: does any of these scenarios change whether I should migrate? And at the very end of the flow, the feedback phase asks users to fill in an external survey without ever asking the one thing that measures whether the plugin worked: could you make your migrate/stay decision from this?

Changes:

- **Benefit-first workshop offer (`estimate.md`):** reworded in plain terms with concrete examples ("cheaper single-AZ database for staging", "ARM-based (Graviton) compute"), plus a data-justified scenario hint — when a material assumption was defaulted (most commonly Multi-AZ availability, a ~2x database cost factor), suggest the one scenario that would bound it. A/B structure and all state-machine transitions untouched.
- **Outcome flip in compare (canonical + skill):** the canonical `skills/shared/workshop/workshop-invariants.md` gains an optional nullable `estimation_summary.recommendation_outcome` field (copied from `recommendation.outcome` when the inner estimate wrote v2 decision fields — see #173); vendored trees re-synced via `shared:sync`, `shared:check` green. The GCP `workshop-compare.md` adds an Outcome column (omitted when no scenario carries it) and a required flip callout: "single-az scenario flips conditional_go → go: the availability condition no longer applies." `schema-workshop-scenarios.md` documents the field.
- **Feedback decision check (`feedback.md`, new Step 2.5):** three optional chat questions before the Pulse link (could you decide? / which way? / what evidence was missing?). Explicitly NOT written to `trace.json` or `feedback.json` — the trace/Pulse contract is byte-identical to before; skipping never blocks the phase.

## Why

Closes the decision loop opened by #173: outcomes now travel into scenario comparison (where users test what would change their mind) and the conversation ends by checking whether the core question got answered. Canonical-vendored discipline followed deliberately — the repo's history includes a drift bug from hand-editing vendored copies, so the invariants edit went to canonical + `shared:sync` only.

## Test plan

- [x] `mise run shared:check` — OK across all 4 vendored trees after sync (no hand-edited vendored files; diff shows canonical + 3 synced copies byte-identical)
- [x] Workshop A/B state machine, scenario-store rules, and inner-run contract untouched (diff-verified: offer copy and compare rendering only)
- [x] Trace schema untouched; decision-check answers never persisted
- [ ] Manual: scenario set where one variant removes the only soft-trigger condition renders the Outcome column and the flip callout

## Out of scope

- Pulse survey form changes (external)
- Heroku/Vercel compare views (field is optional/nullable; they can adopt it later)
- Outcome derivation logic (lives in #173)

Depends on #173 for `recommendation.outcome` to exist; without it the Outcome column simply never renders (field is optional), so merge order is preferred but not strictly required.


herosjourney:
Pushed bb8d557 (rebased on the branch's upstream-merge tip): workshop-refresh.md's Snapshot step never instructed copying `recommendation.outcome` into the manifest — the field was documented in the schema and canonical invariants, but the file agents actually execute during a reprice omitted it, so the compare view's Outcome column and flip callout would silently never render. Snapshot step 6 now performs the copy (omit/null when the inner estimate lacks v2 fields). Found by merging all seven series branches and walking the combined tree.

PR #180, opened 2026-07-23
feat(heroku-to-aws): clarify extraction + assumption sheet, decision outcomes, breadcrumbs (gcp parity)

herosjourney:
## Summary

**The problem, in plain English:** The same plugin now treats a GCP user and a Heroku user very differently. A GCP user gets shown what their Terraform already answered before being asked anything, sees every assumption with its cost consequence, gets a verdict that can honestly say "go, once you confirm X" — and always knows which phase they're in. A Heroku user gets none of that: the interview asks about database HA even though the Postgres plan tier in their own inventory already says whether an HA follower exists, applies defaults silently, can only recommend "stay on Heroku" if cost is literally the customer's only concern, and hands back a verdict with no stated basis. This PR back-ports the gcp-to-aws usability improvements (merged as #170–#176) to heroku-to-aws, adapted to its flat preferences schema and fragment DSL — it does not blindly copy the GCP text.

Changes (all under `skills/heroku-to-aws/`; Vercel and GCP untouched, no canonical `skills/shared/` edits):

- **Extraction rules (`clarify-interview.md`):** answer from `heroku-resource-inventory.json` before asking — Q1 region from Private Space region (detected) or Common Runtime `us`/`eu` (proposed suggestion only), Q6 database HA from Postgres plan tier (`standard-*` = no follower, `premium-*`/`private-*`/`shield-*` = HA), Q7 Redis HA from plan tier, Q12b containerization from app stack. With an honesty rule the GCP flow didn't need: **plan tier says what the customer HAS, not what they NEED** — tier-derived rows show their source so users can correct target posture vs current posture.
- **Assumption Sheet gate (`clarify-interview.md`):** detected + assumed tables with consequence lines, numbered menu actions (Confirm all / Change a setting / Ask about X / Ask everything), plain-language corrections accepted, structured-question-tool note with chat fallback. Batches then contain only unresolved questions, capped at 4 per turn with progress lines.
[...]

PR #30, opened 2026-05-27
fix: AI-only migration quality — Anthropic routing, OpenRouter env detection, discover preview, Clarify auto-skip

herosjourney:
Four targeted fixes to improve accuracy and UX for AI-only migrations, plus follow-up fixes from review feedback. No infra migration behavior changes.

## Problem

1. Anthropic SDK users get wrong guidance — `ai_source: "other"` routes them to the SageMaker rubric
2. OpenRouter users miscategorized — env-var detection missing
3. AI-only Discover ends dry — no preview table for AI-only runs
4. Clarify asks questions already answered by discovery
5. Hybrid path Q19/Q20 missing extraction skip logic (asymmetry with AI-only Q5/Q6)
6. Simple AI-only migrations face full 10-question Clarify with no fast-path

### `discover-app-code.md`
- `ai_source: "anthropic"` added as distinct value for Anthropic SDK detection
- OpenRouter env-var detection: scans `.env`, `docker-compose.yml`, CI configs, shell scripts for `OPENAI_BASE_URL`, `OPENROUTER_API_KEY`, `OR_API_KEY`, `LITELLM_PROXY_BASE_URL`, `LITELLM_API_KEY` — does NOT read secret values

### `design-ai.md`
- Routes `ai_source: "anthropic"` → `ai-anthropic-to-bedrock.md` (not `ai-openai-to-bedrock.md`)
- Adds Anthropic SDK → Bedrock Converse API row to migration pattern table

### `references/phases/discover/discover-preview.md`
- AI-only route (Steps 1A–6A): per-token price comparison, `ai_complexity_signal`, "Your AI migration at a glance" preview table
- Restored IF/OR/AND pseudo-code indentation (cosmetic regression fix)

### `clarify-ai-only.md`
- Auto-skip Q1 (framework), Q5 (model), Q6 (capabilities) when discovery resolved them
- Added Claude/Anthropic as Q5 option
- Added Step 1.5 fast-path: `ai_complexity_signal == "likely_simple"` → 3 questions only (Q2, Q3, Q4)
- Removed duplicate Claude row in Q5 table

### `clarify-ai.md`
- Added extraction skip to Q19 (current model) — mirrors Q5 in `clarify-ai-only.md`
- Added extraction skip to Q20 (input types) — mirrors Q6 in `clarify-ai-only.md`

### `references/design-refs/ai-anthropic-to-bedrock.md` *(new file)*
- Anthropic SDK → Bedrock Converse API client swap reference
[...]

PR #181, opened 2026-07-23
feat(gcp-to-aws): extraction-consistency gates + canonical disk_size_gb schema

herosjourney:
## Summary

**The problem, in plain English:** A real run of Clarify asked all 20 questions — including database size (Q13b) and current AI model (Q19) — even though the user's Terraform said `disk_size = 10` and app-code discovery had already identified the model. The wrong Q13b answer then propagated into the data-migration script (pg_dump vs pgcopydb selection). The wizard spec already says extract-first, but nothing *fails* when an agent asks a question the inventory had already answered, and nothing downstream cross-checks the answer against the Terraform ground truth. On top of that, the ground truth itself was being misread: Terraform `disk_size` is **allocated capacity** (Cloud SQL minimum: 10 GB), not actual data — so even perfect extraction routes a nearly-empty dev database to pgcopydb when pg_dump would finish in seconds.

Changes:

- **Extraction-consistency gate checks (`clarify.md`, checks 7–8):** the fail-closed Clarify handoff gate now emits `GATE_FAIL | reason=extractable_question_asked_Q13b` when Q13b appears in `metadata.questions_asked` despite an unambiguous inventory disk size, and the same for Q19 when `models[0].model_id` was detected at confidence ≥ 0.8. Both checks exempt `clarify_mode: "full"` (user explicitly opted into the long path) and `chosen_by: "user"` (sheet-row conversion via "ask me about X" or a correction). The #174 detect-confirm cards are unaffected — they only fire below the confidence bar.
- **Canonical Cloud SQL config fields (`discover-iac.md`, `schema-discover-iac.md`):** Discover now normalizes `google_sql_database_instance` to `config.disk_size_gb` (from Terraform `settings.disk_size`), `config.availability_type`, `config.tier`, and `config.database_version`, and the schema documents them with an example — previously the schema never mentioned these fields and three path variants circulated (`config.disk_size`, `config.disk_size_gb`, `gcp_config.disk_size_gb`). Clarify and `clarify-database.md` now name the canonical path and keep the legacy variants as accepted fallbacks for older inventories. Also fixes the `gcp_config` vs `config` naming inconsistency in the Step 0 redaction rule.
- **Allocated ≠ actual (`clarify.md`, `clarify-database.md`, `generate-artifacts-scripts.md`):** extracted disk size is now labeled as an upper bound everywhere — the Assumption Sheet row reads "up to N GB allocated (actual data may be less)", and downward user corrections are accepted without pushback (data below allocation is the normal case). The extracted band remains safe for tool selection because it never underestimates.
- **Generate cross-check + runtime measurement (`generate-artifacts-scripts.md`):** before selecting pg_dump / pgcopydb / DMS for `02-migrate-data.sh`, compare `design_constraints.db_size.value` against the inventory band — asymmetrically: the allocated band only overrides a *larger* user answer (data can't exceed its allocation); a smaller user answer stands. And because neither value measures actual data, the generated script now always measures it (`pg_database_size`) before migrating: the pg_dump variant hard-stops when actual ≥ 10 GB (undersized tooling is the dangerous direction), the pgcopydb variant advises regenerating when actual < 10 GB (oversized is merely inconvenient).
- **`SKILL.md` Philosophy:** one-line "Extract before ask" principle at the top, so the posture isn't only discoverable inside clarify.md Step 2.

## Why

Post-mortem of a real migration run (SF beach stack: Cloud Run + Cloud SQL + Gemini). The agent fell into ask-everything mode; the spec's extract-first instructions existed but were advisory at the boundary that matters. This PR makes the two highest-signal extractable questions machine-checkable at the handoff gate (same fail-closed pattern as checks 1–6), pins down the schema ambiguity that made defensive triple-path reads necessary, corrects the allocated-vs-actual semantics that would produce wrong tooling even under perfect extraction, and adds one defense-in-depth check where the bad answer did concrete damage (migration tooling selection).

## Test plan

- [x] `mise run build` — green (lint, fmt:check, gitleaks, checkov, grype)
- [x] `mise run lint:md` — 0 errors across 804 files
- [x] `mise run lint:frontmatter` — OK
- [x] Gate checks 1–6, wizard flow, sheet contract, and detect-confirm cards untouched (diff-verified: additive checks 7–8 only)
- [x] Legacy inventory compatibility: Clarify still accepts `config.disk_size` and `gcp_config.disk_size_gb` as fallbacks
- [x] Cross-check asymmetry: user band smaller than allocated band → user's answer stands, no override
- [ ] Manual: run Clarify against a Terraform stack with `disk_size = 10` and confirm Q13b lands on the Assumption Sheet as "up to 10 GB allocated", and that a preferences.json with Q13b in `questions_asked` fails the gate
- [ ] Manual: generated pg_dump script exits 1 against a database whose actual size ≥ 10 GB

## Out of scope

- A separate `clarify-extractions.json` manifest artifact — the Assumption Sheet + per-constraint `source` provenance + `preferences-draft.json` already cover "show your work"; a parallel artifact would add drift surface
- Extraction-consistency checks for other DETECTED-eligible questions (Q1/Q6/Q12/Q13/Q14/Q20) — start with the two that caused observed damage; extend later if the pattern earns it
- MySQL runtime size measurement (`pg_database_size` is PostgreSQL; the MySQL path keeps band-based selection)
- heroku-to-aws / other skills

## Type of Change

## Team Folder

- [x] `migrate/`


herosjourney:
**Context: how this relates to the merged extraction lineage (#15 → #57 → #30 → #127 → #174)**

A reasonable question is what this adds given that auto-extraction already exists. Short answer: those PRs tell the agent what to do; this one makes Clarify fail closed when the agent didn't do it.

Evidence this is needed: a real migration run on 2026-07-23 — with #57, #30, #127, and #174 all merged — still ran ask-everything mode (20 questions in `questions_asked`), recorded a wrong Q13b answer despite Terraform declaring `disk_size = 10`, and completed Clarify cleanly. Every existing gate checks the input side (sheet before questions, well-formed preferences.json); none inspects the output for extraction consistency. A run that skips Step 2 entirely passes all of them.

What each merged PR does vs. what this PR adds:

- **#57 / #30 / #127 / #174** built the extraction rules, the assumption-sheet wizard, `source` provenance, and detect-confirm cards — all mechanisms that make compliance *more likely*.
- **This PR** adds the first checks that compliance *happened*: gate 7/8 fail Clarify when Q13b or Q19 was asked despite an unambiguous inventory answer (exempting full flow and user-converted rows, so #174's confirm cards are unaffected).
- **Schema fix**: #15 read `gcp_config.disk_size_gb`, #57 tested against `disk_size`, and `schema-discover-iac.md` documented neither — Clarify has been defensively reading three path variants ever since. This PR gives Discover a canonical `config.disk_size_gb` contract and keeps the legacy paths as fallbacks.
- **Generate cross-check**: no prior PR defends downstream of Clarify. The inventory-vs-`db_size` comparison at tool-selection time would have caught the July run's error even after Clarify completed with it.

Same fail-closed pattern as existing gate checks 1–6; purely additive.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/phases/generate/generate-artifacts-scripts.md
```diff
@@ -83,6 +83,35 @@ Based on database and storage resources in `aws-design.json`:
 
 Read `preferences.json` → `design_constraints.db_size.value` to select the migration tool:
 
+**Inventory cross-check (before selecting the tool):** If `gcp-resource-inventory.json` has a `google_sql_database_instance` with a disk size (`config.disk_size_gb` or legacy variants), map it to the Q13b band and compare against `db_size.value`. Terraform disk size is **allocated capacity — an upper bound on actual data**, so the comparison is asymmetric:
+
+- **`db_size.value` band is LARGER than the allocated band** — data cannot exceed its allocation (barring `disk_autoresize` growth since the Terraform was written). Warn and use the allocated band:
+
+  > **Database size inconsistency:** Clarify recorded `db_size: [value]` but Terraform only allocates `disk_size_gb: [N]` ([band]). Using the allocated band for tool selection — say "use my answer instead" if the disk has grown since this Terraform was written.
```
leon1418: [🤖 AI review 🤖]

Nit: The parenthetical "(barring `disk_autoresize` growth since the Terraform was written)" is correct but introduces a concept the reader has no tooling to resolve — Discover doesn't currently extract `disk_autoresize` into the inventory, so there's no way for the agent to know whether autoresize is enabled. The caveat is fine as documentation (explains *why* the user might legitimately push back), but if `disk_autoresize` is commonly set, a future follow-up could extract it and suppress the warning when autoresize=true AND no user band has been recorded.

No action needed — just flagging the implied extension point.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/phases/clarify/clarify.md
```diff
@@ -836,6 +836,8 @@ Load `shared/handoff-gates.md`. **Re-read from disk** before checking.
...
+8. **Extraction consistency — Q19** _(skip when `metadata.clarify_mode` is `"full"`)_: If `ai-workload-profile.json` exists with `models[0].model_id` set at confidence ≥ 0.8 → `Q19` must NOT appear in `metadata.questions_asked` unless `ai_constraints.ai_model_baseline.chosen_by` is `"user"`. Otherwise: `GATE_FAIL | phase=clarify | field=metadata.questions_asked | reason=extractable_question_asked_Q19`. (Detect-confirm cards for sub-threshold or tied detection per `clarify-ai.md` are not affected — they fire only when confidence < 0.8.)
 
 **On any FAIL:** Emit `GATE_FAIL | phase=clarify | field=<path> | reason=missing`. **Do NOT modify artifacts to pass the gate.** **Do NOT update `.phase-status.json`.** Tell the user to answer the missing question or re-run Clarify.
```
leon1418, line 886: [🤖 AI review 🤖]

FYI: Gate check 7 inspects `config.disk_size_gb` "or legacy variants, all mapping to the same Q13b band" — but doesn't specify what happens when multiple `google_sql_database_instance` resources exist with *different* bands (e.g., one 8 GB dev instance + one 200 GB prod instance). The `clarify-database.md` auto-detect table handles this ("If multiple instances disagree, ask Q13b"), but this gate check's wording — "every … carries an unambiguous disk size … all mapping to the same Q13b band" — correctly implies the check fires only when they agree. Just noting this is implicit rather than explicit; the existing wording is technically sufficient.

leon1418 (commented):
[🤖 AI review 🤖]

**Scope reviewed:** All 6 changed files in full, PR description, author context comment, mainline versions of all modified files, sibling PRs #180/#182–#186 for merge-order interactions, and schema/script placement correctness.

**Summary:** This PR closes the gap between existing extraction-first *instructions* and a fail-closed *gate* that enforces them. The design is sound: gate checks 7–8 follow the same additive pattern as 1–6, the allocated-vs-actual semantic correction is well-reasoned with safe asymmetric override logic, and the runtime measurement block in `generate-artifacts-scripts.md` correctly degrades when connection vars are unset. The canonical `config.disk_size_gb` schema pin (with legacy fallback) eliminates an observed triple-path defensive-read pattern.

No blocking findings. Two observations below (Nit, FYI).

**What's good:**
- The fail-closed → `GATE_FAIL` model (same as 1–6) avoids the common antipattern of advisory-only checks that degrade silently.
- Asymmetric cross-check at Generate (allocated only overrides larger user answer, never smaller) respects the semantics correctly.
- The `pg_dump` hard-stop at ≥10 GB (dangerous direction) vs `pgcopydb` soft-advisory at <10 GB (merely inconvenient) is the right safety asymmetry.
- Legacy-path fallback acceptance in both Clarify and the schema prevents older inventories from breaking.

**Interactions with open sibling PRs:**
- **#182** (resume/recovery): touches `clarify.md` Step 0 (resume logic) and `clarify-global.md` — no overlap with #181's gate-check or Step 2 changes. Clean parallel merge.
- **#183** (journey orientation): touches `clarify.md` but in the Step 0/intro area, not the gate section. No conflict.
- **#184** (last-mile generate): touches `generate-artifacts-scripts.md` but only adds the fail-fast fill-in pattern in the shared conventions section (line ~62). #181's additions are at line ~83 (the tool-selection section). Adjacent but non-overlapping — should merge cleanly in either order.
- **#185** (decision gate): touches `SKILL.md` but in the state-machine table (line ~81+), not the Philosophy bullet (line ~10). No conflict with #181's one-liner.
- **#180** (heroku parity): entirely in `heroku-to-aws/` skill — zero file overlap.

**Recommended merge order:** #181 is independent and can land in any position relative to siblings. No dependency in either direction.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/phases/generate/generate-artifacts-scripts.md
herosjourney: Took the extension point now rather than leaving the caveat tooling-less — it was small since this PR already pins the canonical `config` keys (a36e6c6):

- Discover extracts `config.disk_autoresize` for `google_sql_database_instance`, but ONLY when the attribute is explicitly present in HCL. The provider default is `true`, but an omitted attribute isn't an authored choice, so absence stays neutral rather than synthesizing the default.
- This cross-check now branches on it: explicit `true` → allocation is a floor, not a cap; keep the user's band, no warning, note it in the script header. Explicit `false` or absent → existing warn-and-use-allocated behavior, with the warning text now mentioning autoresize-outside-Terraform as the push-back reason.

That keeps the fail-closed default (absent = warn) while suppressing the false-positive exactly where the Terraform proves growth is possible.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/phases/clarify/clarify.md
herosjourney: Made it explicit in a36e6c6 — gate check 7 now states outright that when instances map to different bands the check does not fire and asking Q13b is the expected behavior, quoting the `clarify-database.md` auto-detect rule so the two stay visibly in sync.

leon1418:
Pushed [d1833bf](https://github.com/awslabs/startups/pull/181/commits/d1833bf) to unblock `drift:check` — same cross-plugin situation as #182–#186: the six changed files landed on the migrate side only. Mechanical sync of their advisor twins (re-copy + the prefix/path/$id rewrites the tool normalizes); no changes to your content. Verified locally: drift:check OK (256 identical), vendored-shared OK, dprint clean. Note: overlaps with #182 on SKILL.md/clarify.md were simulated — they auto-merge cleanly (different regions), so whichever lands second just needs an Update branch.

leon1418 (approved):

az-zhu (approved):

PR #183, opened 2026-07-23
feat(gcp-to-aws): journey orientation at start + answer recap before design

herosjourney:
## Summary

**The problem, in plain English:** The flow is missing the first and last lines of a well-designed form. At the front: a founder who types "migrate my GCP stack" gets state-directory mechanics and then questions — never "here's what this is, how little of your time it takes, what you get at the end, and that you can stop and resume anytime." A founder who doesn't know resume exists treats any interruption as starting over. At the back: the essential questions are the only answers in the whole flow the user *states* rather than *confirms* — and they arrive as shorthand ("1A 2C 3 skip") and plain words ("about a grand") that the agent interprets. That interpretation step is exactly where an agent misreads, and it's the one step with zero user visibility: a misheard "2C" flows silently into Design, Estimate, and generated Terraform.

Changes:

- **Journey orientation (`discover.md` Step 0):** fresh runs open with a five-line block — six phases, your part is small (most answers come from your own files; confirm one sheet, answer 2–7 questions), what you get at the end (migrate-or-stay recommendation with costs, Terraform for the in-scope workloads, migration scripts), and that progress saves and resumes. Skipped on resume (the #169 breadcrumbs cover position there); explicitly non-blocking — discovery continues in the same turn.
- **Answer Recap (`clarify.md` Step 4, new subsection):** after the final essential batch (and Q27 when it fired), play back a compact table — question, the user's words, what was recorded — for Step 4 answers only; sheet rows were already confirmed at Step 2.5 and are not repeated. The recap **always waits** for "looks good" or a correction: the batch opener invites shorthand and plain-word answers, so nearly every real run involves interpretation, and a conditional show-or-wait rule would itself be an interpretation hazard for agents. One extra turn at the flow's highest-stakes commit is the intended cost. Sequence is fixed and explicit: recap → wait → Category E opt-in → Step 5. **"Use defaults for the rest" routes through the recap too**, with defaulted rows shown as "(default applied)" — bulk defaulting is the highest-risk interpretation and is the user's one chance to see what "the rest" actually meant. Corrections reuse the existing Step 2.5 override grammar (`chosen_by: "user"`). The full flow runs the same recap after its final batch (answered questions per-row; skipped/defaulted compressed to one line so a long table stays scannable). One new validation-checklist item.

## Why

[...]

PR #182, opened 2026-07-23
fix(gcp-to-aws): resume/recovery reliability — stale reuse, corrupt state, mid-flow read

herosjourney:
## Summary

**The problem, in plain English:** This plugin is built for multi-session use — six phases, state on disk, resume anywhere — but three of its recovery paths punish exactly the user who comes back. Reuse your preferences from a prior run and, if that file predates the current flow, Clarify dead-ends at a gate that forbids both fixing the file and proceeding. Get one truncated write to the state file and the documented remedy ("delete the file and restart the current phase") silently restarts the *entire migration from Discover*, because a missing state file initializes Phase 1 — even though every completed phase's artifacts are sitting intact in the same directory. And one instruction in the middle of Clarify tells the agent to read a file that does not exist yet at that point in the flow — or worse, on a re-run, reads a stale prior run's answer.

Changes:

- **Stale preferences reuse (`clarify.md` Step 0 Case 1):** before offering "re-use and skip questions," check the file for schema currency (`clarify_mode`, `prompt`/`design_consequence`, `source` on extracted/default constraints, `cpu_architecture` when compute exists) and discovery match (AI profile present but no `ai_constraints`, database in inventory but no `availability`/`db_size` → stale). Compatible files get the existing reuse offer, unchanged. Incompatible files get a new offer: keep still-valid answers — carried forward onto the Assumption Sheet for confirmation, gaps filled by the wizard — or start fresh. Case 2 (drafts) has had exactly this legacy handling since the wizard landed; Case 1 never got it. Also: "start fresh" now renames the old file to `preferences-superseded.json` instead of deleting it (irreversible discard of all prior answers), and Step 5 cleans the backup up once a new complete file exists.
- **Corrupt `.phase-status.json` recovery (`SKILL.md` State Validation rule 2):** replaced "Delete the file and restart the current phase" — self-contradictory, since with the file corrupt the current phase is unknowable, and per Workflow Execution a missing file initializes Discover. The remedy is now: infer completed phases from the artifacts on disk (inventory → discover, `preferences.json` → clarify, `aws-design*.json` → design, `estimation-*` → estimate, `generation-*`/`MIGRATION_GUIDE.md` → generate), confirm the inferred status with the user, rewrite the state file, resume. Explicitly noted as ground-truth reconstruction, not artifact-patching — the handoff-gate prohibition targets the latter.
[...]

PR #184, opened 2026-07-24
feat(gcp-to-aws): last-mile usability — fill-in checklist, placeholder guards, rollback runbook

herosjourney:
## Summary

**The problem, in plain English:** Everything before Generate produces decisions; Generate produces the things a founder actually runs against production — and that last mile has three traps, all observed in a real run's outputs rather than hypothesized. First: the run left 20 account-specific TODOs scattered across 10 files, and the only navigation aid is a `grep` command in the README — miss one and consequences range from broken (`ACCOUNT_ID` in the backend bucket name) to silently wrong. Second, the trap inside the first: `validation-report.json` said **passed** while `TODO-ops@example.com` sat in the config as the AWS account's security contact — `terraform validate` checks syntax, so `apply` would ship example.com contacts and a broken ECR URI without a word. Third: the generated cutover script's entire rollback story was one log line ("rollback = restore DNS to Cloud Run") — at the highest-stress, lowest-judgment moment of the migration, with an unanswered data question underneath it (rows written to the AWS database after cutover don't exist in Cloud SQL; a bare DNS rollback silently loses them).

Changes (all in `references/phases/generate/`):

- **Fill-in Checklist (`generate-artifacts-docs.md`):** the README's "TODO Items" grep hint becomes a consolidated table — one row per **value** (the same billing email in four places is one row), each with file locations and a concrete "where to get it" (command, console path, or artifact reference — never "your value here"). Mirrored in MIGRATION_GUIDE's Prerequisites; the grep one-liner stays as a done-check ("should return nothing"). Self-check now verifies every placeholder in the artifacts appears in the checklist.
- **Fill-once variables + placeholder guards (`generate-artifacts-infra.md`):** contact emails become `var.operations_email` / `var.billing_email` / `var.security_email` (no default), referenced by both the alternate contacts and the budget-alert subscribers — each value entered exactly once in tfvars instead of edited in four places. Every non-inferable variable gets a REQUIRED `validation` block rejecting placeholder tokens (`TODO`, `example.com`, `ACCOUNT_ID`, `<`), with error messages that name the tfvars key and point at the checklist — so the failure happens loudly at `terraform plan`, not silently at `apply`. The backend block can't use variables (Terraform limitation), so its `ACCOUNT_ID` is covered by a checklist row.
- **Rollback runbook contract (`generate-artifacts-docs.md`):** the Rollback Procedure subsection must contain (1) triggers with concrete criteria tied to `04-validate.sh` checks, (2) exact DNS commands with TTL implications, (3) an explicit answer to the post-cutover-writes data question — write-freeze + delta export, accepted loss (viable when Q12 said low-write; say so), or reverse sync by name — and (4) RTO. A one-line "rollback via DNS" fails the docs self-check ("rollback instructions match" was already there; now there's something substantive to match).
- **Fail-fast scripts (`generate-artifacts-scripts.md`):** at `--execute`, scripts check all user-supplied env vars up front and exit listing **every** missing one at once — never dying midway on the first empty var after work has started. Dry-run continues to work without them.
- **Execution-readiness line (`generate-artifacts-report.md`):** when artifacts contain placeholders, the report's first Next Step states "Fill in the [N] account-specific values — checklist in MIGRATION_GUIDE.md." The report is the decision document; this is the one line bridging decision to execution.

## Why

Companion to #181/#182/#183, which hardened the decision pipeline (extraction gates, resume/recovery, orientation + answer recap). This PR is where those decisions meet production. The placeholder-guard change carries the same philosophy as #181's gate checks — convert soft conventions ("go find the TODOs") into machine-enforced failures at the earliest safe moment (`terraform plan`) — and the checklist follows the same form-UX principle as #183 ("here's everything you need, in one place, before you start"). All three top findings came from inspecting an actual run's generated artifacts, not from reading the spec.

## Test plan

- [x] `dprint check` — clean
- [x] `mise run lint:frontmatter` — OK
- [x] Spec-only changes; no gate, flow, or schema changes (diff-verified: 4 files, +61/−13)
- [x] `validation` block HCL uses `strcontains` (Terraform ≥ 1.5, matches the plugin's `>= 1.0` floor via required_version bump not needed — function available in all currently supported Terraform releases)
- [ ] Manual: regenerate artifacts for a stack with Cloud SQL + Cloud Run and verify (a) the checklist row count matches `grep -rc TODO`, (b) `terraform plan` fails naming `billing_email` when tfvars keeps the placeholder, (c) the rollback subsection answers the post-cutover-writes question

## Out of scope

- Independent of #181's edits to `generate-artifacts-scripts.md` (different sections of the file; whichever merges second rebases trivially)
- Report chart/visualization additions — deliberately not proposed: the typography-first design is a stated choice, agent-generated SVG is a reliability risk with no validator coverage, and the baseline not-comparable rule (#175) forbids exactly the naive GCP-vs-AWS bar a chart would assert
- heroku-to-aws parity

## Type of Change

## Team Folder

- [x] `migrate/`


leon1418 (commented):
[🤖 AI review 🤖]

## Summary

Evidence-driven last-mile usability fix for generated migration artifacts. The three problems are real, the fixes are well-scoped, and each one converts a soft convention into a machine-enforced failure at the earliest safe moment. Design is sound.

**4 files, +61/−13. No schema, flow, or gate changes.** CI clean (build + 4 security scanners). MERGEABLE but BEHIND main.

### Findings: 1 Nit, 1 Optional

Both are non-blocking. No design, functionality, security, or correctness issues found.

### Sibling PR interactions
- **#181** touches `generate-artifacts-scripts.md` at lines 83–130 (db_size cross-check + runtime measurement). #184 adds at lines 62–76 (fail-fast env var check in the Script Rules section). **No textual conflict** — different hunks, different sections. Either merge order works.
- **#186** heavily refactors `generate-artifacts-report.md` (replaces lines 36–280 with a 43-line delegation to `report-decision-core.md`). #184's one-line change at line 100 of that file ("Next steps" execution-readiness bridge) **will textually conflict** — the paragraph #184 appends to is deleted by #186 and the rule moved elsewhere. Whichever merges second must rebase; the intent should be preserved inside `report-decision-core.md`.
- **#186** also modifies `generate-artifacts-docs.md` (TCO→run-rate, timeline→migration-shape). #184's checklist and rollback sections are in non-overlapping hunks. Low conflict risk but the "Timeline" reference at line 357 (unchanged by #184) will be renamed by #186.
- **#180, #182, #183, #185** touch no files in common with #184.

### Merge order recommendation
Merge #184 **before** #186 (the report file refactor). #184's one-line report change is trivial to preserve during #186's rebase, but rebasing the inverse (large deletion then re-add) is more error-prone.

### Validation gaps
- The manual test plan item (regenerate artifacts, verify checklist count, test `terraform plan` rejection, verify rollback completeness) is unchecked. No automated asserter covers the new checklist-completeness self-check rule or the rollback contract.
- The `strcontains` dependency is safe — the spec already requires `>= 1.5.0`.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/phases/generate/generate-artifacts-scripts.md
```diff
@@ -62,6 +62,20 @@ $MIGRATION_DIR/
 - Scripts are numbered for execution order
 - Scripts use `set -euo pipefail` for safety
 - Scripts log all actions to `$MIGRATION_DIR/logs/`
+- **Fail fast on unset fill-ins at `--execute`:** every script that reads user-supplied env vars (`SOURCE_HOST`, `TARGET_DB_PASSWORD`, `ALB_DNS`, …) checks them at the top of the execute path and exits listing ALL missing ones at once — never fails midway on the first empty var after work has started. Dry-run mode runs without them (printing which are unset). Pattern:
+
+  ```bash
+  if [ "$DRY_RUN" = false ]; then
+    missing=()
+    for v in SOURCE_HOST TARGET_HOST TARGET_DB_PASSWORD; do
+      [ -z "${!v:-}" ] && missing+=("$v")
```
leon1418, line 75: [🤖 AI review 🤖]

Nit: The indirect expansion `${!v:-}` is valid bash 4+ (confirmed empirically), but the script template also specifies `set -euo pipefail` which makes unset variables fatal via `-u`. With `nounset` active, `${!v:-}` on an unset variable expands to empty string safely (the `:-` default handles it) — so this is fine. However, the comment list `SOURCE_HOST TARGET_HOST TARGET_DB_PASSWORD` is an example — the actual variable list varies per script. Consider adding a one-line note like `# Adjust the list above to match this script's required env vars` to make it obvious this is a template to copy-edit, not a universal constant.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/phases/generate/generate-artifacts-docs.md
```diff
@@ -345,7 +364,7 @@ After generating documentation, verify:
 1. **Commands are syntactically correct**: All bash commands use correct syntax
 1. **No unresolved placeholders**: All `[placeholder]` values are replaced with actual data from artifacts
```
leon1418, line 372: [🤖 AI review 🤖]

Optional: The self-check item 8 still reads "Rollback instructions match: Rollback steps match generation plan" — unchanged by this PR. Given the substantial new rollback contract (triggers, exact commands, data-question posture, RTO), this self-check assertion is now underpowered. It could be strengthened to verify the four MUST-contain elements are present, not just that "steps match" generically. Not blocking because the rollback spec itself is the enforcing mechanism and this is a pre-existing check, but the asymmetry between the detailed contract (line 178–183) and the vague self-check (line 369) may let an incomplete rollback section slip through.

herosjourney:
Addressed both review notes:

**Nit (scripts fail-fast pattern):** Added an explicit note that the `for v in …` list is a per-script template to copy-edit, plus `# Adjust the list below to match this script's required env vars` in the bash snippet.

**Optional (rollback self-check):** Strengthened item 8 to require the four MUST-contain contract elements (triggers, exact commands + TTL, post-cutover data posture, RTO) and to fail a one-line "rollback via DNS".

leon1418:
Pushed [be45f1c](https://github.com/awslabs/startups/pull/184/commits/be45f1c11ce65a77bd568de9ada5a301662f27d5) to unblock `drift:check` — same cross-plugin situation as #185/#186: this branch predates #206's consolidation, so the four generate-artifacts files carried the last-mile changes on the migrate side only. Mechanical sync of their advisor twins (re-copy + the prefix/path/$id rewrites the tool normalizes); no changes to your content. Verified locally: drift:check OK (255 identical), vendored-shared OK, dprint clean. @herosjourney shout if you'd rather do this differently.

leon1418 (dismissed):

leon1418:
Resolved the conflict with main (#186 landed). This one was semantic, not mechanical, so for the record:

#186 extracted the executive-summary spec out of `generate-artifacts-report.md` into `references/shared/report-decision-core.md` (single source of truth, "do not restate its rules here"). This branch's edit to the old inline block was one net addition — the execution-readiness rule for Next steps ("when artifacts contain fill-in placeholders, the FIRST next step MUST state: fill in the [N] account-specific values… nothing can be applied until then").

Resolution: took main's side of the report file (the extraction stands) and **relocated that rule to its new home** — `report-decision-core.md` item 7, both plugin copies — so it now applies to the decision report at gate A too, not just the full report. The other three files (fill-in checklist in docs, placeholder guards in infra/scripts) auto-merged cleanly; spot-checked the content survived.

Verified: drift 256 identical, vendored-shared, dprint, 51 tests, 8 asserters all green.

leon1418 (dismissed):

advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/references/phases/generate/generate-artifacts-docs.md
```diff
@@ -175,7 +175,12 @@ If the command fails with dependency errors, investigate the listed resources be
 | High latency after migration                             | Suboptimal instance sizing                          | Review CloudWatch metrics and right-size                                                                                   |
 | `validation-report.json` shows `passed_degraded_offline` | Provider registry was unreachable when Generate ran | From a network-connected shell, run `cd terraform/ && terraform init && terraform validate` to complete the skipped checks |
 
-Rollback Procedure subsection (from generation plan — rollback triggers, steps, and RTO).
+Rollback Procedure subsection (from generation plan). A one-line "rollback via DNS" note is NOT sufficient — cutover is the highest-stress, lowest-judgment moment of the migration, and this subsection is the runbook the reader executes at 2am. It MUST contain:
+
+1. **Triggers — when to pull it:** concrete criteria, not vibes (e.g., "error rate > X% for 10 minutes on the validation checks in `04-validate.sh`", "data validation row counts diverge", "p99 latency > 2x GCP baseline after 30 minutes"). State who decides if the team is more than one person.
```
az-zhu: 04-validate.sh doesn't exist. The self-check can never legitimately pass as written, and the runbook cites a file that isn't in the package

advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/references/phases/generate/generate-artifacts-docs.md
```diff
@@ -323,12 +328,26 @@ combined GCP total or overall savings percentage.
...
+
+Scan every generated artifact (`terraform/`, `scripts/`, `ai-migration/`) for placeholders (`TODO`, `ACCOUNT_ID`, `<...>` tokens, `example.com` values) and render them as ONE consolidated table — this is the reader's single "before you can apply" list. One row per **value**, not per occurrence (the same billing email appearing in four places is one row listing all four locations):
+
+```markdown
+## Fill-in checklist — [N] values needed before `terraform plan`
+
+| #   | Value                               | Where to set it                    | Where to get it                                             |
+| --- | ----------------------------------- | ---------------------------------- | ----------------------------------------------------------- |
+| 1   | AWS account ID                      | `terraform/main.tf` backend bucket | `aws sts get-caller-identity --query Account --output text` |
+| 2   | Ops/billing/security contact emails | `terraform.tfvars`                 | Your team — real inboxes, not aliases you don't read        |
```
az-zhu: `terraform.tfvars` isn't generated - only `terraform.tfvars.example` is (Step 2b). The checklist needs the `cp` step first

advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/references/phases/generate/generate-artifacts-docs.md
herosjourney: Confirmed and fixed in 03e499e — checked the actual script manifest in generate-artifacts-scripts.md: 04 is `migrate-secrets.sh`, validation is always `05-validate-migration.sh`. The trigger example and the self-check both cited the wrong (nonexistent) file, so as you said, the self-check could never legitimately pass. Fixed both citations in this file, in both plugin copies.

advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/references/phases/generate/generate-artifacts-docs.md
herosjourney: Good catch — confirmed Step 2b only ever writes `terraform.tfvars.example`; the real `terraform.tfvars` is gitignored and never generated. Added the `cp terraform.tfvars.example terraform.tfvars` step as row 0 of the checklist template (every other tfvars-pointing row depends on it), noted the Step 2b constraint in the prose above the table, and required row 0 in the docs self-check. Fixed in 03e499e, both plugin copies.

az-zhu (approved):

PR #196, opened 2026-08-04
fix(tf-best-practices): detect IPv6 public ingress and fix order-dependent IPv4 check in Terraform policy validator

leon1418:
## Problem

`skills/tf-best-practices/scripts/validate-terraform-policy.py` reported **`POLICY_OK`, exit 0** for security groups that are open to the entire internet. The validator is used as a gate, so it told users an exposed configuration was safe — worse than having no validator, because it manufactures false assurance.

Two defects, both in `_ingress_allows_public`:

**1. `::/0` was never checked.** No IPv6 "entire internet" range appeared anywhere in the file. An SG opening port 22 or 5432 to `ipv6_cidr_blocks = ["::/0"]` passed cleanly. On IPv6-default and dual-stack VPCs that is a live exposure.

**2. The IPv4 check was order-dependent.** The regex

```python
re.search(r"cidr_blocks\s*=\s*\[(.*?)\]", ingress_body, re.DOTALL)
```

also matches the tail of `ipv6_cidr_blocks`. Since `re.search` returns the *first* match, when the IPv6 attribute was declared first the checker read the **IPv6** list while believing it was reading IPv4, then searched it for `0.0.0.0/0`.

This second defect is the more serious one, because it breaks the rule's original purpose with **no IPv6 exposure involved at all**. Verified against the pre-fix code — the same SG, semantically identical, differing only in the order of two lines:

| Attribute order | `cidr_blocks = ["0.0.0.0/0"]` on port 22 |
|---|---|
| `ipv6_cidr_blocks` first | **`POLICY_OK`, exit 0** ❌ |
| `cidr_blocks` first | `POLICY_FAIL`, exit 1 ✅ |

So a plain IPv4 SSH-open-to-the-world finding was missed in purely IPv4 environments too, whenever `ipv6_cidr_blocks` happened to be written first. Neither Terraform nor common formatters constrain that ordering.

## Solution

Replace `_ingress_allows_public` with `_ingress_public_cidrs`, which:

- reads `cidr_blocks` and `ipv6_cidr_blocks` as **separate, line-anchored** attributes (`^\s*{attr}\s*=`), matching the `_attr_string` / `_attr_int` idiom already used in this file, so the two lists can never be confused;
- checks each address family against its own "entire internet" range (`0.0.0.0/0` / `::/0`);
- returns the matched CIDRs, so violation messages name the family that actually fired instead of hardcoding `0.0.0.0/0`.

Both `db_sg_no_public_ingress` and `sg_no_public_admin_ingress` consumed the old helper, so both rules are fixed by the one change: both now cover dual-stack and IPv6-only VPCs, and neither depends on attribute order.

The existing fail-open scope is deliberately unchanged — only inline `ingress { ... }` blocks are inspected, and separate `aws_security_group_rule` / `aws_vpc_security_group_ingress_rule` resources still fail open, as documented.

### Tests

- **`bad-sg-public-ipv6`** — `ipv6_cidr_blocks = ["::/0"]` declared *before* a benign IPv4 list, on ports 22 and 5432. Was `POLICY_OK`; now `POLICY_FAIL` on both rules.
- **`good-sg-ipv6-scoped`** — false-positive guard: public web over `::/0` (not a sensitive port) plus SSH/Postgres over a scoped `2001:db8:1234::/48` must still pass.
- **`test_ipv4_public_ingress_caught_when_ipv6_list_declared_first`** — asserts the IPv4 verdict is identical under both attribute orderings. Confirmed to fail against the pre-fix validator on the `ipv6-first` ordering, so it is a real regression guard rather than a tautology.

27 tests pass. New fixtures are already covered by the existing `.checkov.yaml` `skip-path` entry for `fixtures/terraform-policy`, and use RFC 3849 documentation addresses.

### Note for maintainers

Two pre-existing items I did **not** change, to keep this PR scoped:

1. `mise run build` and CI do not run this skill's `pytest` suite (`[tasks.test]` runs only `node --test` on the frontmatter validator; no workflow invokes pytest). I ran it manually — `27 passed` — but CI will not catch regressions in these rules. Happy to wire it up in a follow-up if you want.
2. The inline-only fail-open scope means a `::/0` or `0.0.0.0/0` exposure declared via a separate `aws_security_group_rule` resource still passes silently. Documented behaviour, but the same class of false assurance; worth a separate issue if gate coverage there is wanted.

## Type of Change

- [ ] New plugin/power/tool
- [x] Bug fix
- [ ] Documentation update
- [ ] Guardrail/CI update

## Team Folder

- [ ] `advisor/`
- [x] `migrate/`
- [ ] Other: ___

## Checklist

### DIFF
--- a/migrate/plugins/migration-to-aws/skills/tf-best-practices/scripts/validate-terraform-policy.py
+++ b/migrate/plugins/migration-to-aws/skills/tf-best-practices/scripts/validate-terraform-policy.py
@@ -14,13 +14,17 @@ in-block literal evidence, so a valid stack is never falsely blocked):
   - rds_not_public: aws_db_instance / aws_rds_cluster must not set
     publicly_accessible = true (fail-open when variable-driven or absent).
   - db_sg_no_public_ingress: an inline aws_security_group ingress covering a
-    database port (5432 / 3306) must not allow 0.0.0.0/0 (fail-open on
-    separate aws_security_group_rule / aws_vpc_security_group_ingress_rule
-    resources, which this static reader cannot correlate).
+    database port (5432 / 3306) must not allow 0.0.0.0/0 (IPv4) or ::/0 (IPv6)
+    (fail-open on separate aws_security_group_rule /
+    aws_vpc_security_group_ingress_rule resources, which this static reader
+    cannot correlate).
   - sg_no_public_admin_ingress: an inline ingress must not open a well-known
     admin/datastore port (SSH, RDP, Redis, Memcached, Mongo, Elasticsearch,
-    Kibana) to 0.0.0.0/0. Scoped to a fixed never-public port list — web ports
-    and app/game ports are not flagged. Same inline-only fail-open scope.
+    Kibana) to 0.0.0.0/0 or ::/0. Scoped to a fixed never-public port list — web
+    ports and app/game ports are not flagged. Same inline-only fail-open scope.
+    Both SG rules read cidr_blocks and ipv6_cidr_blocks independently: an
+    IPv6-only or dual-stack VPC is exposed by ::/0 exactly as IPv4 is by
+    0.0.0.0/0.
   - no_wildcard_iam: a literal IAM policy document with Effect "Allow" must not
     use Action "*" or Resource "*" (fail-open on aws_iam_policy_document data
     sources, whose statements are not visible as literal JSON here).
@@ -556,16 +560,39 @@ def _ingress_covers_db_port(ingress_body: str) -> bool:
     return bool(_ingress_covered_ports(ingress_body, _DB_PORTS))
 
 
-def _ingress_allows_public(ingress_body: str) -> bool:
-    # Match a cidr_blocks list literal and look for 0.0.0.0/0 inside it.
-    m = re.search(r"cidr_blocks\s*=\s*\[(.*?)\]", ingress_body, re.DOTALL)
-    if not m:
-        return False
-    return "0.0.0.0/0" in m.group(1)
+def _attr_list_inner(block: str, attr: str) -> str | None:
+    """Return the raw text inside a literal `attr = [ ... ]`, else None.
+
+    Anchored at line start (like _attr_string / _attr_int) so that `cidr_blocks`
+    does NOT also match the tail of `ipv6_cidr_blocks` — reading one list while
+    believing it is the other silently inverts the verdict.
+    """
+    m = re.search(
+        rf"^\s*{re.escape(attr)}\s*=\s*\[(.*?)\]",
+        block,
+        re.DOTALL | re.MULTILINE,
+    )
+    return m.group(1) if m else None
+
+
+def _ingress_public_cidrs(ingress_body: str) -> list[str]:
+    """Return the "entire internet" CIDRs this ingress rule allows, both families.
+
+    IPv4 `0.0.0.0/0` in cidr_blocks and IPv6 `::/0` in ipv6_cidr_blocks are
+    equally public: on a dual-stack or IPv6-only VPC, ::/0 on an admin port is a
+    live exposure. Each list is read under its own anchored attribute name so the
+    two are never confused. Empty list => no unambiguous public exposure.
+    """
+    found: list[str] = []
+    for attr, public in (("cidr_blocks", "0.0.0.0/0"), ("ipv6_cidr_blocks", "::/0")):
+        inner = _attr_list_inner(ingress_body, attr)
+        if inner is not None and public in inner:
+            found.append(public)
+    return found
 
 
 def check_db_sg_no_public_ingress(tf_files: list[tuple[str, str]]) -> list[Violation]:
-    """Flag inline security-group ingress that opens a DB port to 0.0.0.0/0.
+    """Flag inline security-group ingress that opens a DB port to 0.0.0.0/0 or ::/0.
 
     Fail-open: only INLINE `ingress { ... }` blocks inside aws_security_group are
     inspected. Separate aws_security_group_rule / aws_vpc_security_group_ingress_rule
@@ -577,7 +604,8 @@ def check_db_sg_no_public_ingress(tf_files: list[tuple[str, str]]) -> list[Viola
         for name, body, line in _extract_blocks(content, "aws_security_group"):
             # Walk each inline ingress block via brace matching.
             for ingress_body in _own_blocks(body, "ingress"):
-                if _ingress_covers_db_port(ingress_body) and _ingress_allows_public(ingress_body):
+                public = _ingress_public_cidrs(ingress_body)
+                if _ingress_covers_db_port(ingress_body) and public:
                     violations.append(
                         Violation(
                             check="policy",
@@ -587,12 +615,12 @@ def check_db_sg_no_public_ingress(tf_files: list[tuple[str, str]]) -> list[Viola
                             severity="error",
                             summary=(
                                 f"aws_security_group '{name}' has an ingress rule that opens a "
-                                "database port (5432/3306) to 0.0.0.0/0"
+                                f"database port (5432/3306) to {' and '.join(public)}"
                             ),
                             fix_hint=(
                                 "Restrict the ingress to the application security group "
                                 "(security_groups = [aws_security_group.app.id]) or a private "
-                                "CIDR — never 0.0.0.0/0 for a database port"
+                                "CIDR — never 0.0.0.0/0 or ::/0 for a database port"
                             ),
                         )
                     )
@@ -601,7 +629,8 @@ def check_db_sg_no_public_ingress(tf_files: list[tuple[str, str]]) -> list[Viola
 
 def check_sg_no_public_admin_ingress(tf_files: list[tuple[str, str]]) -> list[Violation]:
     """Flag inline security-group ingress that opens a well-known admin/datastore
-    port (SSH, RDP, Redis, Memcached, Mongo, Elasticsearch, Kibana) to 0.0.0.0/0.
+    port (SSH, RDP, Redis, Memcached, Mongo, Elasticsearch, Kibana) to 0.0.0.0/0
+    or ::/0.
 
     Deliberately scoped to a fixed list of ports that are ~never legitimately
     public — NOT "any public ingress" — so valid public workloads (web on
@@ -622,7 +651,8 @@ def check_sg_no_public_admin_ingress(tf_files: list[tuple[str, str]]) -> list[Vi
     for rel_path, content in tf_files:
         for name, body, line in _extract_blocks(content, "aws_security_group"):
             for ingress_body in _own_blocks(body, "ingress"):
-                if not _ingress_allows_public(ingress_body):
+                public = _ingress_public_cidrs(ingress_body)
+                if not public:
                     continue
                 hit = _ingress_covered_ports(ingress_body, _SENSITIVE_NONDB_PORTS)
                 if not hit:
@@ -637,7 +667,7 @@ def check_sg_no_public_admin_ingress(tf_files: list[tuple[str, str]]) -> list[Vi
                         severity="error",
                         summary=(
                             f"aws_security_group '{name}' opens sensitive port(s) {labels} "
-                            "to 0.0.0.0/0"
+                            f"to {' and '.join(public)}"
                         ),
                         fix_hint=(
                             "Restrict this ingress to a bastion/app security group or a private "
--- a/migrate/plugins/migration-to-aws/skills/tf-best-practices/scripts/validate-terraform-policy.py
+++ b/migrate/plugins/migration-to-aws/skills/tf-best-practices/scripts/validate-terraform-policy.py
@@ -499,7 +499,8 @@ def check_alb_https_policy(terraform_dir: Path) -> list[Violation]:
 
 _DB_PORTS = (5432, 3306)
 
-# Well-known admin / datastore ports that should never be open to 0.0.0.0/0.
+# Well-known admin / datastore ports that should never be open to the whole
+# internet in either address family (0.0.0.0/0 or ::/0).
 # Deliberately EXCLUDES 5432/3306 (covered by db_sg_no_public_ingress, so no
 # double-reporting) and web ports 80/443 (legitimately public). Kept tight to
 # unambiguous "never public" ports so valid designs (e.g. game servers on high
--- a/migrate/plugins/migration-to-aws/skills/tf-best-practices/scripts/validate-terraform-policy.py
+++ b/migrate/plugins/migration-to-aws/skills/tf-best-practices/scripts/validate-terraform-policy.py
@@ -91,15 +91,19 @@ def _read_tf_files(terraform_dir: Path) -> list[tuple[str, str]]:
 _HEREDOC_OPEN = re.compile(r"<<(-?)\s*([A-Za-z_][A-Za-z0-9_]*)")
 
 
-def _lex_hcl(block: str) -> tuple[list[int], list[bool]]:
-    """Scan HCL once, returning per-index (brace depth, is-real-code) arrays.
+def _lex_hcl(block: str) -> tuple[list[int], list[bool], list[bool]]:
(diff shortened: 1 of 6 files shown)


leon1418 (commented):
[🤖 AI review 🤖]

**Scope reviewed:** All 6 changed files (validator, test file, 2 fixtures, SKILL.md, security-posture-rules.md). Full PR description, CI checks, existing review threads (none), and interaction with PR #197.

**Summary:** Clean, well-scoped security bug fix. The two defects are real and properly demonstrated — I empirically reproduced both the order-dependence (IPv6 list declared first → IPv4 `0.0.0.0/0` missed) and the missing `::/0` coverage. The new `_attr_list_inner` + `_ingress_public_cidrs` design follows the existing `_attr_string` / `_attr_int` idiom with the correct `^\s*` line anchor + `re.MULTILINE`, preventing the `ipv6_cidr_blocks` tail-match that caused the original bug.

**Verification performed:**
- 27/27 pytest tests pass
- `bad-sg-public-ipv6` correctly fires both `sg_no_public_admin_ingress` and `db_sg_no_public_ingress` with `::/0` in the summary
- `good-sg-ipv6-scoped` correctly passes (no false positive on scoped IPv6 or `::/0` on web ports)
- `good-sg-public-webapp` still passes (no regression)
- `bad-sg-public-ssh` still catches IPv4 `0.0.0.0/0` (no regression)
- Tested edge cases: multiline lists, `#`/`//` comments before attr (safe), empty lists, attribute ordering combinations — all correct
- Confirmed the `^\s*` anchor prevents matching `ipv6_cidr_blocks` when searching for `cidr_blocks`

**Design:** The refactor from `_ingress_allows_public` (returns bool) to `_ingress_public_cidrs` (returns matched CIDRs) is the right choice — it enables the violation messages to name the specific family that fired, which is operationally useful.

**PR #197 interaction:** These two PRs will conflict on merge (both modify `check_db_sg_no_public_ingress` and `check_sg_no_public_admin_ingress`). #197 replaces the `re.finditer + _extract_braced_block` ingress iteration with `_own_blocks()`. The resolution is straightforward: use #197's `_own_blocks` iterator with #196's `_ingress_public_cidrs` logic. The conflict is mechanical, not semantic.

**Pre-existing limitations (not introduced by this PR):** `/* */` block comments and heredocs containing `cidr_blocks = [...]` on their own line can cause false positives — same limitation shared by `_attr_string` and `_attr_int`. This is the class of issue #197 addresses for `_extract_braced_block`. Not a blocking concern for this PR.

**No actionable findings.** The fix is correct, the tests are strong regression guards (the order-independence test is especially well-designed — verified it fails against the pre-fix code), documentation is updated to match, and CI passes.

**Recommendation:** Merge. If #197 lands first, rebase and resolve the mechanical conflict in the two `check_*` functions.

migrate/plugins/migration-to-aws/skills/tf-best-practices/scripts/validate-terraform-policy.py
```diff
@@ -330,16 +334,39 @@ def _ingress_covers_db_port(ingress_body: str) -> bool:
...
-    return "0.0.0.0/0" in m.group(1)
+def _attr_list_inner(block: str, attr: str) -> str | None:
+    """Return the raw text inside a literal `attr = [ ... ]`, else None.
+
+    Anchored at line start (like _attr_string / _attr_int) so that `cidr_blocks`
+    does NOT also match the tail of `ipv6_cidr_blocks` — reading one list while
+    believing it is the other silently inverts the verdict.
+    """
+    m = re.search(
+        rf"^\s*{re.escape(attr)}\s*=\s*\[(.*?)\]",
```
ayn-builds: **Blocking: this anchor introduces a false negative the old regex did not have.**

`^\s*` requires the attribute to start a line, so an attribute sharing the line with the opening brace is now invisible:

```hcl
ingress { cidr_blocks = ["0.0.0.0/0"]
  from_port = 22
  to_port   = 22
  protocol  = "tcp"
}
```

| | verdict |
|---|---|
| pre-fix | `POLICY_FAIL` |
| this PR | `POLICY_OK` |

Valid HCL, and low frequency in practice (`terraform fmt` normalises it, and `_attr_int` on `from_port` already shares the blind spot), but it is the same class of false assurance this PR closes, just traded to a different shape. One-character-class fix:

```python
rf"(?:^|\{{)\s*{re.escape(attr)}\s*=\s*\[(.*?)\]"
```

This keeps all three behaviours you want: brace-same-line reads `0.0.0.0/0`, ipv6-first still reads the correct IPv4 list, and a commented-out `# cidr_blocks = [...]` still does not match. A `(?<![\w.])` lookbehind also fixes the ipv6 confusion but reintroduces the comment false positive, so the brace alternation is the better of the two.

ayn-builds (commented):
Re-posting the three blocking items as inline comments on the lines they refer to, replacing my earlier top-level comment. Verification unchanged: reproduced before/after against head a565f06, 27 tests pass, markdownlint clean, dprint clean, fixtures covered by the existing checkov skip-path, IPv6 addresses correctly RFC 3849. Both defects are real and the order-dependence finding is a good catch.

All three below leave the same false-assurance property this PR exists to close: the first is a regression introduced by this diff, the second a gap in the new code path, the third a live false positive that the second fix resolves anyway. Filing as a comment rather than request-changes; non-blocking notes (tests, a stale comment at line 275, two more fail-open gaps, CI wiring) available separately.

migrate/plugins/migration-to-aws/skills/tf-best-practices/scripts/validate-terraform-policy.py
```diff
@@ -330,16 +334,39 @@ def _ingress_covers_db_port(ingress_body: str) -> bool:
...
+
+    IPv4 `0.0.0.0/0` in cidr_blocks and IPv6 `::/0` in ipv6_cidr_blocks are
+    equally public: on a dual-stack or IPv6-only VPC, ::/0 on an admin port is a
+    live exposure. Each list is read under its own anchored attribute name so the
+    two are never confused. Empty list => no unambiguous public exposure.
+    """
+    found: list[str] = []
+    for attr, public in (("cidr_blocks", "0.0.0.0/0"), ("ipv6_cidr_blocks", "::/0")):
+        inner = _attr_list_inner(ingress_body, attr)
+        if inner is not None and public in inner:
```
ayn-builds: **Blocking: non-canonical `::/0` spellings still pass.**

`public in inner` is a substring test, so semantically identical writings slip through:

```hcl
ipv6_cidr_blocks = ["::0/0"]               # POLICY_OK
ipv6_cidr_blocks = ["0:0:0:0:0:0:0:0/0"]   # POLICY_OK
```

Unlike attribute ordering, nothing normalises these: `terraform fmt` treats CIDRs as opaque strings, and IPv6 has many more legal spellings of the zero address than IPv4 does, so this bites hardest on the family this PR is adding. Since the PR establishes the IPv6 path, better to get it right here than in a third round. Split the list on `,` and canonicalise each token:

```python
def _is_entire_internet(token: str) -> bool:
    try:
        return ipaddress.ip_network(token.strip().strip('"'), strict=False).prefixlen == 0
    except ValueError:
        return False   # var.x, non-literal, malformed -> fail open, as today
```

Verified this accepts all four zero-prefix spellings and rejects `10.0.0.0/8`, `2001:db8:1234::/48`, `0.0.0.0/1`, and `var.cidr`, so the documented fail-open scope is unchanged. It also lets the two families collapse into one loop, and fixes the comment false positive below for free.

ayn-builds: **Blocking: false positive - a CIDR mentioned in a comment inside the list fires.**

Because the check searches the raw bracket text, a comment inside the list counts:

```hcl
ingress {
  from_port = 3389
  to_port   = 3389
  protocol  = "tcp"
  cidr_blocks = [
    # TODO: was 0.0.0.0/0
    "10.0.0.0/8",
  ]
}
```

This reports `POLICY_FAIL` on `sg_no_public_admin_ingress` for a correctly scoped security group. Pre-existing, but relevant here for two reasons.

First, your anchoring change already half-fixed this class: a commented-out `# cidr_blocks = ["0.0.0.0/0"]` above a `security_groups = [...]` line went from `POLICY_FAIL` to `POLICY_OK`. That is a real improvement you did not claim in the description, and worth a test to lock in.

Second, per the CONTRIBUTING note that a false positive blocks a real migration, this is the more damaging direction of the two. Tokenising per the comment above closes it: I confirmed the comment token fails `ip_network()` and is skipped.

migrate/plugins/migration-to-aws/skills/tf-best-practices/scripts/validate-terraform-policy.py
leon1418: Confirmed and fixed in `cc8cac8`. You're right that this is a regression I introduced, and it is the more serious of the three because it is a *new* miss on the plain IPv4 case the rule exists for.

**Reproduced your shape exactly:**

| | verdict |
|---|---|
| `upstream/main` | `POLICY_FAIL`, `sg_no_public_admin_ingress` |
| `a565f06` (this PR before the fix) | `POLICY_OK` |
| `cc8cac8` | `POLICY_FAIL`, `22 (SSH) to 0.0.0.0/0` |

I took your `(?:^|\{)` alternation rather than the lookbehind, for the reason you gave — I verified the `(?<![\w.])` variant does reintroduce the commented-out false positive, so the brace arm is strictly better. All three properties hold: brace-line reads `0.0.0.0/0`, ipv6-first still reads the correct IPv4 list, and `# cidr_blocks = [...]` still does not match.

**Worth flagging: the same blind spot is still live in `_attr_int` / `_attr_string`, which #197 now owns.** Your parenthetical called this out, and it turns out to be reachable end to end, not just latent. A brace-line attribute on a resource body is missed on both `main` and #197's head:

```hcl
resource "aws_db_instance" "bad" { publicly_accessible = true
  identifier        = "primary"
  instance_class    = "db.t3.micro"
  storage_encrypted = true
}
```

Both report `POLICY_OK` — a public RDS certified clean. It is pre-existing rather than a #197 regression, but #197 rewrote those readers onto `_first_own_attr` + `_lex_hcl` and kept the line anchor, so that PR is where the same one-character-class fix belongs. I found it because my first cross-PR test put `from_port` on the brace line and stayed `POLICY_OK` even with both PRs applied. I've raised it on #197 rather than widening this PR into readers it does not touch.

`test_cidr_attribute_sharing_line_with_opening_brace_is_read` pins your case; mutation-checked that it fails against `a565f06`.

migrate/plugins/migration-to-aws/skills/tf-best-practices/scripts/validate-terraform-policy.py
leon1418: Confirmed and fixed in `cc8cac8`. Agreed on the reasoning for fixing it here rather than in a third round — this PR is what establishes the IPv6 path, and your point that nothing normalises these is the deciding one: `terraform fmt` treats a CIDR as an opaque string, so unlike attribute ordering there is no tool that will quietly make the canonical spelling the common one.

**Reproduced both**, plus a third spelling:

| `ipv6_cidr_blocks` value | `main` | `a565f06` | `cc8cac8` |
|---|---|---|---|
| `::/0` | `POLICY_OK` | `POLICY_FAIL` | `POLICY_FAIL` |
| `::0/0` | `POLICY_OK` | `POLICY_OK` | `POLICY_FAIL` |
| `0:0:0:0:0:0:0:0/0` | `POLICY_OK` | `POLICY_OK` | `POLICY_FAIL` |
| `0000:0000:...:0000/0` | `POLICY_OK` | `POLICY_OK` | `POLICY_FAIL` |

Adopted your `ip_network(...).prefixlen == 0` canonicalisation, and confirmed the fail-open scope is unchanged — `10.0.0.0/8`, `2001:db8:1234::/48`, `0.0.0.0/1`, `128.0.0.0/1`, `var.admin_cidrs` and `local.allowed` all still pass. That is pinned as `test_scoped_and_nonliteral_cidrs_still_fail_open`, which is deliberately an invariant rather than a regression guard: it passes against `main` too, and its job is to prove canonicalisation did not *widen* what counts as public.

**One departure from the snippet, because it trades this finding for a miss.** Splitting on `,` fuses a comment with the entry that follows it, so the tokeniser turns a real exposure into one unparseable string:

```hcl
cidr_blocks = [
  # temporary, remove before prod
  "0.0.0.0/0",
]
```

`split(",")` yields `'# temporary, remove before prod\n      "0.0.0.0/0"'` as a single token — `ip_network()` rejects it, and the SG reads clean. Verified this is a real `POLICY_FAIL` → `POLICY_OK` flip on the suggested version, i.e. exactly the false-assurance direction we are both trying to close, just moved.

So I strip comments first and then extract quoted literals, which gets your finding 3 as well without the trade. Both directions are pinned in `test_cidr_named_only_in_a_comment_is_not_an_allowed_range` (comment-only → `POLICY_OK`; comment followed by a real entry → `POLICY_FAIL`).

leon1418: Confirmed and fixed in `cc8cac8`. Reproduced your exact fixture — `POLICY_FAIL` on `sg_no_public_admin_ingress` for an SG whose only live entry is `10.0.0.0/8`. Agreed on the priority argument: per CONTRIBUTING a false positive blocks a real migration, so of the three findings this is the one that damages a *correct* stack, and it is now `POLICY_OK`.

Fixed by the comment stripping described in my reply above, so tokenising did close it for free as you predicted — with the one adjustment that comments are removed *before* the list is read rather than filtered per-token, which avoids the fused-token miss.

**On the accidental half-fix — verified and locked in.** You were right that I had not claimed it, and I had not noticed it. Checked it as its own case:

```hcl
ingress {
  from_port = 22
  to_port   = 22
  protocol  = "tcp"
  # cidr_blocks = ["0.0.0.0/0"]
  security_groups = ["sg-abc123"]
}
```

| | verdict |
|---|---|
| `upstream/main` | `POLICY_FAIL` — false positive on a correctly scoped SG |
| `a565f06` | `POLICY_OK` |
| `cc8cac8` | `POLICY_OK` |

`test_commented_out_cidr_attribute_is_not_read` pins it. Mutation-checked in the direction that matters: it **fails** against `upstream/main` and passes from `a565f06` on, so it is a genuine guard for the improvement rather than a test that merely agrees with current behaviour. It would previously have been held only by the anchor; now it is held by the anchor *and* comment stripping, so removing either is caught.

Both these lists and their bodies are still read with regexes rather than a real HCL scanner, so this class is closed for the shapes above but not eliminated in general — a `]` inside a comment inside the list still truncates the match, for instance. That is the same boundary #197 is building `_lex_hcl` for, and once both land the honest follow-up is to route `_attr_list_inner` through that shared lexer instead of keeping a second notion of what counts as code. I would rather note that than imply the regex is airtight.

Docs updated: `references/security-posture-rules.md` now states canonicalisation and the comment-scoping rule, since both change what the gate means.

leon1418: Also picked up the stale comment you mentioned was available separately — I went looking for it rather than waiting. `_SENSITIVE_NONDB_PORTS` was introduced with "ports that should never be open to `0.0.0.0/0`", which this PR made inaccurate: the port list is family-agnostic and both rules now cover `::/0` too. Fixed in `4eb1fa4`.

Swept the rest of the file for the same staleness and the other seven `0.0.0.0/0` mentions are all correct as written — either naming both families explicitly or deliberately referring to the IPv4 range alone.

If you'd still like to post the remaining non-blocking notes (the two further fail-open gaps and the CI wiring), please do. On CI wiring specifically I already agree it is a real gap and said so in the PR description: neither `mise run build` nor any workflow runs this suite, so the green checks on this PR do not cover its 32 tests. I have kept it out of scope here rather than bundle a CI change into a security fix, but I am happy to wire it up in a follow-up.

ayn-builds (approved):

PR #197, opened 2026-08-04
fix(tf-best-practices): read quoted integer ports in policy validator

leon1418:
## Problem

`_attr_int` in `skills/tf-best-practices/scripts/validate-terraform-policy.py` only matched a **bare** integer:

```python
match = re.search(rf"^\s*{re.escape(attr)}\s*=\s*(\d+)", block, re.MULTILINE)
```

Terraform accepts a quoted integer for number-typed arguments — `port = "443"` is valid, idiomatic and `terraform fmt`-clean HCL that converts to `443`. On a quoted value the regex fails, `_attr_int` returns `None`, and all three of its call sites silently degrade:

1. `_parse_listeners` (`port`) → a **correct** HTTPS listener written as `port = "443"` yields `port=None`, so `https_ok` is empty and the config is reported as `POLICY_FAIL / alb_https_listener`. **False positive on a correct stack.**

2. `_ingress_covered_ports` (`from_port` / `to_port`) → this is the more serious direction. A security group with `from_port = "5432"`, `to_port = "5432"`, `cidr_blocks = ["0.0.0.0/0"]` produces an empty covered-ports list, so `db_sg_no_public_ingress` and `sg_no_public_admin_ingress` **never fire**. A genuinely world-open database or SSH port is reported as `POLICY_OK` — a **fail-closed miss on a real exposure**.

Reproduced before the fix:

| input | verdict |
|---|---|
| internet-facing ALB + HTTPS listener `port = "443"` (cert + forward) | `POLICY_FAIL`, `rule=alb_https_listener` |
| same with bare `port = 443` | `POLICY_OK` |
| inline SG ingress `from_port`/`to_port = "5432"`, `cidr_blocks = ["0.0.0.0/0"]` | `POLICY_OK` — silent miss |

The verdict flipped on quoting alone. Beyond the missed exposure, the false positive is corrosive on its own: correct configs getting flagged trains users to distrust and ignore the validator, so they miss the real failures too.

## Solution

Fix `_attr_int` to accept both the bare and quoted integer forms. One change repairs all three call sites.

The bare alternative is byte-identical to the previous pattern. **Correction to an earlier version of this description:** I originally wrote that this made existing behaviour unchanged, which was too strong — as raised in review, accepting the quoted form makes a *nested* block's attribute a match candidate where a bare-only pattern could never match one. A follow-up commit addresses that at the root; see **Block-scoped attribute reads** below.

The quoted arm requires the value to be **entirely** digits (the closing quote is anchored against the digits), so the fix does not become sloppy — `"443x"`, `"443-444"` and `"${var.port}"` still return `None`, preserving the module's documented fail-open posture for non-literal values:

| input | result |
|---|---|
| `port = 443` | `443` |
| `port = "443"` | `443` |
| `port = "443x"` | `None` |
| `port = "443-444"` | `None` |
| `port = "${var.p}"` | `None` |
| `port = var.p` | `None` |

### Block-scoped attribute reads (added after review)

`_extract_blocks` returns a resource body *including* its nested blocks, and these readers used `re.search` — first match anywhere wins. Accepting quoted integers therefore changed *which* match wins for a listener whose own `port`/`protocol` are written after its `default_action` (valid HCL; `terraform fmt` does not reorder attributes relative to blocks):

```hcl
resource "aws_lb_listener" "http_redirect" {
  default_action {
    type = "redirect"
    redirect {
      port     = "443"
      protocol = "HTTPS"
    }
  }

  port     = 80
  protocol = "HTTP"
}
```

`_attr_int(body, "port")` returned **443** on that body where the previous pattern returned **80** — reproduced directly.

**No verdict could flip**, which I verified independently before and after: a `forward` action carries no nested `port`/`protocol`, so `alb_http_redirect` still fires on a genuinely bad HTTP:80 forward listener even with the attributes reordered; and `_default_action_type` reads `type` by brace matching, so a `redirect` action can never satisfy the `forward` requirement in `https_ok`. But the reason it was safe was that **`_attr_string` independently misread `protocol` as `HTTPS` on the same body** — a pre-existing weakness — so both ALB rules skipped the listener either way. Two misreads cancelling is not a property worth depending on, and `aws_lb_listener_rule` carries a `redirect { port }` too.

So match candidates are now filtered by brace depth, and only those at the block's own depth are eligible. This repairs the pre-existing `_attr_string` weakness in the same stroke, so that listener now evaluates as **80 / HTTP** — correct, rather than correct-by-accident. Unbalanced braces inside a string literal skew the depth count and yield `None`, matching the module's documented fail-open posture. `_ingress_allows_public` is deliberately untouched: #196 replaces it with an anchored reader, and editing it here would conflict.

Mutation-checked: disabling the scoping filter fails exactly the new scoping tests. One of them initially passed either way — a collapsed one-line `redirect { port = "443" }` never reaches the line-anchored pattern — so it was rewritten as `terraform fmt` actually emits it to make it a real guard.

#### Depth must ignore lexically-irrelevant braces (`a9ce5b4`)

The first version of the scoping counted raw `{`/`}`, which was a **security-relevant regression** caught in review. HCL permits braces inside `#` / `//` line comments, `/* */` block comments, quoted strings, and heredoc bodies; none are block delimiters. One such brace shifted the apparent depth of a real top-level attribute and hid it — and for these rules a hidden attribute means a **missed** violation, not a conservative one.

Reproduced end to end, and the blast radius was wider than the review reported:

| body (all with a literal `publicly_accessible = true`) | before `a9ce5b4` |
|---|---|
| baseline | `rds_not_public` ✅ |
| `# primary { instance` | **0 violations** |
| `identifier = "primary { db"` | **0 violations** |
| `// primary { instance` | **0 violations** |
| `/* primary { instance */` | **0 violations** |

Beyond `rds_not_public`: an **unencrypted** RDS also lost `rds_encryption_at_rest` (its `has_attr` probe is unscoped, so the `None` read was treated as variable-driven and skipped), and a bad HTTP:80 forward listener lost `alb_http_redirect` via `_attr_int`. So narrowing the change to spare `_attr_string` would not have been sufficient — `_attr_int` was affected too.

Depth now comes from `_lex_hcl`, a single-pass scanner tracking line/block comments, quoted strings with escapes, and heredocs, so only real delimiters count. Candidates that are merely commented out are skipped as well. All five cases above now fail correctly.

The earlier `test_attr_unbalanced_brace_fails_open` **asserted the suppression was acceptable**, which was wrong: the hidden value is a plain top-level literal, so the correct reading is unambiguous and "fail open" is the wrong default when the rule's whole job is to catch it. Replaced with a guard asserting braces in comments/strings/heredocs do **not** hide an attribute, plus a `bad-rds-public-brace-in-comment` fixture that the invariant test enforces end to end.

Mutation-checked in both directions: raw brace counting fails the lexical tests *and* the fixture-invariant test; removing scoping entirely fails the nested-block tests. (Test count is stated once, in Verification below, so it cannot drift between sections again.)

Also added, from the same review: a parametrized **`good-*` → `POLICY_OK` / `bad-*` → `POLICY_FAIL`** test over the fixture directory. The invariant this description cites was previously held by convention, with all 18 fixtures wired up by hand — a new fixture added without its own test silently got no coverage. It is now enforced.

### Verification

- Test suite: **40 passed** (was 25 before this PR), 21 fixtures. Mutation-checked — reverting only the validator and re-running fails exactly the new tests, so they are real guards rather than tests written to fit the new code.
- Interpolation nesting (`2f181fd`): a `"` inside `${...}` no longer ends the outer string. Reported by @ayn-builds; it re-opened the depth-skew and truncation classes `a9ce5b4` closed. Covered in both directions plus `bad-rds-public-interpolated-brace`, which pins that *exactly* `rds_not_public` fires (the same input also produced a bogus `rds_encryption_at_rest`).
- Fixture-count assertion is now exact (`== 21`) rather than a `>=` floor, which could not detect deletion down to the floor.

**On `_attr_int` precision:** the quoted arm is strict about non-integer values, but it is looser than the table above implies — `"0443"` reads as 443, `"00"` as 0, and a long digit string parses as a Python bigint. All are inert here (the rules only compare against a fixed port list) and Terraform would itself accept `"0443"`, so this is deliberate rather than a gap; noting it because the earlier wording overstated the strictness.
- Full fixture invariant re-confirmed: every `good-*` and `internal-alb-only` fixture → exit 0 / `POLICY_OK`; every `bad-*` fixture → exit 1 / `POLICY_FAIL` with its expected rule.
- Both repros flip correctly: quoted-443 listener → `POLICY_OK`; quoted-`from_port` SG → `POLICY_FAIL / db_sg_no_public_ingress`; bare-443 control unchanged.
- `mise run build` passes clean (dprint, markdownlint, frontmatter, gitleaks, checkov 180/0, grype).

Two new fixtures, following the existing `good-*` / `bad-*` convention and marked as intentionally non-compliant:

- `good-quoted-port-https/` — correct HTTPS listener with quoted port, guards the false positive
- `bad-db-sg-public-quoted-port/` — quoted DB port open to `0.0.0.0/0`, guards the silent miss
- `bad-rds-public-brace-in-comment/` — public RDS with a brace in a comment and a string, guards the lexer regression

### Deliberately out of scope

Quoted ports inside `redirect { }` blocks (`generate-terraform.md:899`, `security-posture-rules.md:81`, `good-https-redirect` fixture) are **correct as written** — the AWS provider schema types `redirect.port` as a string. Those are untouched; changing them would be a regression.

## Type of Change

- [x] Bug fix

## Team Folder

- [x] `migrate/`

## Checklist


migrate/plugins/migration-to-aws/skills/tf-best-practices/scripts/validate-terraform-policy.py
```diff
@@ -124,8 +124,30 @@ def _attr_string(block: str, attr: str) -> str | None:
...
+    The quoted arm requires the value to be ENTIRELY digits (the closing quote is
+    anchored against the digits), so "443x", "443-444" and "${var.port}" do NOT
+    read as 443 — they return None and keep the existing fail-open behaviour for
+    non-literal / non-integer values.
+    """
+    match = re.search(
+        rf'^\s*{re.escape(attr)}\s*=\s*(?:(\d+)|"(\d+)")',
+        block,
+        re.MULTILINE,
+    )
```
leon1418: [🤖 AI review 🤖]

**Consider:** adding the quoted alternative changes *which* match wins when the same attribute name also appears in a nested block — so the description's "the bare alternative is byte-identical to the previous pattern, so existing behaviour is unchanged" does not fully hold.

`_extract_blocks` returns the whole resource body *including* nested blocks, and this is `re.search` — first match anywhere wins. A nested `redirect { port = "443" }` was previously invisible to a bare-only pattern; now it is a candidate. When the listener's own `port` is written *after* its `default_action` block — valid HCL, and `terraform fmt` does not reorder attributes relative to blocks — the nested one wins:

```hcl
resource "aws_lb_listener" "http_redirect" {
  load_balancer_arn = aws_lb.app.arn

  default_action {
    type = "redirect"
    redirect {
      port     = "443"
      protocol = "HTTPS"
    }
  }

  port     = 80
  protocol = "HTTP"
}
```

`_attr_int(body, "port")` on that returns **443** at this revision; the previous pattern returned **80**.

I traced the current rules and could not make this flip a verdict: `redirect` blocks only occur on already-compliant actions, and `_attr_string` independently misreads `protocol` as `HTTPS` on the same body, so `https_ok` and `alb_http_redirect` both skip it either way. **Not a blocker for this change.**

Flagging it for two reasons. It is a latent trap for the next port-based rule — `aws_lb_listener_rule` also carries `redirect { port }`. And the docstring's precision argument is the one part of the reasoning that no longer entirely applies: the quoted arm is strict about the *value* but the search is not scoped to the *block*.

If it is worth addressing, scoping the search to the block's own attributes (skipping nested `{ ... }`) would fix `_attr_string` too, which has the same weakness today and is where the `protocol` misread above comes from.

migrate/plugins/migration-to-aws/skills/tf-best-practices/scripts/test_validate_terraform_policy.py
```diff
@@ -254,3 +272,73 @@ def test_block_form_forward_https_listener_does_not_false_fail() -> None:
...
+        code, out = run_policy_validator(Path(tmp))
+        assert code == 1, out
+        assert "sg_no_public_admin_ingress" in out
+        assert "22 (SSH)" in out
+
+
+def test_attr_int_reads_bare_and_quoted_integers() -> None:
+    assert VALIDATOR._attr_int("  port = 443\n", "port") == 443
+    assert VALIDATOR._attr_int('  port = "443"\n', "port") == 443
+    assert VALIDATOR._attr_int('  from_port = "0"\n', "from_port") == 0
```
leon1418, line 324: [🤖 AI review 🤖]

**Consider:** these bodies are all single-attribute, so nothing here pins the behaviour when the attribute name also appears in a nested block — which is the one case where this change alters an existing reading (see the note on `_attr_int`).

Something like:

```python
assert VALIDATOR._attr_int(
    '{\n  default_action {\n    redirect { port = "443" }\n  }\n  port = 80\n}', "port"
) == 80  # or 443, whichever you decide is correct
```

would pin whichever answer you settle on. `good-quoted-port-https/compute.tf` cannot cover it, because that fixture places `port` *before* `default_action` — the ordering that happens to be safe.

**FYI:** unrelated to this change, but adjacent — there is no parametrized/globbed "every `good-*` → `POLICY_OK`, every `bad-*` → `POLICY_FAIL`" test; the 17 fixture constants are each wired up by hand. The invariant the description re-confirmed is therefore held by convention rather than enforced, so a future fixture added without its own test silently gets no coverage. You did add tests for both new fixtures, so nothing is missing here.

leon1418 (commented):
[🤖 AI review 🤖]

Reviewed the validator change, both new fixtures, and the new tests, against the current rules in `validate-terraform-policy.py` — reading the file at this revision rather than only the diff, and running `_attr_int` locally on constructed bodies to check the precision claims.

**The change is worth making and the direction is right.** The fail-closed half is the real value: `from_port = "5432"` with `cidr_blocks = ["0.0.0.0/0"]` reporting `POLICY_OK` is a validator that actively certifies a world-open database, and quoting is not an exotic way to write that. Fixing `_attr_int` rather than patching each of the three call sites is the right level to fix it at.

Verifying the new tests by mutation — revert the validator, confirm exactly the 4 new tests fail — is worth calling out. That is the check separating a test that guards behaviour from one written to fit the code, and it is skipped far more often than it is done. The 8-case negative matrix in `test_attr_int_rejects_non_literal_quoted_values` is similarly more thorough than this kind of fix usually gets.

Two non-blocking notes inline, both the same underlying point: the quoted arm interacts with nested blocks in a way the description does not account for. Neither changes a verdict under today's rules.

---
🤖 *AI-assisted review (Google eng-practices standard). Posted under @leon1418's session.* 🤖

migrate/plugins/migration-to-aws/skills/tf-best-practices/fixtures/terraform-policy/bad-db-sg-public-quoted-port/vpc.tf
```diff
@@ -0,0 +1,15 @@
+# Intentionally non-compliant fixture — DO NOT deploy, DO NOT sanitize.
+# Exercises: db_sg_no_public_ingress must still fire when from_port / to_port are
+# written as QUOTED integers (valid HCL: "5432" converts to 5432). Reading only
+# bare integers made this world-open database report POLICY_OK.
```
leon1418, line 4: [🤖 AI review 🤖]

The `DO NOT deploy, DO NOT sanitize` header earns its place. A fixture whose entire purpose is to be open to `0.0.0.0/0` is exactly the kind of file a future contributor — or an automated remediation pass — would helpfully "fix", quietly turning a guard into a test that proves nothing. Stating the intent in the file is what prevents that, and the comment also names the rule it guards, so the next reader knows what breaks if they touch it.

migrate/plugins/migration-to-aws/skills/tf-best-practices/scripts/validate-terraform-policy.py
leon1418: You're right on both counts, and I've fixed it at the root in `cc56e79`.

**Reproduced exactly as you described.** On that body, `_attr_int(body, "port")` returned `443` at the previous revision where the bare-only pattern returned `80`. And your parenthetical is the more useful half of the finding: `_attr_string(body, "protocol")` returned `HTTPS` — so the reason no verdict flipped was that **two independent misreads cancelled**, not that the reader was safe. That's not a property worth depending on, especially with `aws_lb_listener_rule` carrying a `redirect { port }`.

I'd tested the dangerous direction before opening this (a bad HTTP:80 `forward` listener with `default_action` written first, still correctly `POLICY_FAIL`) and reached the same "cannot flip a verdict" conclusion — but I checked that the *outcome* was right without noticing it was right for the wrong reason. That's the part your comment caught.

**Fix:** match candidates are now filtered by brace depth via a `_first_own_attr` helper; only those at the block's own depth are eligible (base depth 1 for a braced body, 0 for a bare attribute fragment). I took the scoping route you suggested rather than pinning the current answer, because it makes the reader mean what it says and repairs `_attr_string` in the same stroke. That listener now reads **80 / HTTP** — correct, instead of correct-by-accident.

Unbalanced braces inside a string literal skew the count and yield `None`, which I've documented and tested — fail-open, consistent with the module's posture. `${...}` interpolations are balanced, so they don't hide following attributes (also tested, since that was my main worry about depth counting).

**`_ingress_allows_public` deliberately untouched:** it has the same unanchored weakness, but #196 replaces it with an anchored `_attr_list_inner`. Editing it here would conflict, and that PR is the right place.

**On the description:** you're right that "existing behaviour is unchanged" no longer held. I've corrected it in the PR body rather than quietly dropping it, and added a section explaining the depth scoping. Suite is now 34 passed (was 29); mutation-checked that disabling the depth filter fails exactly the new scoping tests.

migrate/plugins/migration-to-aws/skills/tf-best-practices/scripts/test_validate_terraform_policy.py
leon1418: Both added.

**The nested-block pin.** Added as `test_attrs_read_from_own_block_not_nested_block`, asserting `80` — I went with fixing the scoping rather than pinning `443`, so it pins the corrected reading. It also asserts `protocol == "HTTP"` on the same body, since `_attr_string` had the identical weakness and now shares the scoping.

Your note that `good-quoted-port-https` can't cover it was the useful part — that fixture puts `port` before `default_action`, which is exactly the ordering that happens to be safe, so the fixture set structurally could not reach this. It needed to be a unit test.

**One thing worth flagging from writing it:** my first version of the nested-only test used a collapsed `redirect { port = "443" }`, like the snippet in your comment. It passed both before and after the fix — the pattern is line-anchored (`^\s*port\s*=`), and on that line `port` isn't at line start, so it never matches at all. A test that passes against the buggy code isn't a guard, so I rewrote it as `terraform fmt` actually emits it:

```hcl
redirect {
  port = "443"
}
```

That form does reach the pattern, and mutation-checking confirms it: disabling the depth filter now fails all three new scoping tests instead of two.

Also added the interpolation and unbalanced-brace edges, since depth counting introduces both.

**On the FYI — you talked me into it.** You're right that nothing was missing for this PR, but my own description claimed "full fixture invariant re-confirmed" and that claim rested on a manual sweep I ran locally. Since the value of the claim is that it keeps holding, I added `test_every_fixture_matches_its_good_bad_prefix`: it globs the fixture directory and asserts `bad-*` → `POLICY_FAIL`, everything else → `POLICY_OK`. A new fixture now gets coverage whether or not someone remembers to wire up a constant. It has a floor assertion on the directory count so the test can't silently pass by finding nothing.

migrate/plugins/migration-to-aws/skills/tf-best-practices/fixtures/terraform-policy/bad-db-sg-public-quoted-port/vpc.tf
leon1418: Thanks — and the automated-remediation-pass case is the one I was actually writing it for. Both new fixtures carry the header, and the `0.0.0.0/0` one names the rule it guards (`db_sg_no_public_ingress`) so a reader knows what breaks if they "fix" it.

Worth noting it's also load-bearing for CI here: `checkov` runs over the repo in `mise run build`, and these directories are covered by an existing `skip-path` entry for `fixtures/terraform-policy` — so the scanner is already told to leave them alone, and the header explains to humans why that exemption exists rather than leaving it looking like an oversight.

No change needed for this one.

migrate/plugins/migration-to-aws/skills/tf-best-practices/scripts/validate-terraform-policy.py
```diff
@@ -111,21 +111,79 @@ def _extract_blocks(content: str, resource_type: str) -> list[tuple[str, str, in
...
+
+    Fail-open on ambiguity, consistent with the rest of this module: an unbalanced
+    brace inside a string literal skews the count, which can make a real
+    attribute look nested and yield None — never a spurious value.
+    """
+    base = 1 if block.lstrip().startswith("{") else 0
+    for match in re.finditer(pattern, block, flags):
+        prefix = block[: match.start()]
+        if prefix.count("{") - prefix.count("}") == base:
+            return match
```
leon1418, line 308: [🤖 AI review 🤖]

Raw brace counting here introduces a security-relevant regression in every existing `_attr_string` consumer. HCL permits braces inside comments, quoted strings, and heredocs, but those characters are not block delimiters. I ran the current head with complete `aws_db_instance` resources:

```text
normal                         -> 1 rds_not_public violation
# primary { instance           -> 0 violations
identifier = "primary { db"   -> 0 violations
```

The same suppression occurs with `//` and `/* ... */` comments. Before this commit, `_attr_string` still read the literal `publicly_accessible = true` in all of those bodies. Because `_first_own_attr` is now shared by `_attr_string`, an unrelated textual brace can also hide `storage_encrypted`, `at_rest_encryption_enabled`, listener protocol, and other pre-existing security checks.

The new `test_attr_unbalanced_brace_fails_open` currently codifies this as acceptable ambiguity, but the target value is not ambiguous: it is a top-level literal hidden by unrelated lexical content. That means a known-unsafe resource can now be reported as `POLICY_OK`, which is the same class of failure this PR is intended to fix.

Please make depth tracking ignore braces inside HCL comments/strings/heredocs (a small lexical state machine or an HCL parser), or narrow this change so `_attr_string` does not regress. This needs to be addressed before merge because it creates new false negatives in existing security rules.

leon1418 (commented):
[🤖 AI review 🤖]

Re-reviewed `cc56e79`, the three replies, the updated PR description, and the latest CI results. The original nested-block misread is fixed for the intended listener case, and the added tests now pin both `port` and `protocol`; the fixture-invariant test is also a useful improvement.

There is one blocking regression in the generalized brace-depth implementation, detailed inline. I reproduced it against the current head all the way through `check_rds_not_public`: adding a legal brace character to an unrelated comment or string changes a literal `publicly_accessible = true` from one `rds_not_public` violation to zero violations.

**Nit:** the new description says the suite is now 34 tests, but the later Verification section still says “29 passed (was 25).” Please update the stale count when addressing the functional issue.

---
🤖 *AI-assisted re-review (Google eng-practices standard). Posted under @leon1418's session.* 🤖

migrate/plugins/migration-to-aws/skills/tf-best-practices/scripts/validate-terraform-policy.py
leon1418: Confirmed and fixed in `a9ce5b4`. You're right that this was blocking — thank you for catching it. Raw brace counting was the wrong primitive, and I shipped it.

**Reproduced your three cases exactly**, plus the `//` and `/* */` variants:

| body (all with a literal `publicly_accessible = true`) | before `a9ce5b4` |
|---|---|
| baseline | `rds_not_public` ✅ |
| `# primary { instance` | **0 violations** |
| `identifier = "primary { db"` | **0 violations** |
| `// primary { instance` | **0 violations** |
| `/* primary { instance */` | **0 violations** |

**The blast radius was wider than reported, in two ways worth recording:**

1. **`rds_encryption_at_rest` was also suppressed.** An `aws_db_instance` with `storage_encrypted = false` plus a comment brace produced **zero** violations. Its `has_attr` probe is a separate unscoped regex, so `has_attr` stayed true while `_attr_string` returned `None` — which the rule interprets as variable-driven and `continue`s past. An unencrypted database read as compliant.
2. **`_attr_int` was affected too.** A bad HTTP:80 `forward` listener with a comment brace lost its `alb_http_redirect` violation. So your alternative — narrowing the change so `_attr_string` doesn't regress — would have reduced the blast radius but not eliminated it. I took the lexer route.

**Fix:** depth now comes from `_lex_hcl`, a single-pass scanner tracking `#`/`//` line comments, `/* */` block comments, quoted strings (with `\\` escapes), and heredocs (`<<TAG` / `<<-TAG`, terminated by a line holding only the tag). Only real delimiters count. Interpolations are treated as opaque string content — both braces ignored, so the net depth is unaffected either way. Candidates that are merely commented out are now skipped as well, so `# port = 443` is correctly not an attribute.

All five cases above now fail correctly, and both wider-radius regressions are restored.

**On the test that blessed it — you're right, and this is the part I got wrong on my own terms.** `test_attr_unbalanced_brace_fails_open` asserted the suppression was acceptable ambiguity. Your framing is the correct one: the target is a plain top-level literal, so nothing about it is ambiguous — only my depth counter was confused. And "fail open" is not a safe default for a rule whose entire job is to catch that literal; it is the exact failure mode. I've replaced it with a guard asserting braces in comments/strings/heredocs do **not** hide an attribute, and added `bad-rds-public-brace-in-comment` so the fixture-invariant test catches it end to end rather than only at unit level.

Mutation-checked in both directions: reinstating raw counting fails the lexical tests **and** the fixture-invariant test; removing scoping entirely fails the nested-block tests. Suite is 35 passed; `mise run build` exit 0 (checkov 180/0).

**Nit fixed:** the stale "29 passed (was 25)" in Verification now reads 35, and the earlier section no longer carries its own count — one place, updated.

One thing I'd flag as pre-existing rather than fold in here: `check_rds_encryption_at_rest`'s `has_attr` probe is still an unscoped whole-body regex. It is now consistent with the scoped reader for every case I tested, but the two use different scoping rules, which is the sort of mismatch that produced the interaction in (1). Happy to align it in this PR if you'd prefer, or leave it as a follow-up — it touches a rule this PR otherwise doesn't.

migrate/plugins/migration-to-aws/skills/tf-best-practices/scripts/validate-terraform-policy.py
```diff
@@ -111,21 +111,173 @@ def _extract_blocks(content: str, resource_type: str) -> list[tuple[str, str, in
...
+                    state = "code"
+                i += 1
+        else:  # heredoc — body is literal until a line holding only the tag
+            end = block.find("\n", i)
+            line_end = n if end == -1 else end + 1
+            for j in range(i, line_end):
+                depths[j] = depth
+            if block[i:line_end].strip() == tag:
+                state = "code"
+            i = line_end
```
leon1418, line 200: [🤖 AI review 🤖]

A regular heredoc (`<<EOT`) requires an **unindented** terminator; only an indented heredoc (`<<-EOT`) permits leading spaces. `.strip() == tag` treats both forms as indentation-tolerant, so this exits a regular heredoc early on body text such as `  EOT`.

I ran the current head with this valid shape:

```hcl
description = <<EOT
  EOT
  publicly_accessible = false
EOT
publicly_accessible = true
```

The indented `EOT` is literal heredoc content, as is the following `publicly_accessible = false`. The lexer instead marks that false value as code at base depth; `_attr_string` returns `false` before reaching the real top-level `true`, and `rds_not_public` reports **zero violations**.

Please preserve whether `_HEREDOC_OPEN` matched `<<` or `<<-` and only apply indentation tolerance to the latter. A regression test should include a regular heredoc with an indented tag-looking line followed by attribute-looking body text; the existing heredoc test uses a clean unindented terminator and cannot catch this.

leon1418 (commented):
[🤖 AI review 🤖]

Re-reviewed `a9ce5b4` and the response in r3717903276. The lexer fixes every case reported in the previous round, the replacement tests now assert the correct safety property, the new end-to-end fixture is valuable, and the stale test count is fixed. CI is green.

Three blocking gaps remain in the parsing chain:

1. **`_extract_braced_block` still counts raw braces before `_lex_hcl` ever runs.** A closing brace in a legal comment, string, or heredoc truncates the resource body before `publicly_accessible = true`. Against this head:

```text
normal resource                  -> 1 rds_not_public violation
# primary } instance             -> 0 violations
identifier = "primary } db"     -> 0 violations
heredoc line containing }        -> 0 violations
```

The new fixture only guards opening `{`, which makes extraction run to EOF; it does not cover premature truncation on `}`. The extractor needs the same lexical awareness (ideally one scanner shared by extraction and attribute scoping), plus an end-to-end closing-brace fixture.

2. **The unscoped `has_attr` probes are now a regression, not merely a pre-existing cleanup.** With the new scoped `_attr_string`, a standalone attribute inside `/* ... */` is correctly ignored by the reader but still matched by `has_attr`; `val is None` is then treated as variable-driven. I reproduced both:

```text
RDS with only block-commented storage_encrypted = false       -> 0 violations
ElastiCache with only block-commented at_rest_encryption=false -> 0 violations
```

Both attributes are actually absent, and both services therefore fail the rule. Before scoped reading, those comment lines were read as literal false and happened to produce violations; this PR changes them into false negatives. Please align both presence probes with the lexical/scoping rules in this change rather than defer them.

3. The regular-heredoc terminator issue is inline.

These are all the same architectural boundary: once the change introduces lexical scope as the source of truth, extraction, attribute presence, and attribute values need to consume that same truth. Otherwise independently reasonable helpers disagree at exactly the security decisions this validator exists to make.

---
🤖 *AI-assisted re-review (Google eng-practices standard). Posted under @leon1418's session.* 🤖

migrate/plugins/migration-to-aws/skills/tf-best-practices/scripts/validate-terraform-policy.py
leon1418: Confirmed and fixed in `2bf671b`.

Reproduced your exact shape. On that body `_attr_string(body, "publicly_accessible")` returned **`'false'`** — the value from inside the heredoc — instead of the real top-level `true`, and `rds_not_public` reported zero violations.

You're right about the HCL rule: `<<TAG` requires the terminator at column 0; only `<<-TAG` permits leading whitespace. `_HEREDOC_OPEN` now captures the `-` as its own group and the lexer keeps a `tag_indent_ok` flag, so the terminator test is `line.strip() == tag` only for `<<-`, and `line.rstrip() == tag` for plain `<<` — which rejects `  EOT` as body text while still allowing a trailing newline.

Both directions are now pinned in `test_regular_heredoc_terminator_must_be_unindented`, using your shape for the regular form (indented tag-looking line followed by attribute-looking body text, asserting `80`) and an indented terminator for `<<-` so the fix can't over-correct and break the tolerant form. Mutation-checked: making both forms tolerant again fails exactly that test.

leon1418:
All three gaps fixed in `2bf671b`. Your framing was the right one — I'd been patching readers one at a time while extraction and presence kept their own notion of what counts as code, and they disagreed exactly where the rules make security decisions. Addressed as one boundary rather than three symptoms.

### 1. `_extract_braced_block` raw counting

Reproduced. One note on the numbers: my first attempt showed exit=1, but that was the *isolated* case being masked — `rds_encryption_at_rest` fired on the truncated body for an unrelated reason. Ordering `storage_encrypted` before the stray brace isolates it and reproduces your figure exactly:

| body (literal `publicly_accessible = true`) | before `2bf671b` |
|---|---|
| baseline | `rds_not_public` ✅ |
| `# capacity review } pending` | **0 violations** |
| `identifier = "primary } db"` | **0 violations** |
| heredoc body containing `}` | **0 violations** |

`_extract_braced_block` now consumes `_lex_hcl`, and `_extract_blocks` performs the scan **once per file** and passes it down — so extraction and attribute scoping share one scan rather than one algorithm. Went slightly further on the same principle: a commented-out `resource "aws_db_instance" ... {` is no longer extracted, and both inline-ingress walkers now go through a shared `_own_blocks` helper instead of their own `re.finditer` + raw extraction.

Added `bad-rds-public-closing-brace` as the end-to-end fixture you asked for — comment, string, **and** heredoc closing braces in one resource.

### 2. `has_attr` probes

You're right that this is a regression and not a pre-existing cleanup, and I was wrong to offer to defer it. My reasoning was that the probe predates this PR; what I missed is that its *behaviour* only becomes wrong once value reading is scoped. Before, the reader returned the commented `false` and the violation fired by accident. Reproduced both:

```
RDS, storage_encrypted = false only inside /* */          -> 0 violations
ElastiCache, at_rest_encryption_enabled = false in /* */  -> 0 violations
```

In both, the attribute is genuinely absent, so the service is unencrypted by default and the rule should fire. Replaced both probes with `_has_own_attr`, which routes through `_first_own_attr` — presence and value now share one rule by construction. Both now fail correctly.

`test_attribute_presence_probe_matches_value_scoping` pins the agreement in both directions: commented-only → absent for *both* helpers; `= var.encrypt` → present-but-unreadable, so the variable-driven fail-open path still works and I haven't broken the intended behaviour.

### 3. Heredoc terminator — [replied inline](https://github.com/awslabs/startups/pull/197#discussion_r3717989027)

### Verification

- **39 passed** (was 25 before this PR); 20 fixtures.
- **Mutation-checked each fix independently** — reverting one at a time fails exactly its own guard: raw-brace extraction → the truncation test **and** the fixture-invariant test; always-tolerant heredoc → the terminator test; unscoped probe → the presence-agreement test.
- `mise run build` exit 0 (checkov 180/0).
- Also confirmed `<<-EOT` still tolerates an indented terminator, so the heredoc fix doesn't over-correct.

One remaining asymmetry I'll name rather than leave implicit: `_iam_key_is_wildcard` and `_ingress_allows_public` still use unanchored whole-body regexes. Neither is reachable by the reported failure modes (`_ingress_allows_public` is replaced by an anchored reader in #196, and the IAM reader deliberately scans literal JSON where brace depth is not a scoping signal), but they are the two places left that don't consume the lexer. Happy to bring them in here if you'd rather the boundary be complete in one PR.

ayn-builds (commented):
Reviewed at head 2bf671b: reproduced the before/after behaviour locally, ran the suite (39 pass, note the description says 35), markdownlint 0 errors, dprint check clean.

The diagnosis and all four fixture claims hold up. I verified each against main: good-quoted-port-https goes POLICY_FAIL -> POLICY_OK, bad-db-sg-public-quoted-port goes POLICY_OK -> POLICY_FAIL, and bad-rds-public-closing-brace goes POLICY_OK -> POLICY_FAIL. The quoted-integer fix is correct, the precision table is accurate, and moving the presence probes onto the shared lexer is a real improvement beyond the stated scope.

One blocking item: the lexer added in a9ce5b4 has a gap that re-opens the same class of missed violation that commit set out to close, reachable from ordinary HCL. Details inline, with a prototyped fix. Also flagging a merge conflict with #196 that needs a decision on ordering, plus non-blocking notes. Filing as a comment rather than request-changes.

migrate/plugins/migration-to-aws/skills/tf-best-practices/scripts/validate-terraform-policy.py
```diff
@@ -83,49 +83,254 @@ def _read_tf_files(terraform_dir: Path) -> list[tuple[str, str]]:
...
+                state = "code"
+                i += 2
+            else:
+                i += 1
+        elif state == "string":
+            if ch == "\\" and i + 1 < n:
+                depths[i + 1] = depth
+                i += 2
+            else:
+                if ch == '"':
```
ayn-builds: **Blocking: a quote nested inside `${...}` drops the lexer back into code state, which re-opens the exact failure class `a9ce5b4` closed.**

In the `string` state a `"` unconditionally sets `state = "code"`. Inside an interpolation, the *inner* string's closing quote therefore ends the outer string as far as the lexer is concerned. Everything after it is scanned as code, so braces in string content count toward depth again and `is_code` goes `True` for text that is not code.

Traced on `identifier = "${lookup(var.m, "a{b")}"`: depth at the following `publicly_accessible` is **2**, not 1, so `_first_own_attr` rejects it as nested. This is the same "lexically-irrelevant brace shifts a real top-level attribute out of scope" mechanism the description says is fixed, reached by a different route.

End to end, a public RDS is missed on this PR and caught on `main`:

```hcl
resource "aws_db_instance" "bad" {
  identifier          = "${lookup(var.names, "primary{db")}"
  instance_class      = "db.t3.micro"
  publicly_accessible = true
  storage_encrypted   = true
}
```

| | verdict |
|---|---|
| `main` | `POLICY_FAIL`, `rule=rds_not_public` |
| this PR | `rds_not_public` **missing**; instead a **false** `rds_encryption_at_rest` (`storage_encrypted = true` is right there) |

So it is both a missed exposure and a bogus finding on the same resource, which is worse than the pre-existing behaviour on this input.

It also truncates bodies, not just skews depth, so `_extract_braced_block` is affected as well. With a `}` in the nested string:

```hcl
resource "aws_db_instance" "pg" {
  n = "${replace(v, "}", "")}"
  publicly_accessible = true
}
```

`_attr_string(body, "publicly_accessible")` returns `None` - the body is cut before the attribute.

Reachable from ordinary HCL. `templatefile("t.tpl", {"a"="{"})` and `replace(v, "{", "")` both trigger it. Interpolations whose nested strings carry no brace are fine (`lookup(var.tags, "Name")`, `join(",", ...)`, `format("%s-%s", ...)`, `jsonencode({"a"="b"})` all still read correctly), which is why the suite passes - a brace inside the nested string is the trigger, and no test covers that.

The docstring at line 99 states the intended behaviour, so this reads as an implementation gap rather than a design decision:

> Interpolations are treated as opaque string content: `{` and `}` inside a string are both ignored, so the running depth is unaffected either way.

Fix is to track interpolation nesting so a quote inside `${...}` re-enters string state instead of code state. I prototyped a counter incremented on `${`, decremented on the matching `}`, with quotes skipped as inner strings while the counter is above zero: depth at `publicly_accessible` goes back to 1 on all the cases above, and the existing lexer tests still hold. Worth adding cases for a brace inside a nested quote in both directions, since that is the specific gap.

migrate/plugins/migration-to-aws/skills/tf-best-practices/scripts/validate-terraform-policy.py
```diff
@@ -350,8 +555,7 @@ def check_db_sg_no_public_ingress(tf_files: list[tuple[str, str]]) -> list[Viola
     for rel_path, content in tf_files:
         for name, body, line in _extract_blocks(content, "aws_security_group"):
             # Walk each inline ingress block via brace matching.
-            for m in re.finditer(r"ingress\s*\{", body):
-                ingress_body, _ = _extract_braced_block(body, m.end() - 1)
+            for ingress_body in _own_blocks(body, "ingress"):
```
ayn-builds, line 579: **Merge conflict with #196, which rewrites these exact two loops.**

#196 replaces `_ingress_allows_public` with `_ingress_public_cidrs` at both call sites; this PR replaces the surrounding `re.finditer` + `_extract_braced_block` walk with `_own_blocks`. I merged the two heads locally and git cannot auto-resolve: two `CONFLICT (content)` regions, at the `db_sg_no_public_ingress` and `sg_no_public_admin_ingress` loops.

The resolution is mechanical - take `_own_blocks` from here and the `public = _ingress_public_cidrs(...)` body from #196 - but whichever merges second needs a rebase, and the combination is not covered by either suite. Worth deciding merge order explicitly and having the second PR rebase, rather than resolving it in the merge commit untested.

Your note that `_ingress_allows_public` is deliberately untouched to avoid conflicting with #196 is the right instinct; the conflict just lands one line up instead.

migrate/plugins/migration-to-aws/skills/tf-best-practices/scripts/validate-terraform-policy.py
```diff
@@ -396,8 +600,7 @@ def check_sg_no_public_admin_ingress(tf_files: list[tuple[str, str]]) -> list[Vi
     violations: list[Violation] = []
     for rel_path, content in tf_files:
         for name, body, line in _extract_blocks(content, "aws_security_group"):
-            for m in re.finditer(r"ingress\s*\{", body):
-                ingress_body, _ = _extract_braced_block(body, m.end() - 1)
+            for ingress_body in _own_blocks(body, "ingress"):
```
ayn-builds, line 624: Non-blocking observations, in descending order of importance.

**Both encryption rules' `has_attr` probes now share the lexer, which is the right call** - `_has_own_attr` agreeing with `_attr_string` about what counts as set is a genuine correctness improvement over the old unscoped `re.search`, and `test_attribute_presence_probe_matches_value_scoping` pins exactly the right invariant. Same for `_own_blocks` skipping a commented-out `ingress {`. Neither was in the stated scope of the PR and both are worth having.

**Description says 35 tests; actual is 39.** Presumably drifted across the review commits.

**`test_every_fixture_matches_its_good_bad_prefix` is a good addition but its count assertion is weak.** `assert len(fixture_dirs) >= 18` with 20 present cannot detect fixtures being removed down to 18. An equality check against the expected count, or dropping the assertion, would make the intent clearer.

**`_first_own_attr` re-lexes its block on every call, with no caching.** Each `_attr_int` / `_attr_string` / `_has_own_attr` is a fresh O(n) scan of the block. On a synthetic 15KB SG with 300 ingress blocks I measured ~14ms for the 600 `_attr_int` calls, plus ~5ms each for `_extract_blocks` and `_own_blocks` - fine at fixture scale and not worth restructuring now, but the lexed result is already threaded through `_extract_braced_block` via the `lexed` parameter, so the same could be passed down here if it ever matters.

**`_attr_int` accepts leading zeros and unbounded values**: `"0443"` reads as 443, `"00"` as 0, and a 20-digit string parses as a Python bigint. None of these are exploitable given the rules only compare against a fixed port list, and rejecting them could plausibly be wrong (Terraform would accept `"0443"`), so this is just a note that the quoted arm is looser than the precision table in the description suggests.

**Docs are unchanged.** Reasonable here, since none of this alters the documented rule semantics - only what the reader can see. Flagging it only because #196 does touch `SKILL.md` and `references/security-posture-rules.md`, so the two PRs differ in that respect.

migrate/plugins/migration-to-aws/skills/tf-best-practices/scripts/validate-terraform-policy.py
leon1418: Confirmed and fixed in `2f181fd`. You're right that this re-opens the class `a9ce5b4` closed, and the mechanism is exactly as you traced it.

**Reproduced end to end on your resource:**

| | `rds_not_public` (real) | `rds_encryption_at_rest` (bogus) |
|---|---|---|
| `main` | fires | — |
| `2bf671b` | **missing** | **falsely fires** |
| `2f181fd` | fires | — |

So it was both a missed exposure and an invented finding on the same resource, worse than `main` on this input — that framing is right, and it is the part that made me treat this as blocking rather than an edge case. The truncation half reproduces too: with `n = "${replace(v, "}", "")}"`, `_attr_string(body, "publicly_accessible")` returned `None` on `2bf671b` and returns `'true'` now.

**Fix:** took your prototype's shape. `${` increments an interpolation counter, the matching `}` decrements it, and while the counter is above zero a `"` toggles an inner-string flag rather than ending the outer string. Depth is still unaffected by anything inside a string, which is what the docstring claimed all along — this makes the implementation match the documented intent instead of the reverse.

**On why the suite passed.** Your diagnosis was the useful part: the benign interpolations are the ones I had tested. I re-verified each of the four you listed — `lookup(var.tags, "Name")`, `join(",", ...)`, `format("%s-%s", ...)`, `jsonencode({"a"="b"})` — all still read correctly, and none of them would ever have caught this, because a brace *inside the nested string* is the trigger and no test had one. That is a coverage-shape gap, not a volume gap; more tests of the same shape would not have found it.

**Tests, both directions as you asked:**
- `test_quote_nested_in_interpolation_does_not_end_the_string` — opening brace (depth skew), closing brace (truncation), both braces, a nested quote with no brace, and two interpolations on one line.
- A nested-quote case added to the existing truncation guard, so `_extract_braced_block` is covered as well as depth.
- `bad-rds-public-interpolated-brace`, so the fixture invariant covers it end to end and pins that *exactly* `rds_not_public` fires — the bogus encryption finding is asserted away, not just the missing one asserted back.

Mutation-checked against `2bf671b`: both the new test and the extended truncation guard fail there.

Deep nesting beyond one level of quoted string inside an interpolation is still not modelled, and I would rather say so than imply the counter is a full expression parser. Every shape in your comment and in the suite is covered; a construct like `"${f("${g("a{b")}")}"` is not, and I did not add a test asserting behaviour I have not thought through properly.

**Non-blocking items from your other comment, in this commit:** the fixture count assertion is now exact (`== 21`) — you were right that a `>=` floor cannot detect deletion down to the floor, which is the removal it exists to catch. The remaining notes (no `_first_own_attr` caching, `_attr_int` accepting `"0443"` and unbounded ints, docs unchanged) I have left alone deliberately: the first is a measured non-issue at fixture scale, and the other two are cases where changing behaviour could plausibly be the wrong call, so I would rather not fold them into a fix commit. The stale test count in the description is corrected there.

migrate/plugins/migration-to-aws/skills/tf-best-practices/scripts/validate-terraform-policy.py
leon1418: Confirmed, and agreed on deciding the order explicitly rather than resolving it in a merge commit. I re-ran the merge at the current heads (`cc8cac8` for #196, `2f181fd` here) and validated a resolution end to end.

**Conflict:** 2 regions, one file, exactly where you said — the `db_sg_no_public_ingress` and `sg_no_public_admin_ingress` ingress loops.

**Resolution is mechanical, and I verified it rather than assuming:** take `_own_blocks(body, "ingress")` from here and the `public = _ingress_public_cidrs(ingress_body)` body from #196. `_ingress_allows_public` disappears entirely (#196 deletes it), so there is no leftover caller.

**One step beyond the two regions, which git will not flag:** the fixture-count assertion I just made exact (`== 21`, per your other comment) has to be bumped to `23`, because #196 adds two fixtures. Without that the merged suite fails on `test_every_fixture_matches_its_good_bad_prefix` — I hit it. Arguably the assertion doing its job, but worth naming so it is not mistaken for a real regression during the rebase. With both applied: **49 pass**.

**Suggested order: this PR (#197) first, then #196 rebases.** Two reasons, and the second is the one I actually care about:

1. #197 is the deeper change and further along in review; #196 is additive on top of it.
2. #196 currently carries its own comment-stripping regex (`_HCL_COMMENT`) to fix a false positive where a CIDR named in a comment inside the list counted as an allowed range. That is a second, weaker notion of "what counts as code" sitting next to `_lex_hcl` — precisely the disagreement-between-helpers problem your earlier round on this PR was about. If #197 lands first, #196's rebase can route `_attr_list_inner` through the shared lexer and delete the duplicate. The other order leaves that cleanup as a follow-up nobody is obliged to do.

To be clear about what I have and have not validated: the mechanical resolution plus count bump is tested (49 pass). Routing `_attr_list_inner` through `_lex_hcl` is *not* — that is the follow-up I am proposing, not something already proven, and I would do it as its own commit on the rebase so it is reviewable separately from the conflict resolution.

The combination is now covered, at least for the case that needs both PRs: a quoted port with a non-canonical `::0/0` is `POLICY_OK` on `main`, on #196 alone, and on #197 alone, and only `POLICY_FAIL` with both applied.

migrate/plugins/migration-to-aws/skills/tf-best-practices/scripts/validate-terraform-policy.py
leon1418: Thanks — taking these in order.

**Fixture count assertion — fixed in `2f181fd`.** You were right; a `>=` floor cannot detect deletion down to the floor, which is the removal it exists to catch. Now an exact `== 21` with the directory names in the failure message. Noting the trade I accepted: an exact count means whichever of #196/#197 merges second must bump it (to `23`), which I hit when validating the merge resolution and have flagged in my reply on that thread.

**Test count in the description — fixed.** It read 35, actual was 39, and is now 41 after this commit's additions. I have moved the count to a single place so the same drift cannot recur across review rounds.

**`_first_own_attr` re-lexing per call — leaving as is.** Your measurement matches mine and the conclusion is the same: ~14ms for 600 `_attr_int` calls on a synthetic 15KB SG is not worth restructuring for, and the `lexed` parameter is already threaded through `_extract_braced_block` if it ever becomes worth it. Not touching it in a fix commit.

**`_attr_int` accepting `"0443"` and unbounded values — leaving as is, and you are right that the description overstates it.** Terraform *would* accept `"0443"`, so rejecting it could plausibly be the wrong behaviour, and the rules only ever compare against a fixed port list so a bigint is inert. I have softened the precision claim in the description rather than tighten the code, since the looseness is deliberate and the inaccurate part was the prose.

**Docs unchanged — deliberate, and I agree with your framing of why it differs from #196.** Nothing here changes what any documented rule *means*; it changes what the reader can see, which is an implementation property. #196 does change rule semantics — it adds an address family and, after review, canonicalisation and comment scoping — so it owes `SKILL.md` and `references/security-posture-rules.md` an update and has one. If you would rather the lexer's guarantees be written down somewhere reviewable regardless, the natural home is a short note in `references/terraform-validation.md` about the fail-open boundary; say the word and I will add it here instead of leaving it implicit in docstrings.

**On the two improvements beyond stated scope** (`_has_own_attr` sharing the lexer, `_own_blocks` skipping a commented-out `ingress {`) — thank you for calling out that they were worth having rather than treating them as scope creep. They only exist because your earlier round reframed three symptoms as one boundary, which is what made it obvious that presence and value had to consume the same truth.

leon1418: Correction to my reply above: the suite is **40**, not 41 — I wrote the wrong number in a message about fixing a wrong number. The description now states 40 in one place (Verification), with the duplicate count in the earlier section removed so the two cannot drift apart again.

ayn-builds (approved):

PR #198, opened 2026-08-04
feat(agent-advisor): standalone bundle for a headless, single-skill executor (stacked on #195)

leon1418:
**Stacked on #195.** Until that merges, the diff here also shows #195's commit; the only commit this PR adds is `dcb0e80` (23 files, +1529, no deletions). Review #195 first.

## Why

The plugin keeps the full `agent-advisor` skill, including the Migration Plan stage that reads the sibling `gcp-to-aws` engine. An executor that installs **one** skill has no siblings to read, so the variant published there has to be self-contained.

Rather than fork the prose — which drifts, as this repo has learned more than once — the bundle is a **strict subset** of the canonical skill plus exactly one generated file:

```
transformation_definition.md = preamble.md + SKILL.md (frontmatter stripped)
```

Every other file in the bundle is byte-identical to its canonical source, so there is no second copy of any instruction to keep in sync.

## What's here

| Path | Role |
| --- | --- |
| `atx/agent-advisor/preamble.md` | The only environment-specific text: no interactive tools (values come from a run seed, then repository prose, then `UNANSWERED.md`), no subagent dispatch (Estimate runs inline), the `${CLAUDE_PLUGIN_ROOT}` fallback, engine-produced scoring, optional MCP, and the required validation-summary format |
| `atx/agent-advisor/manifest.txt` | Whitelist of canonical files that ship. A canonical file that is neither listed nor matched by an exclusion rule fails the check — new files must be classified, never silently dropped |
| `tools/atx-bundle/build.ts` | Zero-dependency builder + verifier (`mise run atx:build` / `atx:check`) |
| `fixtures/agent-advisor-anthropic-demo/` | Target repository for a headless run: an existing Anthropic Messages-API agent, a seed under `seeds/`, and its own `check_recommendation.py` |

`atx:check` is in the `lint` chain, which makes **"this bundle does not depend on `gcp-to-aws`"** a machine-verified property rather than a claim. It checks that the manifest and the canonical tree agree, that no shipped file resolves a path into another skill, that the capability gate exists, that the phase graph closes inside the shipped set, and that no runtime script needs a third-party import at module load.

## Behaviour difference, stated plainly

The Migration Plan stage resolves to `not_applicable` when the sibling engine is absent, with a note to the user. **The POC still runs** — design-backed rather than plan-backed, and labelled as such — because losing the plan should not cost the user the deliverable they most often want.

## The fixture, and what the seed bought

`seeds/README.md` records the measurement rather than asserting the property. Six headless runs over the fixture with the same seed produced a **byte-identical `scoring-result.json`** (sha256 `0fcf9967…`). Getting there surfaced three drift causes — a seeded `region` object reshaped into sibling keys, two source-scan false positives inflating the findings counts, and `source_paths` listing evidence files as call sites. All three are fixed in the skill prose (in #195), not pinned around in the seed, and the README shows the before/after per run.

## Verification

`mise run lint` clean (`atx:check` 68 files / 12 warnings, `shared:check`, `fixtures:check` 98 json, frontmatter, 57 node tests), 301 pytest, and the fixture's own 3 tests.


leon1418:
Closing to restructure. Two problems with this PR as it stands:

1. `fixtures/agent-advisor-anthropic-demo/` does not belong here. The existing fixtures in that directory are replay captures (JSON snapshots plus `check_expected_*.py` assertions, ~30K each); this one is a runnable demo application with `src/`, `uv.lock`, `.env.example` and `run-claude.sh`. Nothing in this repo consumes it — no test, no lint task, no tool — so it added review surface and maintenance cost (a `.venv` trap that needed a bandit exclusion, a lockfile for dependency scanners) without anything public depending on it. It moves to the internal package that owns the benchmark.

2. The four commits here are not a reviewable sequence — two of them fix problems the first one introduced.

Will reopen as a single commit carrying only the bundle: the preamble, the manifest, the builder, and the two mise tasks.

PR #199, opened 2026-08-04
fix(heroku-to-aws): replace 0.0.0.0/0 interim RDS exposure with a scoped allowlist

leon1418:
## Problem

When `migration_approach == "interim_cutover_data_first"`, the generated customer-facing migration docs (`skills/heroku-to-aws/references/phases/generate/generate-docs.md:248-273`) instructed the user to hand-edit their database to:

- make RDS publicly accessible,
- **allow inbound 5432 from `0.0.0.0/0`** — justified in prose as "required for Heroku's dynamic IPs",
- set `rds.force_ssl = 1` as the compensating control.

A customer following these steps converts a private, encrypted database into one open to the entire internet on 5432, with the only compensating control sitting as a separate, skippable step in the same prose. `interim_cutover_data_first` is a live branch of the clarify phase (`clarify-interview.md:376,509`), so this guidance really does ship.

Three aggravating factors:

1. **No Terraform counterpart.** `grep -c 'interim_cutover\|publicly_accessible\|force_ssl' generate-terraform.md` → **0**, across 1,825 lines. The generated `aws_db_instance` is private-subnet with `storage_encrypted = true` and the database SG allows 5432 only from the app SG — so the prose asked for a console hand-edit that contradicts the IaC.
2. **The exposure was invisible to code review.** Because the change was made in the console, the next `terraform apply` reverts it. The open window existed only in the running account, never in Git — you could not audit when it opened or whether it closed.
3. **It reads as authoritative.** The instruction is explicit and justified-sounding, which invites compliance rather than scrutiny.

Separately, the old guidance was also **non-functional**: AWS requires that the subnets in a DB subnet group have an internet gateway for a DB instance to be publicly accessible. The generator places the database in *private* subnets, so flipping `publicly_accessible` in the console would not actually have made it reachable.

### 1. `0.0.0.0/0` is gone; every path ends in a bounded allowlist

The section is now *"Interim Database Exposure (TLS Prerequisite, Then a Scoped Allowlist)"*, opening with an explicit prohibition, then three connectivity paths in descending preference:

- **Path A (preferred) — Private Space VPC peering.** Cedar-generation Private Spaces; traffic does not traverse the public internet, so the database stays private. Allowlist is the Space's dyno CIDRs.
- **Path B — Private Space stable outbound IPs.** Cedar *and* Fir; allowlist ~4 `/32`s from `heroku spaces:info`.
- **Path C — Common Runtime + static-egress SOCKS5 add-on** (QuotaGuard Static, Fixie Socks); allowlist the proxy's `/32`s.

Closing with "**There is no fourth path**" — if no bounded allowlist can be enumerated, the guidance says switch to `full_cutover` rather than widen the rule. Paths B/C also disclose the internet-gateway requirement above, which is a further argument for Path A.

> Note: Heroku documents no PrivateLink capability for Private Spaces — only Private Space Peering and IPSec VPN, both Cedar-only. No PrivateLink path was invented.

### 2. TLS enforcement is now a structural prerequisite, not a parallel step

`rds.force_ssl = 1` is emitted by Terraform (`aws_db_parameter_group`, `apply_method = "pending-reboot"`) so there is no console step and nothing to skip. Step 1 ends in a **verification gate**: `pg_hba_file_rules` must show `hostssl`, and an `sslmode=disable` connection must be rejected. If either check fails the reader is told to stop and open no network path. `DATABASE_URL` now uses `sslmode=verify-full` (with a note that `require` encrypts but does not authenticate the server).

### 3. The interim exposure is now IaC-visible

Instead of a console hand-edit, the docs direct the change through Terraform, and the generator emits the gated resources to make that possible — closing aggravating factor 2:

- `dynamic "ingress"` on `aws_security_group.database`, `for_each = length(var.interim_heroku_ingress_cidrs) > 0 ? [1] : []`
- `variable "interim_heroku_ingress_cidrs"` — `default = []`, with a `validation` block whose condition is `!contains(var.interim_heroku_ingress_cidrs, "0.0.0.0/0")`
- `variable "interim_db_public_access"` — `bool`, `default = false`; only Paths B/C need it true
- commented `terraform.tfvars.example` entries, and Step 4 covers closing the path at cutover

Both variables **default to closed**, so the interim block emits nothing unless deliberately populated. All insertions use the file's existing in-HCL comment-gate style (`# {{IF ...}}` / `# {{ENDIF}}`, as at line 513).

`clarify-interview.md` prose that promised "your RDS instance to be publicly accessible" was corrected to match; no new interview questions were added.

### Verification

- **`terraform plan` proves the guard is real:** `interim_heroku_ingress_cidrs = ["0.0.0.0/0"]` → *"Planning failed … Error: Invalid value for variable"*; `["203.0.113.10/32"]` → accepted. (`terraform validate` alone does not evaluate variable validations, so it was checked at plan time.) `terraform fmt -check` clean on the added HCL.
- **Policy validator** (`skills/tf-best-practices/scripts/validate-terraform-policy.py`): the emitted template → `POLICY_OK`; a counterfactual of the *old* guidance → `POLICY_FAIL` on both `rds_not_public` and `db_sg_no_public_ingress`.
- `mise run build` → exit 0 (dprint clean, markdownlint 805 files / 0 errors, frontmatter, `shared:check`, `fixtures:assert`, gitleaks, checkov 180/0, grype).
- `{{IF}}`/`{{ENDIF}}` balance: generate-docs 43/43, generate-terraform 9/9.
- No `0.0.0.0/0` remains on any database-port path. Remaining hits in this file are the new **prohibitions** themselves; elsewhere they are all-outbound `egress` rules and public ALB 80/443.
- The doc-verification checklist was updated to match the restructured section, and gained a regression assertion: *"Does NOT instruct opening a database port (5432) to `0.0.0.0/0` anywhere."*

### Known follow-up (not in this PR)

The new guidance branches on the reader's Heroku runtime, which clarify does not plumb through: `network.private_space_detected` exists in `clarify-assemble.md:62` but is never surfaced to generate, and space generation (Cedar vs Fir, which decides Path A availability) is not captured. Rather than expand scope with new interview questions, the docs branch reader-side (`heroku spaces:info` tells them everything) and CIDRs arrive via `terraform.tfvars` at apply time. Plumbing `private_space_detected` into generate-docs Step 0 to pre-select the path would be a good follow-up.

## Type of Change

- [x] Bug fix
- [x] Documentation update

## Team Folder

- [x] `migrate/`

## Checklist


migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/phases/generate/generate-docs.md
```diff
@@ -657,8 +752,8 @@ After successful verification (recommend 48–72 hours of parallel running):
 
 Once your application is fully running on AWS (no longer connecting from Heroku):
 
-- [ ] **Disable public access on RDS/Aurora:** Set the database instance to "Not publicly accessible"
-- [ ] **Restrict security groups:** Remove any `0.0.0.0/0` inbound rules; allow only VPC-internal traffic on database ports
+- [ ] **Disable public access on RDS/Aurora:** In Terraform, set `interim_db_public_access = false` and apply; confirm the database reports "Not publicly accessible"
```
leon1418: [🤖 AI review 🤖]

This "Post-Migration Lockdown" section renders unconditionally (outside any `{{IF migration_approach == "interim_cutover_data_first"}}` block), but the two updated checklist items now reference Terraform variables (`interim_db_public_access`, `interim_heroku_ingress_cidrs`) that only exist in the `interim_cutover_data_first` template path.

For a `full_cutover` migration the generated guide will instruct the reader to reset variables that don't exist in their Terraform. The mainline wording ("Set the database instance to 'Not publicly accessible'", "Remove any `0.0.0.0/0` inbound rules") was generic enough to apply to both paths.

Suggested fix: wrap the first two checklist items in `{{IF migration_approach == "interim_cutover_data_first"}}` / `{{ENDIF}}` and keep the old generic wording in an `{{ELSE}}` branch (or unconditionally, since the generic instructions are always safe).

leon1418 (commented):
[🤖 AI review 🤖]

Reviewed all 3 changed files end-to-end (clarify-interview.md, generate-docs.md, generate-terraform.md). The security improvement is excellent — replacing an open-internet `0.0.0.0/0` instruction with a structured, TLS-first, bounded-allowlist approach backed by IaC is a clear win for code health.

**1 blocking finding:** The Post-Migration Lockdown checklist now references `interim_db_public_access` and `interim_heroku_ingress_cidrs` unconditionally, but these variables only exist when `migration_approach == "interim_cutover_data_first"`. A `full_cutover` migration will render nonsensical instructions referencing non-existent Terraform variables.

**1 Nit:** The `interim_heroku_ingress_cidrs` validation only rejects the literal `"0.0.0.0/0"` — a split like `0.0.0.0/1` + `128.0.0.0/1` would bypass it. In practice the documentation is extremely clear about intent, and accidental bypass is implausible, so this is informational only.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/phases/generate/generate-docs.md
leon1418: Confirmed and fixed in `c8a415c`. This was a regression I introduced, and your diagnosis of the mechanism is exact.

**Verified the section really is unconditional.** I walked the `{{IF}}`/`{{ENDIF}}` nesting to those exact lines rather than eyeballing it:

```
L751 ### Post-Migration Lockdown   -> UNCONDITIONAL
L755 interim_db_public_access      -> UNCONDITIONAL
L756 interim_heroku_ingress_cidrs  -> UNCONDITIONAL
```

Meanwhile every declaration of those variables in `generate-terraform.md` sits behind `# {{IF migration_approach == "interim_cutover_data_first"}}`. So a `full_cutover` reader was told to reset variables that don't exist in their config — and `terraform apply` would reject an undeclared variable, so it isn't even a no-op.

And you're right that this was my regression, not a pre-existing wart. The diff against `upstream/main` shows the previous wording was deliberately path-agnostic:

```diff
-- [ ] **Disable public access on RDS/Aurora:** Set the database instance to "Not publicly accessible"
-- [ ] **Restrict security groups:** Remove any `0.0.0.0/0` inbound rules; ...
```

I replaced generic instructions with interim-specific mechanics without checking the section's render scope.

**Fix:** restored the two mainline items to path-agnostic wording, and added the Terraform reset as a separate gated item. One deviation from your suggestion: this template dialect has **no `{{ELSE}}`** — I grepped the whole plugin and there are zero occurrences — so rather than introduce a construct the renderer may not implement, I used the file's existing gated-list-item idiom (as at `generate-docs.md:109`). The gated item also points at "Interim Database Exposure" Step 4 instead of restating the reset procedure, so the two copies can't drift.

Verified after the change that **no** reference to either variable is left outside an interim gate, and `{{IF}}`/`{{ENDIF}}` balance is 44/44 with nothing unclosed.

**On the nit — I closed it rather than leaving it informational.** You're right that accidental bypass is implausible, but "reject one string" is a weak shape for a guard whose entire purpose is "never the open internet," and I'd rather the guard constrain the shape than a spelling. Added a second `validation` requiring every entry to be a well-formed CIDR with a prefix of **/8 or longer**. Exercised against real `terraform plan`:

| input | result |
|---|---|
| `[]` (default) | ACCEPTED |
| `["203.0.113.10/32","203.0.113.11/32"]` (Path B/C) | ACCEPTED |
| `["10.0.128.0/20","10.0.144.0/20"]` (Path A Space CIDRs) | ACCEPTED |
| `["10.0.0.0/8"]` (whole RFC 1918) | ACCEPTED |
| `["0.0.0.0/0"]` | REJECTED |
| `["0.0.0.0/1","128.0.0.0/1"]` (your bypass) | REJECTED |
| `["0.0.0.0/2","64.0.0.0/2","128.0.0.0/2","192.0.0.0/2"]` | REJECTED |
| `["0.0.0.0/7"]` | REJECTED |
| `["not-a-cidr"]` / `["203.0.113.10"]` | REJECTED |
| `["203.0.113.10/32","0.0.0.0/1"]` | REJECTED |

I chose /8 deliberately over a stricter floor: it admits every shape the guide prescribes — including a whole `10.0.0.0/8` — so there is **no false-block risk** for a customer with an unusually large Private Space range, while no permitted combination can cover the internet without hundreds of entries. A tighter floor like /16 would have matched the documented Space defaults exactly but risked blocking a legitimate /12.

While there I also fixed an inconsistency of my own: the variable description said "bounded **/32** allowlist" while Path A prescribes Space CIDRs like /20. Corrected to name both shapes.

`terraform fmt -check` clean on the emitted HCL; `mise run build` exit 0; plugin tests 37 passed.

leon1418 (commented):
[🤖 AI review 🤖]

Re-reviewed all 3 files at `c8a415c` after the author's fix commit addressing the prior blocking finding (ungated lockdown references) and the nit (literal-only validation bypass).

**Previous blocking finding — FIXED.** The Post-Migration Lockdown section now renders generic, path-agnostic items unconditionally ("Confirm the database reports 'Not publicly accessible'", "Ensure no `0.0.0.0/0` inbound rules remain"), and the Terraform-reset item is properly gated inside `{{IF migration_approach == "interim_cutover_data_first"}}` / `{{ENDIF}}` (lines 760–764). No interim-only variable reference escapes the gate. `{{IF}}`/`{{ENDIF}}` balance: generate-docs 44/44, generate-terraform 9/9, clarify-interview 0/0.

**Previous nit — FIXED (and strengthened beyond what was asked).** The validation now has two layers: (1) explicit `!contains(…, "0.0.0.0/0")` for defense-in-depth + clear error message, (2) `alltrue([for cidr: can(cidrnetmask(cidr)) && can(regex("/(8|9|[12][0-9]|3[0-2])$", cidr))])` — requiring well-formed IPv4 CIDR with prefix /8 or longer. This correctly rejects: the literal, the /1+/1 split, all prefixes /0–/7, IPv6 (cidrnetmask rejects it), bare IPs, malformed strings, null, and empty strings. The variable description was also corrected from "/32 allowlist" to name both /32 hosts and Private Space CIDRs.

**Deep inspection of Terraform validation semantics:**
- `cidrnetmask()` — IPv4-only, rejects IPv6/malformed/bare-IPs → `can()` returns false.
- `regex("/(8|9|[12][0-9]|3[0-2])$", cidr)` — anchored end-of-string, matches exactly /8 through /32, rejects /0–/7 and /33+.
- `alltrue([])` — vacuously true for empty default, correctly emitting no rule.
- Combined: to cover all IPv4 you'd need 256× /8 entries, exceeding AWS SG limits (60 default, 200 max) and being obvious in `terraform plan` output.
- The /8 floor is a deliberate design choice: admits legitimate `10.0.0.0/8` without false-blocking, while the threat model is network allowlisting + TLS + DB auth (3 layers deep, not perimeter-only).
- `terraform validate` does NOT evaluate variable validations — author correctly verified at `plan` time. This is documented in the PR description.

**Lockdown section for `full_cutover`:** The generic items ("Confirm … 'Not publicly accessible'", "Ensure no `0.0.0.0/0`") are valid security guidance for any migration path — they verify the final state rather than referencing path-specific variables.

**Other checks:**
- clarify-interview.md: All four changed lines replace "publicly accessible" / "public RDS" with bounded-allowlist language. Consistent with the new design.
- `rds.force_ssl` is correctly gated to interim-only (full_cutover on PG 15+ already defaults to 1; on 14 and earlier the database stays private so the risk model is different).
- CI: all 8 checks pass (build, bandit, checkov, gitleaks, semgrep).
- No remaining `0.0.0.0/0` on any database-port path. The literal appears only in prohibition prose and in the validation error message.
- PR is MERGEABLE, no conflicts.

**No actionable findings.** Both prior issues are fully resolved. The security improvement is substantial — replacing world-open prose guidance with a TLS-first, IaC-visible, validated bounded allowlist.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/phases/generate/generate-terraform.md
```diff
@@ -1377,6 +1442,14 @@ variable "db_password" {
...
+# {{IF migration_approach == "interim_cutover_data_first"}}
+variable "interim_db_public_access" {
+  description = "INTERIM ONLY: expose the database on a public endpoint while the app still runs on Heroku. Leave false unless MIGRATION_GUIDE.md 'Interim Database Exposure' Step 2 Path B or C applies. Reset to false at cutover."
+  type        = bool
+  default     = false
+}
+# {{ENDIF}}
 ```
 
 ### Aurora PostgreSQL (when `aws_service == "Aurora PostgreSQL"`)
```
ayn-builds, line 1455: **Blocking: none of the three new interim controls reach the Aurora branch, and Aurora is the one engine where TLS is off by default.**

`aws_service == "Aurora PostgreSQL"` is a live branch, not a hypothetical: `design-mapping.md:275` routes `multi-az-ha` and `multi-region` Heroku plans to it, and the interim question (Q5b/Q6b) is independent of the engine choice, so `migration_approach == "interim_cutover_data_first"` co-occurs with Aurora for exactly the customers with the largest databases and the strongest reason to migrate data first.

Of the three gates this PR adds, only one lands on Aurora:

| gate | RDS PostgreSQL | Aurora PostgreSQL |
|---|---|---|
| `dynamic "ingress"` on `aws_security_group.database` | yes | yes (shared SG) |
| `publicly_accessible = var.interim_db_public_access` | `aws_db_instance`, line 1381 | absent - `aws_rds_cluster_instance` (line 1485) gets no such attribute |
| `rds.force_ssl = 1` | `aws_db_parameter_group`, line 1413 | absent - and `grep -n "cluster_parameter_group" generate-terraform.md` returns nothing, so there is no parameter group resource on this branch to put it in |

The third row is the one that bites, because the default is engine-specific:

- **RDS for PostgreSQL:** `rds.force_ssl` defaults to `1` on major version 15 and later. `design-mapping.md:299` pins `engine_version: "15"`, so this branch was already TLS-enforcing and the new `parameter` block usefully makes that explicit and version-proof.
- **Aurora PostgreSQL:** `rds.force_ssl` defaults to **`0` (off)** on Aurora PG 16 and older ([Security with Amazon Aurora PostgreSQL](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/AuroraPostgreSQL.Security.html)). Same pinned `15`. So the Aurora interim path ships with TLS **not** enforced, and nothing in the generated Terraform turns it on.

Two consequences, both in the direction this PR exists to close:

1. `generate-docs.md:260` tells the reader "there is no console step and no way to skip it" about `rds.force_ssl`. On Aurora that is not true - the parameter is nowhere in the generated code, so `terraform apply` leaves the database accepting plaintext connections. Step 1's verification gate does still catch it (check (b) succeeds, and the reader is told to stop), so the guide is fail-safe in mechanism. But Step 1 first tells them the code already handled it, which is precisely the framing that turns a gate into a formality.
2. `interim_db_public_access` does not exist on the Aurora branch either, so a reader on Path B or C has no Terraform lever for public access at all. Step 3 says to make every interim change in Terraform and "never by hand in the console", and gives the two reasons why. On Aurora that instruction cannot be followed - the console is the only route left, which is the exact unaudited, silently-reverted change Step 3 warns against.

Two mechanics notes for whoever fixes this, since the RDS blocks do not port over directly:

- On Aurora, `rds.force_ssl` is a **cluster**-level parameter, so it needs a new `aws_rds_cluster_parameter_group` plus `db_cluster_parameter_group_name` on `aws_rds_cluster`. Attaching the existing instance-level `aws_db_parameter_group` to the cluster instances will not work.
- In a cluster parameter group it is **dynamic**, so `apply_method = "pending-reboot"` and Step 1's `aws rds reboot-db-instance --db-instance-identifier` do not carry over. `reboot-db-instance` is also the wrong shape for a cluster.

The rest of the change is solid and I verified it: IF/ENDIF balance is 44/44 in `generate-docs.md` and 9/9 in `generate-terraform.md`, `# {{IF ...}}` inside an HCL fence matches three existing uses on `main`, the parameter group is attached at line 1391, and the `/8` shape guard correctly rejects `0.0.0.0/0`, `/1`, `/7`, and the zero-padded `/00` and `/08` evasions while accepting `/8` through `/32`. The Path A/B/C decomposition and the "there is no fourth path" close are the right call. This is a scope gap on one branch, not a problem with the approach.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/phases/generate/generate-docs.md
```diff
@@ -247,30 +247,125 @@ Compare row counts between source and target to confirm data integrity.
...
-   - Update your Heroku `DATABASE_URL` to include SSL parameters:
+##### Step 1 (prerequisite) — Enforce TLS and prove it is in effect
 
-     ```bash
-     heroku config:set DATABASE_URL="postgres://{{TARGET_DB_USER}}:{{TARGET_DB_PASSWORD}}@{{TARGET_DB_HOST}}:{{TARGET_DB_PORT}}/{{TARGET_DB_NAME}}?sslmode=verify-full&sslrootcert=config/rds-ca-bundle.pem" -a {{app_name}}
-     ```
+Complete every part of Step 1 and pass its verification gate before touching Step 2.
 
-   - Add the RDS CA cert to your application repository (e.g., `config/rds-ca-bundle.pem`) and deploy
+1. **Require TLS on the database.** The generated Terraform already sets `rds.force_ssl = 1` in this database's custom parameter group, so `terraform apply` puts it in place — there is no console step and no way to skip it. Because the parameter is static, reboot the instance once so it takes effect:
```
ayn-builds: **Blocking: this claim is false on the Aurora branch.**

"there is no console step and no way to skip it" holds for `aws_service == "RDS PostgreSQL"`, where the PR adds the `rds.force_ssl` parameter at `generate-terraform.md:1413`. It does not hold for `aws_service == "Aurora PostgreSQL"`: that branch has no parameter group resource at all, so the parameter is never emitted, and Aurora PG 16 and older default `rds.force_ssl` to `0` (off). The generated guide therefore tells the reader TLS enforcement is already in place on a branch where `terraform apply` leaves the database accepting plaintext.

Detail and suggested shape of the fix in my comment on `generate-terraform.md`. Either extend the Terraform to cover the Aurora branch, or gate this sentence on the engine so the reader on Aurora is told to set the parameter rather than told it is already set.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/phases/generate/generate-terraform.md
leon1418: Confirmed and fixed in `439088e`. Every claim in this comment held up, including the two mechanics notes — thank you for pre-empting the wrong port of the RDS blocks, because both would have been easy to get wrong.

**Verified the scope gap rather than taking the table on trust:**

- `grep -c "cluster_parameter_group" generate-terraform.md` → **0**, so there was no resource on the Aurora branch to hold the parameter.
- `publicly_accessible` appeared **exactly once** in the file, at line 1381 — and `1381 < 1455`, the `### Aurora PostgreSQL` heading, so it was inside the RDS section only.
- `design-mapping.md:275` routes `multi-az-ha` / `multi-region` → Aurora, and `:299` pins `engine_version: "15"`, so the co-occurrence you describe is real and not hypothetical.

**And the default, which is the part that makes it bite** — confirmed against [Security with Amazon Aurora PostgreSQL](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/AuroraPostgreSQL.Security.html): Aurora PostgreSQL defaults `rds.force_ssl` to **0 (off)** on version 16 and older, and to 1 only from version 17. With `15` pinned, the Aurora interim path was shipping TLS unenforced — for exactly the customers with the largest databases and the strongest reason to migrate data first.

**Both mechanics notes confirmed and followed:**

1. Added `aws_rds_cluster_parameter_group` attached via `db_cluster_parameter_group_name`, rather than pointing the cluster instances at the existing instance-level `aws_db_parameter_group`.
2. **No `apply_method` and no reboot step.** AWS's Aurora parameters guidance states this parameter "can be dynamically changed in the DB cluster parameter group and doesn't require a restart", so `pending-reboot` would be wrong here, and `reboot-db-instance` is the wrong shape for a cluster regardless. `terraform apply` alone is sufficient on this branch.

`publicly_accessible` went onto `aws_rds_cluster_instance`, since on Aurora public access is instance-level. That restores the Path B/C Terraform lever, so Step 3's "never by hand in the console" instruction is now actually followable on Aurora.

**One deliberate departure, and I want to be explicit about it because it is a judgement call.** I did *not* gate anything on `{{IF aws_service == ...}}`. There are **zero** existing uses of an `aws_service` condition anywhere in the skill — per the Step 0 table the renderer's flags are derived booleans (`has_postgres`, `has_beanstalk`, …) — so introducing one would be inventing a construct the renderer may not implement, which is the same reason `{{ELSE}}` was avoided in the previous round. In `generate-terraform.md` the engine split is already done by section heading, so the new resources sit naturally under the Aurora heading. In `generate-docs.md` I made the prose engine-aware instead, which matches how that file already talks about "RDS/Aurora" throughout.

If the renderer *does* support arbitrary `aws_service` comparisons, gating the docs would be tighter than prose covering both engines, and I would rather switch to that than have the reader skim past a branch that does not apply to them. I did not want to assume a capability I could not see used anywhere.

**Verified after the change:** `{{IF}}`/`{{ENDIF}}` 44/44 in `generate-docs.md` (unchanged — no new gate) and 9/9 → 12/12 in `generate-terraform.md`; zero `interim_*` references outside an interim gate in either file, checked by walking gate depth programmatically rather than by eye; `mise run build` exit 0 with checkov 180/0 and 57 node tests passing.

Not folded in, flagged instead: the Aurora branch still has no `log_connections` / `log_disconnections` parity with the RDS parameter group. Pre-existing and unrelated to the interim controls, so it seemed wrong to expand a security fix into it.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/phases/generate/generate-docs.md
leon1418: Confirmed and fixed in `439088e`. The claim was false on Aurora exactly as you describe, and your framing of *why* it matters is the part I want to acknowledge: the guide was fail-safe in mechanism — Step 4's check (b) does still catch a plaintext-accepting database — but Step 1 first told the reader the code had handled it, and "the code already did this, now go verify it" is how a gate becomes a formality. The verification passing is not much comfort if the text upstream of it has already told them what answer to expect.

I fixed both halves, since either alone leaves the reader in a bad state:

- **The Terraform now covers the Aurora branch** (cluster parameter group, detailed in my reply on `generate-terraform.md`), so the claim becomes *true* rather than merely hedged.
- **The claim is now engine-aware anyway**, because the two engines genuinely differ in where the parameter lives and whether a reboot is needed:
  - RDS for PostgreSQL — instance-level `aws_db_parameter_group`, **static**, keeps the `reboot-db-instance` step.
  - Aurora PostgreSQL — cluster-level `aws_rds_cluster_parameter_group`, **dynamic**, no reboot, and `reboot-db-instance` does not apply to a cluster.

The note beneath it now gives both engines' defaults and says plainly that on the Aurora version this guide pins, TLS is **not** enforced by default — rather than only mentioning the RDS 15+/14 split, which is what made the original read as reassuring.

I also added one line that is not strictly required by your finding: *"Either way, do not treat this as done because the code exists — step 4 below is the gate that proves it."* Your point was that the framing turns a gate into a formality, and fixing only the factual error would have left that framing intact for RDS readers too, where the claim was accurate all along.

Took the prose route rather than `{{IF aws_service == "Aurora PostgreSQL"}}` — reasoning and the alternative I would prefer if the renderer supports it are in my reply on the other thread.

ayn-builds (approved):

PR #200, opened 2026-08-05
fix(heroku-to-aws): add eco/basic rows to the Fargate + EKS dyno sizing tables

leon1418:
## Problem

**An `eco` or `basic` worker scaled past one instance disappears from both the design and the cost estimate.** This is the default compute path — no unusual configuration required.

The three dyno→AWS sizing tables under `knowledge/design/` did not agree on which tiers they cover:

| Table | Rows | `eco` / `basic` |
|---|---|---|
| `dyno-eb-sizing.json` | 21 | **present** (`eco` :14, `basic` :20) |
| `dyno-fargate-sizing.json` | 19 | **absent** |
| `eks-pod-sizing.json` | 19 | **absent** |

The failure chain:

1. The **Horizontal Non-Web Capacity Guard** (`design-mapping.md:113-119`, re-checked at `:172`) routes any persistent non-web formation with `quantity > 1` off Elastic Beanstalk onto **Fargate**, because EB SingleInstance cannot preserve horizontal worker scaling.
2. A `basic` worker at `quantity: 2` therefore hits the Fargate table.
3. The Fargate table has no `basic` row → lookup misses.
4. Every table's `_on_not_found` is reject-and-warn-and-continue (`dyno-fargate-sizing.json:4`, prose at `design-mapping.md:190-196`).
5. **The formation is silently dropped from the design and the cost estimate**, leaving only a warning. The customer's worker vanishes.

**Why CI missed it:** the only `basic` fixture is a `web` process at `quantity: 1`, which stays on the EB branch where the lookup succeeds. There was **no `eco` fixture anywhere**, and **no non-web `quantity > 1` fixture at all** — so the guard itself had zero fixture coverage.

## Solution

**1. `eco` and `basic` rows added to both the Fargate and EKS tables.** All three tables now cover the same 21 tiers:

```
EB       rows=21  eco=True  basic=True
Fargate  rows=21  eco=True  basic=True
EKS      rows=21  eco=True  basic=True
EB == Fargate: True    Fargate == EKS: True
```

Sizing anchors on what the EB table already asserts for these tiers (`heroku_memory_mb: 512`). Fargate uses `256` CPU / `512` MiB — the smallest legal task size. Per the [ECS task CPU/memory matrix](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/fargate-tasks-services.html), 256 CPU units permit 512 MiB / 1 GB / 2 GB only, so this is valid and is the floor. (Heroku Eco, Basic, and Standard-1X are all 512 MB dynos, so `eco`/`basic` matching `standard-1x` is correct, not a copy-paste.)

Both rows carry a `_note` recording that **Heroku Eco is a shared \$5/month 1000-dyno-hour pool whose dynos sleep when idle**, while Fargate tasks and EKS pods are always-on — the size matches but the *cost shape* changes. Sizing an Eco worker silently would have traded one surprise for another.

**2. EKS not-found prose added.** `design-eks.md` documented only the lookup; unlike the EB and Fargate branches, the reject-and-warn rule existed solely in the JSON `_on_not_found`. It also contained a line saying the unrecognized-dyno case should "halt with error", contradicting the JSON's reject-and-continue — corrected to match. No semantics changed; the docs now agree with the tables.

**3. `schema-discover-heroku.md` enum corrected.** It listed only 7 tiers, omitting `eco`, `basic`, and six others all three tables support. Verified it is agent-facing reference rather than an enforced validator (nothing in `tools/` or `tests/` reads it), so the fix was not strictly required — but a stale enum is a live hazard for an agent normalizing live-capture values, and `discover-live.md:52` lowercases Heroku's `Basic` to `basic` specifically to match these tables.

### Verification

- **New fixture `fixtures/heroku-nonweb-scaling/`** exercises the exact uncovered case: a persistent non-web formation on a `basic` dyno at `quantity: 2`, routed by the guard to Fargate. Registered as a GOLDEN asserter (`run-asserters.py`, +1 line — required, since that runner fails the build on any unregistered `check_expected_*.py`).
- **Negative-tested the guard:** deleting `eco`/`basic` from the Fargate table makes the asserter exit 1 (`fargate sizing table has no 'eco' row`, `tiers in the EKS table but not Fargate: ['basic','eco']`). Restored → PASS.
- `mise run build` → exit **0**; `fixtures:assert` reports `PASS — 7 asserters (3 golden, 4 smoke)`.
- All touched JSON parses.

### Known pre-existing bug, deliberately not fixed here

`performance-xl`, `private-xl`, and `shield-xl` specify `fargate_cpu: 8192` with `fargate_memory: 65536`. 8192 CPU units permit **16 GB–60 GB in 4 GB increments**, so 64 GB is rejected at `RunTask` — a real deploy-time failure, independently reproduced by validating all 21 rows against the AWS matrix. It is **not** fixed here because the correct remedy is a sizing/cost decision, not a typo: dropping to 60 GB violates the table's meet-or-exceed rule against `heroku_memory_mb: 63488`, so the likely fix is `fargate_cpu: 16384`, which doubles vCPU and changes the customer's bill. The three rows are baselined in the new asserter (following the repo's existing `.bandit-baseline.json` / `.gitleaks-baseline.json` pattern) so the matrix check is live for every other row and every future row. **Recommend a follow-up issue.**

Also not covered: the Estimate phase for this path. The fixture stops at Design, so whether Eco's flat-pool baseline is priced sensibly against an always-on Fargate projection remains untested.

## Type of Change

- [x] Bug fix

## Team Folder

- [x] `migrate/`

## Checklist


leon1418 (commented):
[🤖 AI review 🤖]

Reviewed the complete change and relevant surrounding code. No actionable findings.

Verified all 21 tiers are aligned across the Fargate, EKS, and schema tables; the Eco Fargate CPU/memory pair is valid; 37 tests and all 7 asserters pass; and a negative mutation correctly fails when the Eco tier is removed. This closes the silent worker-omission path and adds an effective regression guard. Recommended for merge.

ayn-builds (approved):

PR #201, opened 2026-08-05
fix(llm-to-bedrock): correct Opus 4.8 fallback rate to $5/$25 per 1M

leon1418:
## Problem

`STATIC_FALLBACK` in `skills/llm-to-bedrock/scripts/bedrock_pricing.py` priced Claude Opus 4.8 at **\$15 / \$75 per 1M tokens**:

```python
"anthropic.claude-opus-4-8-20250610-v1:0":      {"input_per_1k_usd": 0.015, "output_per_1k_usd": 0.075},
"us.anthropic.claude-opus-4-8-20250610-v1:0":   {"input_per_1k_usd": 0.015, "output_per_1k_usd": 0.075},
```

The correct rate is **\$5 / \$25 per 1M** — `0.005` / `0.025` per 1K. The wrong row was copied from Opus 4.1's legacy rate. Confirmed against two independent sources: this repo's own rate card (`skills/gcp-to-aws/references/shared/pricing-cache.md:379` → `5.00 | 25.00`) and current published Anthropic pricing for Claude Opus 4.8.

**This is not a rare fallback path.** `lookup()` consults the static table **first**, before `import boto3` and the PriceList API — so this was the primary source, not a last resort. It also carried an authoritative-sounding provenance comment (`# Source: https://aws.amazon.com/bedrock/pricing/ (checked 2026-06)`), giving a reader no signal that the number was wrong.

Every Opus 4.8 estimate the skill produced was **3× too high**. For a 1M-input / 200K-output workload: **\$30.00 reported vs \$10.00 actual**. Of the defects in this batch this is the one most likely to flip a model-selection decision — against Opus 4.8, on a fabricated cost basis.

## Solution

**1. Both rows corrected** to `0.005` / `0.025`.

**2. Every other row audited** against `pricing-cache.md` — a table with one bad copy-paste may have more. All 10 rows checked; the two Opus 4.8 rows were the only wrong ones:

| Key | per-1K in/out | cache per-1M | verdict |
|---|---|---|---|
| `[us.]anthropic.claude-haiku-4-5-20251001-v1:0` | 0.001 / 0.005 | 1.00 / 5.00 | correct |
| `[us.]anthropic.claude-sonnet-4-6-20250514-v1:0`, `us.anthropic.claude-sonnet-4-6` | 0.003 / 0.015 | 3.00 / 15.00 | correct |
| `[us.]anthropic.claude-opus-4-8-20250610-v1:0` | ~~0.015 / 0.075~~ | 5.00 / 25.00 | **corrected** |
| `amazon.nova-micro-v1:0` | 0.000035 / 0.00014 | 0.035 / 0.14 | correct |
| `amazon.nova-lite-v1:0` | 0.00006 / 0.00024 | 0.06 / 0.24 | correct |
| `amazon.nova-pro-v1:0` | 0.0008 / 0.0032 | 0.80 / 3.20 | correct |

No models added — that would be scope creep (see follow-ups).

**3. Provenance comment repaired rather than re-dated.** Bumping "checked" over a value that was never checked is how this defect survived. The comment now names the cross-check, records what was verified and when, states the root cause, and points at the guard. The `lookup()` note string was stale in the same way and now reads `verified 2026-08-04 against aws.amazon.com/bedrock/pricing and the vendored pricing cache`.

**4. Drift guard added to the existing `pricing:staleness` tool** (`tools/pricing-staleness.ts`), which re-derives the table from the rate card and reports on every build. The root cause is that `STATIC_FALLBACK` and `pricing-cache.md` hold the same facts with nothing keeping them in sync.

It went there rather than into pytest for a specific reason: **the Python suites in `skills/llm-to-bedrock/scripts/` are not run by any CI workflow or mise task**, so a pytest-only guard would have been unenforced. It follows the tool's existing warn-only / `--strict` contract, so it surfaces in `mise run build` and fails the weekly `pricing-staleness.yml` job without breaking unrelated PRs. Handles region prefixes, `-vN` / `:N` / date-stamp suffixes, and long-context rows that normalize to the same base ID.

### Verification

- **Mutation-checked the test:** with the fix, **10 passed**; reverting only `bedrock_pricing.py` gives **1 failed, 9 passed** (`assert 0.015 == 0.005`). The regression test covers both the bare and `us.` inference-profile keys, since the bug was duplicated across both.
- **Mutation-checked the guard:** against the buggy source it emits `MISMATCH` for both keys and exits 1 under `--strict`.
- `mise run build` → exit **0**, with `pricing drift: OK (10 STATIC_FALLBACK row(s) match the rate card)` in the output.

### Recommended follow-ups (not in this PR)

1. **Models missing from the fallback table:** Claude Sonnet 5, Claude Fable 5, Opus 4.6, Nova 2 Lite/Pro, Nova Premier. Sonnet 5 is the notable one — it is active, cheaper than Sonnet 4.6, and its intro pricing (2.00/10.00) expires 2026-08-31, so a hardcoded row would need a time-dependent value. Worth deciding whether the fallback table should carry intro-priced models at all.
2. **`skills/llm-to-bedrock/scripts/` pytest suites are unenforced** — five test files with no CI hook. Adding a mise task would give the whole directory real coverage.
3. **`_static_fallback()` prefix matching is loose:** `key.startswith(base)` on a version-stripped prefix means a future `claude-opus-4-8-1`-style ID would silently match the 4.8 row. Pre-existing; untouched.
4. `agent-advisor` also has pricing logic in its `scripts/` tree, not covered by this guard and not audited here.

## Type of Change

- [x] Bug fix

## Team Folder

- [x] `migrate/`

## Checklist


leon1418 (commented):
[🤖 AI review 🤖]

Reviewed the complete change and relevant pricing/lookup paths. No actionable findings.

Cross-checked the Opus 4.8 fallback against the rate card at $5/$25 per 1M tokens, verified model-ID normalization and floating-point comparisons across the fallback entries, and confirmed the strict drift guard reports both staleness and pricing drift before exiting. The correction and regression protection are sound. Recommended for merge.

ayn-builds (approved):

PR #206, opened 2026-08-06
feat(advisor): port migration-to-aws skills into aws-startup-advisor

icarthick:
## Problem

The migration skills (`gcp-to-aws`, `heroku-to-aws`, `llm-to-bedrock`, `agent-advisor`, `tf-best-practices`) currently live in a standalone `migration-to-aws` plugin at `migrate/plugins/migration-to-aws/`. A stale, outdated copy of one skill (`migration-to-aws`, the old prose gcp-to-aws) ships inside `aws-startup-advisor`. Leadership has asked for **one canonical plugin** — `aws-startup-advisor` — containing all skills, published to Claude Code, Codex, and Cursor.

## Solution

Port the 5 migration skills + canonical `shared/` + 6 agents + CI tooling into `advisor/plugins/aws-startup-advisor/`, making it the consolidated plugin (v2.0.0). The standalone `migration-to-aws` plugin is intentionally left in place — retirement is a follow-up.

**What moved:**
- 5 skills: `agent-advisor`, `gcp-to-aws`, `heroku-to-aws`, `llm-to-bedrock`, `tf-best-practices` + canonical `skills/shared/`
- 6 agents: `generic-phase-worker-rw`, `llm2bedrock-code-analyzer`, `-code-rewriter`, `-log-ingestor`, `-prompt-evaluator`, `-report-generator`
- CI tooling: `tools/` (frontmatter-validator, sync-vendored-shared, fixtures-check, pricing-staleness, run-asserters), `tests/`, `fixtures/`, `scripts/`, `docs/`, `bandit.yml`, `tsconfig.json`

**What changed (not just copied):**
- 27 plugin-scoped invocation prefixes rewritten: `migration-to-aws:` → `aws-startup-advisor:`
- Prose/diagram references rewritten to name actual sibling skills — with careful distinction between platform-generic handoffs and genuinely gcp-specific references
- `.mcp.json`: added `temporal-docs` + `aws-pricing-calculator` servers
- All 4 manifests bumped to 2.0.0 with expanded descriptions/keywords/defaultPrompts
- CI tools parameterized by plugin path (one source, two validation targets)
- `mise.toml` extended to validate both plugins in parallel
- `.github/workflows/security-scanners.yml` bandit job: three-pass
- `.checkov.yaml` skip-paths mirrored for advisor fixtures
- JSON Schema `$id` URLs → URN form (no GitHub Pages publishing; avoids skills.sh flags per #204)
[...]

PR #210, opened 2026-08-10
feat(migration-to-aws): map OpenAI sources to the same GPT model on Bedrock

leon1418:
## Why

OpenAI's proprietary models — GPT-5.6 Sol/Terra/Luna, GPT-5.5, GPT-5.4 — are GA on Amazon Bedrock (mantle/Responses-only, in-region only). The OpenAI mapping logic still assumed they were unavailable on AWS: it cross-mapped every GPT source to Claude/Nova and argued the migration on a price delta, and the execution pipeline (resolve → preflight → evaluate → rewrite → IAM) failed outright against a mantle-only target.

## What changed — one commit per concern

**1. `feat(gcp-to-aws)` — fact base + same-model-first policy (20 files).** `shared/openai-on-bedrock.md` holds the verified facts: model IDs, the `/openai/v1` path (a bare `/v1` 404s), region matrix, TPM-only quotas, GPT-5.6-only prompt caching, and pricing. **Pricing headline: Bedrock in-region is at parity with OpenAI's *data-residency* tier — exactly 1.10× the standard list price — so a same-model move costs ~10% more, never "the same"; the GPT-5.6 long-context (1M) tier is a separate rate (2.0× in / 1.5× out).** The mapping guide becomes a tiered policy: Tier 0 keeps the model; a region gate handles in-region-only; Tier 1 presents same-vendor upgrade and cross-family side by side; cost priority adds an alternative rather than replacing the recommendation. Cross-family comparisons price Claude Sonnet 5 at standard $3/$15 (the $2/$10 promo through Aug 31 appears only as dated asides).

**2. `fix(llm-to-bedrock)` — execution pipeline for mantle targets (32 files).** Model-ID resolution stops expecting an inference profile; preflight probes the Responses API and fails closed on missing deps (now pinned); `iam_policy.py` emits the `bedrock-mantle:*` statement set (`CallWithBearerToken` on `*`, which AWS does not permit narrowing); the evaluator gains Responses connectivity/vision/per-case paths with image payloads built by a tested helper (`.jpg`→`jpeg`, SDK-required `detail`, message wrapper); control states are schema-representable; Gate (a.5) prefix-matches `migration_path`; `bedrock_pricing.py` carries the five GPT rates with exact-match-only lookup. 247 tests per plugin, including static guards over prompt text verified by injecting each regression.

**3. `fix(agents)` — the five llm2bedrock prompts (8 files).** `same_model_family` widens to "the model is unchanged"; evaluator short-circuits scoring but keeps image verification; rewriter never emits Converse code for an `openai.gpt-5*` target; report generator prices same-model pairs at the ~10% premium instead of "cost-neutral".

**4. `fix(advisor)` — drift gate covers `agents/` (1 file).** Four prompts had diverged with `build` green — the exact failure mode #206's gate exists to stop; it just didn't reach `agents/`. Transitional by design: retires with the twin-tree ritual at deprecation.

## Corrections worth flagging

- Earlier revisions of this branch (and the AWS News Blog it cited) used OpenAI standard-tier figures; every rate was low by exactly 1.10×. The blog quotes Luna at $0.20/$1.20 — the standard figure, not the Bedrock rate ($0.22/$1.32); both sources are recorded with the pricing page preferred.
- GPT-5.4 vs Sonnet flips direction: previously "GPT-5.4 5% cheaper", actually Sonnet ~5% cheaper. Also corrected: GPT-5.2 vs Sonnet 50%→17%, Terra vs Sonnet →16%, Luna vs Haiku →75%.
- Reconciled with #167 (Sonnet 4.6→5; standard price identical so most figures carry) and #164 (model-id lint passes on both plugins), and re-synced the agent prompts after #225.

## Verification

model-id lint ×2, `drift:check` (266 identical, 25 allowlisted, 6 skill trees + `agents/`), `shared:check`, `lint:frontmatter`, `lint:types`, `test`, `fixtures:assert`, `pricing:staleness` all pass; **247 pytest per plugin**. The squash was verified tree-identical to the reviewed history before force-pushing (4 prior AI-review rounds, final round clean, all 10 threads answered; anchors reference the pre-squash history).

**Not verified — needs an account:** no live/billable Mantle call has been made. The Responses wire format (image `detail`, message wrapper, `/openai/v1` path) is verified against the pinned SDK's type definitions and AWS docs, not against the service:
`uv run --project . python preflight_bedrock.py --region us-east-1 --models openai.gpt-5.6-luna`

## Sequencing

Lands on both plugin copies, as `drift:check` requires. Per the deprecation plan, the standalone plugin first gains a deprecation banner and both-sided commits remain the rule until the code is actually removed; the `agents/` gate extension retires with it.


migrate/plugins/migration-to-aws/agents/llm2bedrock-prompt-evaluator.md
```diff
@@ -177,7 +177,7 @@ If `total_golden_cases == 0` (T2-2 abort / paste / vision-no-images / embeddings
 
 # 8. Same-model-family short-circuit
 
-If `same_model_family: true` (Anthropic 1P → Bedrock Claude):
+If `same_model_family: true` — either Anthropic 1P → Bedrock Claude, or OpenAI → the same GPT model on Bedrock (`openai.gpt-5*`):
```
leon1418, line 252: [🤖 AI review 🤖] This OpenAI same-model branch cannot be reached successfully for proprietary GPT targets. Section 6 runs first and unconditionally calls `boto3.converse`; these `openai.gpt-5*` models are Mantle-only, so that probe fails and stops before section 8. The per-prompt verification also points to the Converse-only section 10 script. Route Mantle targets through the Responses API for both connectivity and prompt verification, as the new preflight path does.

migrate/plugins/migration-to-aws/skills/llm-to-bedrock/references/helpers/behavior-delta-detection/references/openai-to-bedrock.md
```diff
@@ -10,6 +10,60 @@ In the `detect_grep` recipes below, `<REPO>` is the repository path supplied in
...
+```bash
+grep -rEn 'chat\.completions\.create' <REPO> --include="*.py" --include="*.js" --include="*.ts" | grep -v node_modules | grep -v __pycache__
+grep -rEn 'choices\[0\]\.(message|delta)' <REPO> --include="*.py" --include="*.js" --include="*.ts" | grep -v node_modules | grep -v __pycache__
+```
+
+`user_visible` classification: `true` when the response shape is surfaced to users or persisted (chat transcript rendering, stored conversation history, streaming to a UI); `false` for internal one-shot calls whose text is consumed programmatically.
+
+## reasoning-items-must-round-trip
+
+- `resolution_kind`: `mechanical`
```
leon1418: [🤖 AI review 🤖] `analysis.json` rejects this value: `resolution_kind` is limited to `ux_choice|impl_path`. The analyzer is required to emit these same-model deltas and then validate its output, so detecting this reasoning delta (or either later `mechanical` delta) makes T2-3 unable to produce a valid result. I reproduced the schema failure locally. Use `impl_path` for no-choice rewrites, or extend the schema and every downstream consumer consistently.

migrate/plugins/migration-to-aws/skills/llm-to-bedrock/scripts/preflight_bedrock.py
```diff
@@ -53,6 +55,80 @@ def is_embedding_model(model_id: str) -> bool:
...
+    """Real minimal probe against the mantle Responses API.
+
+    A missing local OpenAI SDK is a caller-machine gap, not a Bedrock verdict, so
+    it passes with an explicit unverified caveat (same policy as an unrecognized
+    embedding family) rather than failing the whole preflight."""
+    try:
+        from aws_bedrock_token_generator import provide_token
+        from openai import BedrockOpenAI
+    except ImportError as e:
+        return {"ok": True, "reason": "mantle_unprobed",
```
leon1418: [🤖 AI review 🤖] This success path is the standard clean-install path, not an edge case. `scripts/pyproject.toml` declares neither `openai` nor `aws-bedrock-token-generator`, while Phase B2 invokes this script with `uv run --project $SCRIPTS`; the imports therefore fail and return `ok: true`. Phase B4 only surfaces `embedding_unprobed`, so endpoint/model/IAM access is never checked and the migration proceeds as if preflight passed. Declare the runtime dependencies and fail closed, or treat `mantle_unprobed` as an explicit stop in B4.

migrate/plugins/migration-to-aws/skills/llm-to-bedrock/references/helpers/behavior-delta-detection/references/openai-to-bedrock.md
```diff
@@ -10,6 +10,60 @@ In the `detect_grep` recipes below, `<REPO>` is the repository path supplied in
...
+grep -rEn 'responses\.create' <REPO> --include="*.py" --include="*.js" --include="*.ts" | grep -v node_modules | grep -v __pycache__
+grep -rEn 'function_call_output|tool_call' <REPO> --include="*.py" --include="*.js" --include="*.ts" | grep -v node_modules | grep -v __pycache__
+```
+
+## endpoint-path-and-credential
+
+- `resolution_kind`: `mechanical`
+- Base URL must be `https://bedrock-mantle.{region}.api.aws/openai/v1` — the `openai/v1` segment is required and differs from the `v1` path other mantle models use. A hardcoded `/v1` returns 404.
+- The API key must be a Bedrock API key or an auto-refreshing token provider, **not** an existing OpenAI key. A long-lived `OPENAI_API_KEY` read from the environment will fail authentication.
+- IAM must grant `bedrock-mantle:*` actions; `bedrock:InvokeModel` does not authorize these models.
```
leon1418, line 51: [🤖 AI review 🤖] The final report still generates `.saws-migrate/iam-policy.json` through `scripts/iam_policy.py`, which emits only `bedrock:InvokeModel*` against foundation-model/inference-profile ARNs. That policy does not authorize the Mantle bearer-token call documented here: AWS requires `bedrock-mantle:CallWithBearerToken` (with `Resource: "*"`). Applying the generated artifact therefore leaves the migrated app unauthorized. Add a Mantle-specific policy branch and coverage for `openai.gpt-5*` targets.

leon1418 (commented):
[🤖 AI review 🤖] Found four blocking correctness gaps in the same-model OpenAI path: evaluation still uses Converse, emitted behavior deltas violate the analysis schema, the standard preflight silently skips Mantle probing, and the generated IAM policy cannot authorize Mantle calls. I read all 22 changed human-written files plus the relevant schema, callers, tests, and IAM helper; verified the API, quota, model, and IAM claims against AWS documentation; and reproduced the schema rejection and `mantle_unprobed` success path. GitHub checks are green, but these functional paths are not covered by the current validation.

migrate/plugins/migration-to-aws/agents/llm2bedrock-prompt-evaluator.md
leon1418: Confirmed and fixed in `893f88d`.

Reproduced the ordering: `§6` starts at line 125 and `§8` at line 178, and `§6`'s probe calls `boto3.converse` unconditionally — so a mantle-only target aborted the agent before `§8` was ever evaluated. You were also right that `§10` had the same problem independently.

What changed:

- Added **`§6a` Mantle connectivity check** using `BedrockOpenAI` + `client.responses.create`, with its own failure taxonomy: 404 → `model_unavailable_in_region` (these are in-region only, so the `resolve-bedrock-model-id` inference-profile ranking does not apply and there is no cross-region fallback to offer), 401/403 → `model_access` vs `authz` where the authz detail names `bedrock-mantle:*` rather than `bedrock:InvokeModel`, `ModuleNotFoundError` → `mantle_deps_missing`.
- Added an API-path selection table at the top of `§6` keyed on `<TARGET_MODEL_ID>`, so the routing decision is made once and applies to `§6`, `§9.5` and `§10`.
- `§8` now states explicitly that per-prompt verification uses the path `§6` selected — the `§6a` Responses client for mantle targets — and must not call `§10`'s script.
- Marked `§10` Converse-only, and tightened `§8`'s skip list to name `§9.5` and `§10` explicitly. `§9.5` is also Converse-based and its exclusion was previously only implied by "skip to §14", which is exactly the ambiguity that produced this bug.
- Also corrected two stale lines in `§3`: `<TARGET_MODEL_ID>` no longer says "substitute in every `boto3.converse` call", and `same_model_family` no longer reads "only for Anthropic 1P".

migrate/plugins/migration-to-aws/skills/llm-to-bedrock/references/helpers/behavior-delta-detection/references/openai-to-bedrock.md
leon1418: Confirmed and fixed in `893f88d`.

`scripts/schemas/analysis.json:71` restricts `resolution_kind` to `["ux_choice", "impl_path"]`, so all three `mechanical` blocks would have failed the analyzer's own output validation.

Fixed by using `impl_path` — **no schema change**. Checking how `impl_path` is already used made clear the new value was redundant rather than missing: `behavior-delta-detection.md:97` defines it as "do NOT ask the user, pick the default specified in the reference", and `:119` records it as `impl_path_default`. That is precisely the semantics I was reaching for, so inventing `mechanical` added a second name for an existing concept on top of breaking the schema.

Changed at lines 35, 48 and 56; `grep` for `mechanical` in that file now returns 0.

migrate/plugins/migration-to-aws/skills/llm-to-bedrock/scripts/preflight_bedrock.py
leon1418: Confirmed, and your framing is the correct one — fixed in `893f88d` by doing both remedies.

I had flagged this as a deliberate judgment call, reasoning that it mirrored the `embedding_unprobed` precedent ("a caller-machine gap is not a Bedrock verdict"). That analogy does not hold, and your point is what breaks it: `embedding_unprobed` fires only for an unrecognized embedding vendor, which is rare, whereas these imports were **guaranteed** to fail because `scripts/pyproject.toml` declared neither package. Treating a certain path as an edge case turned a fail-fast preflight into an unconditional green light, with endpoint, model and IAM access never checked.

- Declared `openai>=2.45.0,<3` and `aws-bedrock-token-generator>=1,<2` in `scripts/pyproject.toml`, so probing actually runs. Verified `uv` resolves and installs them.
- `probe_mantle_model` now fails closed: `ok: False`, `reason: mantle_deps_missing`, with a detail that states access was NOT verified and points at re-syncing the pinned env.
- Added a `mantle_deps_missing` row to the B4 branch table in `llm-to-bedrock/SKILL.md` as an explicit stop, since B4 previously surfaced only `embedding_unprobed`.
- Two tests: one monkeypatches the imports away and asserts the fail-closed verdict, one asserts the pinned env declares both packages so the fail-closed branch cannot become the normal path again.

migrate/plugins/migration-to-aws/skills/llm-to-bedrock/references/helpers/behavior-delta-detection/references/openai-to-bedrock.md
leon1418: Confirmed and fixed in `893f88d`. I verified the permission model against AWS docs before implementing rather than taking the claim on faith — the `AmazonBedrockMantleInferenceAccess` managed policy document confirms `CallWithBearerToken` is scoped to `Resource: "*"`.

`iam_policy.py` now partitions targets and emits up to three statements:

- `BedrockInvokeModelScoped` — unchanged, for `bedrock-runtime` targets
- `BedrockMantleInference` — `bedrock-mantle:CreateInference` / `Get*` / `List*` on `arn:aws:bedrock-mantle:{region}:{account}:project/*`
- `BedrockMantleCallWithBearerToken` — on `*`, which AWS does not permit narrowing

Two details worth flagging beyond the report:

1. **An all-mantle run omits `BedrockInvokeModelScoped` entirely.** Keeping it would have emitted `"Resource": []`, which is an invalid policy. Mantle models have no foundation-model or inference-profile ARN to scope to, so the absence is correct rather than a generation failure — noted in the report-generator prompt so it is not read as a bug.
2. **The two mantle statements are not redundant.** The AWS reference policies split them: the first covers SigV4, the second covers Bedrock API-key (bearer token) auth. Since the rewriter's generated client authenticates with a bearer token, both are required. Recorded in a code comment.

Mantle authorizes at project granularity, so the policy cannot be scoped to specific model IDs; the report now says so explicitly and points at an SCP for restricting the model set.

Six tests cover this: the mantle action set, the bearer-token wildcard scope, region/account scoping, mixed runtime+mantle runs, `gpt-oss` staying on the runtime path, and the empty-Resource guard.

One self-inflicted issue found while fixing: my first cut omitted the statement whenever `resources` was empty, which broke the pre-existing `test_empty_models` (it asserts the legacy `Statement[0]["Resource"] == []` shape for an empty model list). Narrowed the condition to `resources or not mantle_ids` so only the all-mantle case is omitted and existing callers see no behaviour change.

migrate/plugins/migration-to-aws/skills/llm-to-bedrock/references/helpers/behavior-delta-detection/references/openai-to-bedrock.md
```diff
@@ -10,6 +10,60 @@ In the `detect_grep` recipes below, `<REPO>` is the repository path supplied in
...
+grep -rEn 'function_call_output|tool_call' <REPO> --include="*.py" --include="*.js" --include="*.ts" | grep -v node_modules | grep -v __pycache__
+```
+
+## endpoint-path-and-credential
+
+- `resolution_kind`: `impl_path`
+- Base URL must be `https://bedrock-mantle.{region}.api.aws/openai/v1` — the `openai/v1` segment is required and differs from the `v1` path other mantle models use. A hardcoded `/v1` returns 404.
+- The API key must be a Bedrock API key or an auto-refreshing token provider, **not** an existing OpenAI key. A long-lived `OPENAI_API_KEY` read from the environment will fail authentication.
+- IAM must grant `bedrock-mantle:*` actions; `bedrock:InvokeModel` does not authorize these models.
+- Not user-visible — always `user_visible: false`, `resolution_kind: mechanical`. Apply without prompting.
```
leon1418: [🤖 AI review 🤖] This leaves the original schema-breaking value in the same normative block. The header now says `impl_path`, but this sentence still instructs the analyzer to emit `resolution_kind: mechanical`. `analysis.json` accepts only `ux_choice|impl_path`; a validator mutation check returned `impl_path` valid and `mechanical` invalid. Because T2-3 reads this reference to construct `behavior_deltas`, the contradiction can still make analyzer output fail validation. Change this sentence to `impl_path` (and ideally add a reference/schema consistency test).

migrate/plugins/migration-to-aws/agents/llm2bedrock-prompt-evaluator.md
```diff
@@ -166,6 +179,54 @@ Interpret the result:
...
+```
+
+Interpret the result:
+
+- **Exit 0 + "OK:" line** — proceed to §7.
+- **HTTP 404 / model-not-found** — the model is not available at this endpoint or
+  region. These models are **in-region only**, so there is no cross-region
+  inference profile to fall back to and `resolve-bedrock-model-id`'s
+  inference-profile ranking does not apply. Return
+  `{ blocked: { reason: 'model_unavailable_in_region', detail: '<the exact error> — <TARGET_MODEL_ID> in <REGION>. These models are in-region only; a supported region or a different model is required.' } }`.
```
leon1418: [🤖 AI review 🤖] These new control states do not satisfy the evaluator contract. `scripts/schemas/eval.json` and the C3 blocked-flow handler accept only `model_access`, `source_key_auth`, and `model_unresolvable`. I passed the documented reasons through `validate_result.py --schema eval`: `model_access` returned 0, while `model_unavailable_in_region`, `authz`, and `mantle_deps_missing` each returned 1. A 404, IAM denial, or missing-dependency run therefore cannot produce a valid phase file and will enter the stateless retry path without a representable outcome. Extend the schema and orchestration handling for these reasons, or map them to supported typed states with equivalent recovery semantics.

migrate/plugins/migration-to-aws/agents/llm2bedrock-prompt-evaluator.md
```diff
@@ -177,14 +238,15 @@ If `total_golden_cases == 0` (T2-2 abort / paste / vision-no-images / embeddings
...
 - Output pass / fail per prompt; count successes as `success_count`.
 - Compute `pass_rate = success_count / total_cases` (connectivity-only ratio) and write `failures = total_cases - success_count`.
 - In §14, set `live_source_baseline: false` (no live comparison ran) and add `notes` prefix `same_model_family: true — connectivity-only verification, no rubric scoring`. T2-6 reads that prefix to render the report banner with "connectivity verified" instead of "judge scored X/Y prompts". Set `source_baseline_quality: 'unknown'` (no live baseline ran).
-- Skip to §14 (no §9 baseline, no §11 scoring, no §12–§13 adaptation).
+- Skip to §14 (no §9 baseline, **no §9.5 vision smoke test**, no §10 Converse eval, no §11 scoring, no §12–§13 adaptation). §9.5 and §10 both call `boto3.converse`, so running either against a mantle target would fail for reasons unrelated to the migration.
```
leon1418: [🤖 AI review 🤖] This suppresses the mandatory Layer 2 check for **all** same-model runs. `same_model_family` also includes Anthropic→Claude, where §9.5’s Converse vision call is valid, so this regresses the existing path. For proprietary GPT-5.x, the public AWS model card marks Image input as supported through Responses, but §8’s `input=<prompt>` loop never sends the golden case’s `image_path`. A vision migration can therefore report `pass_rate: 1.0` and become ready-to-merge without making any image request. Keep §9.5 for Converse targets and add a Responses image probe/per-case image input path for Mantle before short-circuiting.

leon1418 (commented):
[🤖 AI review 🤖] Re-reviewed `893f88d` against `04901c8`. The Responses routing, fail-closed preflight dependencies, and Mantle IAM policy are materially improved; 177 script tests pass, and public AWS documentation confirms the API, image-input, and IAM facts. Three blockers remain: one `mechanical` resolution-kind instruction still violates the analysis schema, the new evaluator emits three blocked reasons rejected by `eval.json`, and same-model vision paths now skip every image call. Details inline.

migrate/plugins/migration-to-aws/skills/llm-to-bedrock/references/helpers/behavior-delta-detection/references/openai-to-bedrock.md
leon1418: Confirmed and fixed in `0588a3a`.

You are right that the previous fix left the contradiction inside the same normative block. My edit was a `sed` matching the bullet form `- \`resolution_kind\`: \`mechanical\``, so the prose mention on this line never matched — the header said `impl_path` while the instruction still said `mechanical`.

Changed to `impl_path`, and took the "ideally add a consistency test" suggestion, because a targeted edit missing one form of the same statement is exactly the failure that should not depend on review catching it:

`scripts/test_delta_reference_schema.py` parses **every** `resolution_kind` mention across all files in the delta-reference directory — bullet and prose forms, value inside or outside backticks — and asserts each is in `analysis.json`'s enum. It reads the enum from the schema rather than hardcoding it, so extending the schema does not require touching the test. There is also a guard-the-guard case asserting the regex still matches both forms, so the test cannot silently stop checking.

Verified by reintroducing `mechanical` on this exact line: the test fails with `openai-to-bedrock.md:52 -> 'mechanical'`, then passes again once reverted.

migrate/plugins/migration-to-aws/agents/llm2bedrock-prompt-evaluator.md
leon1418: Confirmed and fixed in `0588a3a`. I reproduced it with `validate_result.py --schema eval`: `model_access` validated, while `model_unavailable_in_region`, `authz`, and `mantle_deps_missing` were each rejected.

I took your first option for two of them and the second for one, deciding per reason rather than uniformly:

- **`model_unavailable_in_region` → mapped to `model_unresolvable`.** No new state. Its documented recovery — user picks or pastes an id, recorded in `resolved_model_overrides` — is precisely the remedy for a mantle 404, so the semantics already match. The `detail` now explains the in-region-only constraint, and the prompt explicitly says not to offer a `us.`-prefixed candidate since those do not exist for these models.
- **`authz` and `mantle_deps_missing` → added to the enum**, with their own recovery rows in the C3 blocked-flow table. Neither has an equivalent existing state, and folding them into `model_access` would have been worse than the bug: it sends someone to the console Model access page for what is actually an IAM policy gap or a broken pinned environment. The `authz` row spells out that the action set differs by path (`bedrock:InvokeModel*` vs the `bedrock-mantle:*` actions), and `mantle_deps_missing` notes that access was never verified so an earlier pass must not be treated as still valid.

`§14`'s own reason list is updated to match, and all four now validate.

Added a second test in `test_delta_reference_schema.py` that extracts every `reason: '...'` the evaluator documents and asserts it is representable in `eval.json` — checking the `blocked` and `partial` enums together, since `throttled` legitimately belongs to `partial`. That closes the class rather than just this instance.

migrate/plugins/migration-to-aws/agents/llm2bedrock-prompt-evaluator.md
leon1418: Confirmed and fixed in `0588a3a`. This was the worst of the three and the diagnosis is exactly right on both halves.

It was also self-inflicted in a specific way worth naming: I added that blanket skip in the previous round *to prevent* a Converse call against a mantle target, and in doing so reintroduced the same silently-green failure mode this branch exists to remove — a vision migration reporting `pass_rate: 1.0` having never sent an image. It also regressed the pre-existing Anthropic → Claude path, where `§9.5`'s Converse vision call was valid.

- **`§9.5` no longer skipped.** It runs whenever `special_patterns.vision == true`, including same-model runs, and selects its API path the same way `§6` does. The section now states explicitly that `§8` short-circuits *scoring*, not image verification.
- **`§9.5a` added** — a Responses image probe for mantle targets, sending the image as a base64 data URL in an `input_image` content block, with the same three outcomes as `§9.5`. On `VISION_FAIL` it must not fall back to text-only, since a text-only pass rate would misrepresent an unverified vision capability. I flagged in the prompt that the content-block shape is the documented OpenAI Responses form and should be reported verbatim if rejected for a non-access reason, rather than guessed at — consistent with how this branch treats Chat Completions support as unverified.
- **`§8`'s per-case loop now requires image input.** Added a per-path table for the text-only and `image_path` cases, and a rule that a case which cannot be sent as an image is recorded as a failure and surfaced in `notes` — never silently downgraded.
- **`§6`'s routing table now covers all three call sites** (connectivity, vision smoke, golden eval) and states that path selection is independent of `same_model_family`: the short-circuit decides whether to score, the table decides how to call.

leon1418 (commented):
[🤖 AI review 🤖] Re-reviewed final head `0588a3a` against prior head `893f88d9` and all three author replies. The `resolution_kind` contradiction is fixed, and `model_unresolvable`/`authz`/`mantle_deps_missing` are now representable end-to-end. The same-model vision fix is incomplete: the smoke probe uses the correct Responses message wrapper, but the per-golden-case table tells the evaluator to pass raw content blocks, which are not a valid `ResponseInputParam`; `.jpg` is also not normalized to `jpeg`/`image/jpeg`. This can make every GPT-5.x vision golden case fail after a successful smoke test.

Validation: targeted regressions 45 passed; full llm-to-bedrock scripts suite 180 passed; `uv lock --check` passed; eval schema accepted the three intended reasons and rejected `model_unavailable_in_region`; in-memory `mechanical` mutation was rejected; pinned OpenAI SDK 2.53.0 type inspection confirms `input` is a list of message/items and content blocks belong under a message's `content`. No live/billable Mantle calls were made.

Merge is not recommended until the inline vision request-shape blocker is fixed and regression-tested.

migrate/plugins/migration-to-aws/agents/llm2bedrock-prompt-evaluator.md
```diff
@@ -177,14 +246,28 @@ If `total_golden_cases == 0` (T2-2 abort / paste / vision-no-images / embeddings
...
+
+Per-prompt verification call, by API path:
+
+| Path               | Text-only case                                                                     | Case with `image_path`                                                                                                             |
+| ------------------ | ---------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------- |
+| Converse (Claude)  | `converse(modelId=…, messages=[{'role':'user','content':[{'text': prompt}]}])`     | add `{'image': {'format': <ext>, 'source': {'bytes': <file bytes>}}}` to the same `content` list, before the text block            |
+| Responses (mantle) | `client.responses.create(model=…, input=prompt, max_output_tokens=…, store=False)` | pass a content list: `[{'type':'input_text','text': prompt}, {'type':'input_image','image_url': 'data:image/<ext>;base64,<b64>'}]` |
```
leon1418: [🤖 AI review 🤖] This per-case image shape is not a valid Responses `input`. In the pinned OpenAI SDK 2.53.0, `ResponseInputParam` is a list of message/items; `input_text` and `input_image` blocks belong inside a message's `content`, as §9.5a correctly shows. Following this row literally sends raw content blocks, so every GPT-5.x vision golden case can fail even after the smoke probe succeeds. Use the full `input=[{"role":"user","content":[...]}]` wrapper here. Also normalize the preserved filename format (`.jpg` → Converse `jpeg` and Responses `image/jpeg`) instead of passing `<ext>` verbatim. Add a static request-shape regression for both the Claude and Mantle per-case paths.

leon1418: Confirmed and fixed in `e677823`. Both halves were real, and the reason neither showed up in my own verification is exactly what you identified: `§9.5` and `§9.5a` hardcode a known-good jpeg, so the smoke tests pass while every golden case would fail.

**The wrapper.** `§8` documented a bare block list as `input` while `§9.5a` correctly wrapped the same blocks in `{"role": "user", "content": [...]}`. I wrote the two in different rounds and never reconciled them — the inconsistency was visible in my own diff and I did not catch it.

**The format.** `format: <ext>` and `data:image/<ext>` produce `jpg` and `image/jpg` for a `.jpg` case. `.jpg` is the common case, which makes this worse than a corner: the smoke image itself is a `.jpg` and only passes because the format is hardcoded.

Rather than restate the correct shapes in prose a third time, both now come from code:

`scripts/image_input.py` — `converse_message` / `responses_message` return complete message items (the Responses one returns the `role`/`content` wrapper, so omitting it is no longer possible); `converse_format` maps extensions onto Bedrock's accepted set with `.jpg -> jpeg`; `mime_type` and `data_url` derive the Responses form from that same mapping so the two cannot drift; an unsupported extension raises rather than passing through to an opaque API rejection. `§8` now calls the helper and the inline dicts are gone.

**Regression tests, per your request:**

- 16 unit tests over the helper: extension normalization (including uppercase), `image/jpg` never appearing in a data URL, png/gif/webp, unsupported types raising, the user-message wrapper, image-before-text ordering for Converse, text-only cases, and the missing-bytes misuse.
- 2 static guards over the prompt itself, since prompt text is not reachable by unit tests: one fails if a Responses image block is documented without a `role`/`content` wrapper, one fails on `image/<ext>` or `format: <ext>` templating. **Both verified by injecting the regression and confirming the specific failure line, then reverting** — the same way I validated the `resolution_kind` guard last round.

Suite is now 198 passing; `uv lock --check` passes.

**One adjacent pre-existing gap I did not change, and want your read on.** `§10`'s Converse golden eval has the same blind spot independently of this branch: it builds `messages.append({"role": "user", "content": [{"text": prompt["user_prompt"]}]})` and never reads `image_path`, so a **cross-family** vision migration (e.g. GPT-4o → Claude) also evaluates vision cases as text-only. `converse_message` is already built and tested for it, so the fix is small — but it changes the primary evaluation path for every cross-family run, so I did not fold it into a review-response commit. Happy to do it here or split it out, whichever you prefer.

migrate/plugins/migration-to-aws/agents/llm2bedrock-prompt-evaluator.md
```diff
@@ -177,14 +246,53 @@ If `total_golden_cases == 0` (T2-2 abort / paste / vision-no-images / embeddings
...
+
+```python
+import sys; sys.path.insert(0, "<scriptsDir>")
+from image_input import converse_message, responses_message
+
+# case["image_path"] is null for text-only cases; pass the file bytes when set.
+raw = open(case["image_path"], "rb").read() if case.get("image_path") else None
+
+# Converse target:
+msgs = [converse_message(case["prompt"], case.get("image_path"), raw)]
```
leon1418: [🤖 AI review 🤖] The canonical golden-dataset record has `user_prompt`, not `prompt` (see the log ingestor schema, dedupe key, source-baseline helper, and §10 below). Following this new example raises `KeyError: prompt` before either Converse or Responses is called, so both same-model Anthropic and GPT-5.x verification paths fail for every case. Use `case["user_prompt"]` in both helper calls and add a regression that executes this documented §8 caller shape against a canonical golden record; the current helper-only tests cannot catch a caller/schema mismatch.

migrate/plugins/migration-to-aws/skills/llm-to-bedrock/scripts/image_input.py
```diff
@@ -0,0 +1,78 @@
...
+    """One Responses API `input[]` item, wrapped as a user message.
+
+    Returning the wrapper (rather than a bare content list) is the point: an
+    unwrapped list is invalid input, and that is the easiest thing to get wrong.
+    """
+    content: list[dict] = [{"type": "input_text", "text": prompt}]
+    if image_path is not None:
+        if raw is None:
+            raise ValueError("raw image bytes are required when image_path is given")
+        content.append({"type": "input_image", "image_url": data_url(image_path, raw)})
```
leon1418: [🤖 AI review 🤖] This image block still does not satisfy the pinned OpenAI 2.53.0 Responses contract: `ResponseInputImageParam` declares both `type` and `detail` as required, while the helper emits only `type` and `image_url`. `BedrockOpenAI` subclasses the same `OpenAI` client, and §9.5a also omits `detail`, so GPT-5.x vision remains outside the pinned request contract even though the wrapper and MIME type are now correct. Add an explicit `detail` value (for example `"auto"`) in the shared helper, use the helper or the same complete shape in the smoke path, and assert all required image fields in the regression test.

leon1418 (commented):
[🤖 AI review 🤖] Re-reviewed current head `e677823` against prior head `0588a3a` and the author reply. The original wrapper and `.jpg` normalization defects are materially improved: request construction is centralized, `.jpg` maps to Converse `jpeg` / Responses `image/jpeg`, and 16 helper tests plus static guards cover those two regressions. Two mandatory execution-contract gaps remain. The new §8 caller reads `case["prompt"]`, but the canonical golden record contains `user_prompt`, so both same-model Anthropic and GPT-5.x paths raise `KeyError` before any API call. The Responses image helper also omits `detail`, which pinned OpenAI 2.53.0 marks required on `ResponseInputImageParam`; §9.5a has the same omission. Details inline.

Validation (no live/billable inference): 76 targeted tests passed; the full scripts suite passed 198/198; pinned OpenAI 2.53.0 types confirm message `role`/`content` wrapping and required image `detail`; pinned botocore 1.43.24 confirms image formats are `png|jpeg|gif|webp`; a canonical golden record reproduces `KeyError: prompt`. All 8 GitHub checks are green. Related overlap: #201 merged into current main and a synthetic merge preserves both its Opus 4.8 rate correction and this PR’s Mantle pricing; #195 has no file overlap; #206 carries a separate, older evaluator port and will need a follow-up sync after this path is corrected.

Merge is not recommended until both inline blockers are fixed and regression-tested.

migrate/plugins/migration-to-aws/agents/llm2bedrock-prompt-evaluator.md
leon1418: Confirmed and fixed in `2445213`. I reproduced the `KeyError: 'prompt'` against the canonical record.

The field name was invented rather than read off the record. What makes it worse is that the correct usage was already in the same file — `§10` line 535 has `prompt["user_prompt"]` — so this was reachable by looking six hundred lines down, not by consulting external docs.

Fixed to `case["user_prompt"]` on both the Converse and Responses branches. Correcting it also surfaced that the loop silently dropped `system_prompt`, which the canonical record carries and `§10` handles: it now passes it as Converse `system`, and on the Responses branch emits it as a `developer`-role item ahead of the user item.

**Regression test** — rather than asserting the one key, the guard parses the canonical `prompts.jsonl` record out of the `llm2bedrock-log-ingestor` prompt and asserts that every golden-case field the evaluator reads (`case[...]`, `prompt[...]`, `entry[...]`, and `.get(...)` forms) exists in it. A field-name drift on either side now fails in CI rather than at eval time. Verified by injecting `case["prompt"]` and confirming the guard names the line.

migrate/plugins/migration-to-aws/skills/llm-to-bedrock/scripts/image_input.py
leon1418: Confirmed and fixed in `2445213`. Verified directly against the pinned SDK: `ResponseInputImageParam.detail` is `Required[Literal["low","high","auto","original"]]` in openai 2.53.0.

`responses_message()` now always emits `detail` (default `auto`, the SDK's documented default; overridable and validated against the literal), and the `§9.5a` smoke snippet sets it inline.

**On the guards** — I made them read the contract from the SDK rather than restate it, so an SDK bump that adds a required field fails here instead of at runtime:

- Required and known keys are derived from the installed `ResponseInputImageParam`, and the emitted block is checked for both missing required fields and fields the SDK does not define.
- The accepted `detail` values are compared against the SDK's own `Literal` args.

One implementation note worth recording, since it would silently defeat a naive version of this test: **`__required_keys__` is unusable for this class.** It is declared `total=False` with per-field `Required[...]` markers, which CPython 3.11 does not fold in — `__required_keys__` is `frozenset()` and `__optional_keys__` contains every key, including `detail`. My first attempt asserted against `__required_keys__` and passed vacuously. The guard now reads requiredness from the annotation origin (`get_type_hints(..., include_extras=True)` + `get_origin(...) is Required`), and asserts `detail` is in the derived required set so the test fails loudly if that derivation ever stops working.

A fourth static guard rejects any inline `input_image` block in prompt text that omits `detail`.

Suite is 209 passing. Still no live/billable Mantle call — the wire format is verified against the SDK's type definitions, not against the service.

leon1418 (commented):
[🤖 AI review 🤖] Re-reviewed current head `2445213` against prior head `e677823` and the author replies. Both prior blockers are fixed: §8 now reads the canonical `user_prompt` field for both Anthropic Converse and GPT-5.x Responses paths (and preserves `system_prompt`), while `responses_message()` and §9.5a now include the OpenAI 2.53.0-required image `detail` field. The previously fixed Responses message wrapper, `.jpg` → Converse `jpeg` / Responses `image/jpeg` normalization, `resolution_kind` consistency, and evaluator control reasons remain intact. I found no new mandatory or non-blocking findings in the four-file delta.

Validation made no live or billable Mantle calls: 32 targeted tests passed; the complete scripts suite passed 209/209; `uv lock --check --offline` and `git diff --check` passed; direct botocore/OpenAI contract checks validated both API payloads and reproduced the old `KeyError` / missing-`detail` failures as negative controls. All 8 GitHub checks pass. Related PRs: #195 is merged with no file overlap; #201 overlaps pricing files but a synthetic merge is clean and preserves both its Opus 4.8 correction and this PR’s Mantle pricing; #206 is a separate older port and still needs a follow-up sync of this evaluator path. The code changes are ready from this review perspective; merge after updating the branch from `main`, removing draft status, and completing normal human review.

herosjourney (changes requested):
All factual claims below were checked against live AWS documentation on **2026-08-21**. Summary up front: the direction is right and most of the mechanics verify cleanly, but the central "mantle-only / in-region-only / never cost-neutral" premise no longer holds for the GPT-5.6 family — the docs appear to have moved between this branch's fact-base verification date (2026-08-10) and today. The fact file's own Refresh Checklist item 5 anticipated exactly this. Requesting a refresh pass before merge.

## What verifies correctly

- **Model catalog and IDs** — GPT-5.6 Sol/Terra/Luna, GPT-5.5, GPT-5.4 GA with exactly the IDs used here ([model card index](https://docs.aws.amazon.com/bedrock/latest/userguide/model-cards-openai.html)).
- **`/openai/v1` path** — confirmed. The GPT-5.5 card states the model is served at `openai/v1/responses`, explicitly "different from the `v1/responses` path used by other models" on the mantle responses endpoint. The "bare `/v1` 404s" guard is well grounded.
- **Quotas** — the [mantle quotas page](https://docs.aws.amazon.com/bedrock/latest/userguide/quotas-mantle.html) matches this PR in substance: two per-model TPM quotas, explicitly **no RPM quota**, cached input exempt from input-TPM. The removal of the old "shared 10,000 RPM" claim is correct.
- **IAM** — `bedrock-mantle:CallWithBearerToken` is a permission-only action, and [`AmazonBedrockMantleInferenceAccess`](https://docs.aws.amazon.com/aws-managed-policy/latest/reference/AmazonBedrockMantleInferenceAccess.html) grants it on `Resource: "*"`, consistent with the "cannot be narrowed" note in `iam_policy.py`.
- **Prompt caching GPT-5.6 only** — confirmed. The 5.6 cards list prompt caching; the 5.5 card lists client-side tool calling in that slot. Cache economics (90% read discount, 1.25× write, 1,024-token minimum prefix, 4 breakpoints, ≥30 min retention, implicit + explicit modes) match the [launch post](https://aws.amazon.com/blogs/machine-learning/get-started-with-openai-gpt-5-6-sol-terra-and-luna-on-amazon-bedrock/).
- **In-region pricing figures** — every in-region rate in `openai-on-bedrock.md` matches the current model cards, including the long-context 2.0×-in / 1.5×-out tier and GovCloud GPT-5.4 ($3.30 / $19.80).
- **The Luna blog discrepancy is real** — the News Blog's $0.20 / $1.20 appears on the [Luna card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-openai-gpt-56-luna.html) as the **Global CRIS** rate, not the in-region rate. Preferring the pricing page over the blog was the right call (though see Major issue 3 for what that figure actually is).
- **GPT-5.5 / GPT-5.4 scope** — for these two models, "mantle-only, in-region-only" still holds: their cards show a single `bedrock-mantle` Programmatic Access row and in-region pricing only.

## Major issues — all specific to the GPT-5.6 family

**1. "There is no `bedrock-runtime` / Converse path for these models" is now false for GPT-5.6.**
The current [Sol](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-openai-gpt-56-sol.html) and [Luna](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-openai-gpt-56-luna.html) cards list **two** Programmatic Access rows: `bedrock-mantle` (in-region) and `bedrock-runtime` via cross-region inference profiles — `us.openai.gpt-5.6-sol`, `global.openai.gpt-5.6-sol`, and even `in.openai.gpt-5.6-luna` for the India Regions. The [Responses API page](https://docs.aws.amazon.com/bedrock/latest/userguide/bedrock-mantle.html) now recommends `bedrock-runtime` for new applications. `openai-on-bedrock.md` § "Access Path — bedrock-mantle Only" (constraint 2) needs to be split by family.

**2. "No cross-region inference / in-region only" and the hard region gate are wrong for GPT-5.6.**
Geo and Global CRIS are supported for 5.6 on `bedrock-runtime`, whose regional footprint on the Sol card spans 30+ regions. The region gate in the mapping policy ("if the target region is not in this table, the same-model path is unavailable — there is no cross-region fallback") would incorrectly block viable same-model migrations. The gate should be endpoint-aware: mantle in-region matrix for 5.5/5.4, CRIS availability for 5.6. This also affects `resolve-bedrock-model-id` — the PR makes resolution stop expecting an inference profile, but 5.6 CRIS profiles **do** exist on `bedrock-runtime` and `bedrock:ListInferenceProfiles` should return them.

**3. The pricing headline "never cost-neutral, always ~10% more" doesn't survive Global CRIS.**
Global CRIS pricing **is** published on the 5.6 model cards (Sol $5.00 / $30.00; Terra $2.00 / $12.00; Luna $0.20 / $1.20) — the very rates this PR records as "the blog's error." Global CRIS lands at what the fact file itself identifies as OpenAI's standard list price, so for a customer with **no residency requirement**, a same-model move to GPT-5.6 can be cost-parity. The 1.10× premium is real for in-region and Geo CRIS only. The line "Not yet published: global cross-region inference pricing. Do not estimate it." is stale. Downstream, the report generator's flat ~10% premium and `bedrock_pricing.py`'s five single rates are wrong whenever Global CRIS applies — the pricing table needs an inference-option dimension for 5.6.

**4. "Chat Completions is unverified" is contradicted by the Sol card.**
The Sol card's APIs-supported list now includes `Responses`, `Chat Completions`, `Invoke`, and `Converse`, with Guardrails noted as Converse-only and prompt caching as Responses-only on `bedrock-runtime`. This weakens the rewriter's categorical rule "never emit Converse code for an `openai.gpt-5*` target" — still a sane default for a Responses-based source app, but it should not be stated as a platform impossibility.

## Minor issues

- **Catalog omissions:** the Daybreak Red / Daybreak Blue GPT-5.6 Cyber variants are missing from the model catalog. They require Trusted Access for Cyber enrollment and are unlikely migration targets, but the "single source of truth" file should list them with that caveat so they aren't misread as "not on Bedrock."
- **"Every other Bedrock target is available in more regions than these"** — no longer true once 5.6's `bedrock-runtime` CRIS footprint is counted.
- **Live verification gap** (already flagged in the PR description, credit for the honesty): no billable Mantle call has been made. Given the wire-format specifics this PR encodes (image `detail`, message wrapper, `/openai/v1`), one live preflight run against `openai.gpt-5.6-luna` before merge would retire the largest remaining risk cheaply.
- **Refresh Checklist worked as designed** — item 5 ("recheck whether Chat Completions and `bedrock-runtime` support have been added or clarified") predicted precisely the drift found here. Consider adding a "re-verify within N days of merge" rule for fast-moving families, since this branch's facts went stale in under two weeks.

## Suggested resolution

Keep the pipeline work, quota corrections, IAM statement set, and the `agents/` drift-gate extension as-is — they verify. Refresh `openai-on-bedrock.md` (and its dependents: mapping policy region gate, `resolve-bedrock-model-id`, `bedrock_pricing.py`, report generator) to model the two access paths per family: 5.5/5.4 mantle-only in-region at 1.10×, 5.6 dual-endpoint with Geo/Global CRIS where Global CRIS is at standard-price parity.

---
*This review was generated with AI assistance (verified against AWS documentation via the AWS Knowledge MCP server).*

leon1418:
@herosjourney Thank you — this is exactly the review this PR needed, and all four major findings are correct. Before changing anything I re-verified each claim against the live Sol / Terra / Luna / 5.5 / 5.4 model cards (2026-08-21): GPT-5.6 now lists two Programmatic Access rows with Geo (`us.` / `in.`) and Global (`global.`) CRIS ids, publishes per-option pricing with **Global CRIS at the standard list price**, and lists Chat Completions / Invoke / Converse as supported — while GPT-5.5 / GPT-5.4 remain single-row mantle, in-region, In-Region-priced only. So the refresh is split by family throughout, as you suggested. Addressed in `8c65c92`:

- **`openai-on-bedrock.md`** — "Access Path — mantle Only" is now "Access Paths — Split by Family": the 5.6 dual-endpoint table (with the runtime base URL `bedrock-runtime.{region}.amazonaws.com/openai/v1` — notably the `/openai/v1` guard survives on both endpoints), the endpoint-aware region section, and per-option pricing tables. The Luna figure is re-recorded for what it actually is: the News Blog's $0.20/$1.20 was the Global CRIS rate all along, not an error — my earlier "the blog is wrong / global pricing unpublished" note is corrected in place. Daybreak Blue/Red are cataloged with the Trusted Access caveat, and the Refresh Checklist gains your re-verify-within-14-days rule, recording that its own item 5 fired in under two weeks.
- **Region gate** — endpoint-aware: 5.6 is rarely blocked (Geo/Global offered by residency posture, Global being the cost-parity option); 5.5/5.4 keep the hard gate and now also offer a within-vendor 5.6 upgrade alongside cross-family when the region is fixed.
- **Cost statement** — conditional everywhere it appears (mapping guide, report generator, pricing script comments): parity on Global CRIS, ~10% in-region/Geo and for all 5.5/5.4 moves. A new `runtime_openai_cris` migration path joins `mantle_openai_responses`.
- **`resolve-bedrock-model-id`** — 5.6 CRIS ids flow through normal `ListInferenceProfiles` resolution; the catalog check is retained for mantle-form ids and for 5.5/5.4.
- **`bedrock_pricing.py`** — CRIS keys added (Geo at 1.10×, Global at parity), with the exact-match guard widened so an unlisted CRIS variant can't partial-match another tier or option. Tests now assert the per-option factor (1.10× vs 1.0×) per id.
- **Rewriter/evaluator** — "never Converse for `openai.gpt-5*`" is family-dependent as you suggested; CRIS-form ids route down the Converse path (and `is_mantle_model` deliberately not matching prefixed ids is now documented, tested behavior rather than a coincidence).

Gates: model-id lint ×2, drift:check, 249 pytest per plugin, all green.

herosjourney (approved):
Re-reviewed after `8c65c92`, with every factual claim re-checked against the live AWS model cards (2026-08-21). **Approving** — this supersedes my earlier changes-requested review. All four major findings are properly addressed, and the fix quality is high: the refresh didn't just patch the flagged sentences, it restructured the claims so they can't silently regress the same way.

## Majors from my previous review — all resolved

1. **Runtime path** — "Access Path — mantle Only" is now a per-family split with a dual-endpoint table for GPT-5.6 (correct id forms, correct base URLs, and the accurate observation that the `/openai/v1` segment survives on *both* endpoints). `resolve-bedrock-model-id` routes 5.6 CRIS ids through normal `ListInferenceProfiles` resolution while keeping the catalog check for mantle forms and 5.5/5.4.
2. **Region gate** — endpoint-aware: 5.6 falls back to the CRIS path with the data-residency implication stated; 5.5/5.4 keep the hard gate. The over-broad "tightest constraint in the whole path" claim is gone.
3. **Pricing** — conditional everywhere it appears: parity on Global CRIS, 1.10× in-region/Geo and for all 5.5/5.4 moves. `bedrock_pricing.py` gains the eight CRIS keys at the correct rates — including the detail that `in.openai.gpt-5.6-sol` is correctly *absent* (the Sol card carries no India geo profile, only Terra/Luna do). The exact-match guard covers prefixed forms, `test_gpt_rates_by_inference_option` asserts the per-option factor per id, and the inverted 404-remedy test is honestly annotated. The Luna "blog discrepancy" note is corrected in place to what it actually was — the Global CRIS rate, not an error.
4. **Chat Completions** — the "unverified" claim is replaced with the per-endpoint API-surface facts, and the rewriter's Converse rule is now family-dependent. `is_mantle_model` deliberately not matching prefixed ids is now documented, tested behavior.

The minors landed too: Daybreak Blue/Red cataloged with the Trusted Access caveat, the stale "more regions than these" comparison removed, and the Refresh Checklist gains the re-verify-within-14-days rule with an honest record that its own item 5 fired in under two weeks.

## One leftover — non-blocking, but please fix in a follow-up

**`llm2bedrock-prompt-evaluator.md` (both plugin copies):** the §6 API-path table gained the CRIS row, but the sentence immediately below it still reads:

> "The proprietary OpenAI GPT models are served ONLY on the `bedrock-mantle` endpoint. Calling `boto3.converse` against one fails, so running the Converse..."

That directly contradicts the new table row two lines above it — a `us.`/`in.`/`global.`-prefixed 5.6 id is exactly a `boto3.converse` target now. This is the same class of stale-prose-next-to-updated-table drift this PR exists to fix, so worth catching before it confuses an agent run.

## Trivia

- The Price List API note now reads "still carries no GPT-5.x rows (checked 2026-08-04)" under a "Last verified 2026-08-21" header — if it was rechecked on the 21st, the date should say so; if not, "still" overstates the freshness.
- The live-call gap remains open (acknowledged in the PR description; unchanged). The preflight probe against a real GPT-5.6 CRIS id is now doubly worth running, since the new `runtime_openai_cris` path is wire-format-verified only against SDK types and docs.

---
*This review was generated with AI assistance (verified against AWS documentation via the AWS Knowledge MCP server).*

az-zhu (approved):

PR #211, opened 2026-08-10
fix: replace github.io URLs and soften MCP install instructions in prompt-library-for-startups

jkzietz:
Replaces all 7 *.github.io URLs in the prompt-library-for-startups skill with github.com equivalents (or upstream README pointers), and softens three agent-directed MCP install instructions to a review-first phrasing. These URLs trigger Gen Digital's Agent Trust Hub URL reputation scanner, contributing to a Critical Risk rating on skills.sh.

### URL replacements (docs links -> github.com)
- `multi-region-assessment.md` (x2): `awslabs.github.io/mcp/servers/well-architected-security-mcp-server` -> `github.com/awslabs/mcp/tree/main/src/well-architected-security-mcp-server`
- `openapi-to-agentcore-gateway-deployment.md`: `awslabs.github.io/mcp/servers/amazon-bedrock-agentcore-mcp-server` -> `github.com/awslabs/mcp/tree/main/src/amazon-bedrock-agentcore-mcp-server`

### Install-instruction rewording (3 prompt files)
- `aws-architecture-assessment-with-mcp-integration.md`, `resiliency-baseline-evaluation.md`, `security-baseline-evaluation.md`: the AWS Knowledge MCP prerequisite line changed from "fetch the install instructions at <github.io URL> and install it for me" to "see the setup instructions at <github.com URL> and configure it". Removes the agent-directed auto-install pattern and points at the canonical repo. (Note: the old `aws-knowledge-mcp-server/` github.io URL was a 404; the other two swapped github.io docs pages were live 200s, so those two swaps are scanner-driven rather than broken-link fixes.)

### Helm chart repo (open-source-llm-inference.md)
- Removed the `helm repo add ... kubernetes-sigs.github.io ...` line and the dependent `helm install` command, replacing them with a pointer to the upstream chart README (`github.com/kubernetes-sigs/secrets-store-csi-driver/blob/main/charts/secrets-store-csi-driver/README.md`). Unlike the docs links above, this is a functional Helm chart-repo endpoint and this chart is only published via that GitHub Pages repo, so there is no non-github.io form of the command; rather than leave a github.io URL or a broken `<alias>/<chart>` install with no `repo add`, the section now defers entirely to the upstream README.

## Scope
This PR is intentionally scoped to the `prompt-library-for-startups` skill. Other *.github.io URLs remain elsewhere in the repo (e.g. `CODE_OF_CONDUCT.md`, `CONTRIBUTING.md` boilerplate, `migration-to-aws` schema `$id` identifiers, and `heroku-to-aws/.../generate-eks.md`). If the scanner reads the whole repo rather than the published skill, those would need a separate change.

## Verification
- All github.com destination URLs verified live (HTTP 200)
- No *.github.io URLs remain in the prompt-library-for-startups skill
- `markdownlint-cli2`: 0 errors
- No broken commands left in the changeset (the previously non-runnable `helm install` was removed)

### DIFF
--- a/advisor/plugins/aws-startup-advisor/skills/prompt-library-for-startups/references/prompt-library/open-source-llm-inference.md
+++ b/advisor/plugins/aws-startup-advisor/skills/prompt-library-for-startups/references/prompt-library/open-source-llm-inference.md
@@ -97,11 +97,7 @@ aws secretsmanager create-secret \
   --region $AWS_REGION
 ```
 
-Then install the Secrets Store CSI driver and AWS provider. Follow the official Helm install instructions in the upstream chart README: <https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/main/charts/secrets-store-csi-driver/README.md>
-
-```bash
-helm install csi-secrets-store secrets-store-csi-driver/secrets-store-csi-driver --namespace kube-system
-```
+Then install the Secrets Store CSI driver and AWS provider by following the official Helm install instructions in the upstream chart README: <https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/main/charts/secrets-store-csi-driver/README.md>
 
 Create a SecretProviderClass that maps the Secrets Manager secret to a Kubernetes secret:
 
--- a/advisor/plugins/aws-startup-advisor/skills/prompt-library-for-startups/references/prompt-library/open-source-llm-inference.md
+++ b/advisor/plugins/aws-startup-advisor/skills/prompt-library-for-startups/references/prompt-library/open-source-llm-inference.md
@@ -97,10 +97,9 @@ aws secretsmanager create-secret \
   --region $AWS_REGION
 ```
 
-Then install the Secrets Store CSI driver and AWS provider:
+Then install the Secrets Store CSI driver and AWS provider. Follow the official Helm install instructions in the upstream chart README: <https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/main/charts/secrets-store-csi-driver/README.md>
 
 ```bash
-helm repo add secrets-store-csi-driver https://kubernetes-sigs.github.io/secrets-store-csi-driver/charts
 helm install csi-secrets-store secrets-store-csi-driver/secrets-store-csi-driver --namespace kube-system
 ```
 
--- a/advisor/plugins/aws-startup-advisor/skills/prompt-library-for-startups/references/prompt-library/aws-architecture-assessment-with-mcp-integration.md
+++ b/advisor/plugins/aws-startup-advisor/skills/prompt-library-for-startups/references/prompt-library/aws-architecture-assessment-with-mcp-integration.md
@@ -12,7 +12,7 @@ Systematic roadmap for scaling infrastructure as you grow. Get phased architectu
 
 ## Scaling from 10K to 10M Users - AWS Architecture Assessment
 
-Prerequisite: This prompt requires the AWS Knowledge MCP server. If it isn't already installed and available to you, then fetch the install instructions at <https://awslabs.github.io/mcp/servers/aws-knowledge-mcp-server/> and install it for me before re-running this prompt.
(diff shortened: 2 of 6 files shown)


advisor/plugins/aws-startup-advisor/skills/prompt-library-for-startups/references/prompt-library/open-source-llm-inference.md
```diff
@@ -97,10 +97,9 @@ aws secretsmanager create-secret \
   --region $AWS_REGION
 ```
 
-Then install the Secrets Store CSI driver and AWS provider:
+Then install the Secrets Store CSI driver and AWS provider. Follow the official Helm install instructions in the upstream chart README: <https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/main/charts/secrets-store-csi-driver/README.md>
```
ayn-builds: **Blocking.** Removing the `helm repo add` line breaks the `helm install` directly below it. `secrets-store-csi-driver/secrets-store-csi-driver` is a `<repo-alias>/<chart>` reference that only resolves if the alias was registered locally, so with the `repo add` gone this fails with `Error: repo secrets-store-csi-driver not found`. The fenced block reads as copy-pasteable, so anyone running it hits that.

This one is also a different kind of URL from the other six. Those are documentation links; `kubernetes-sigs.github.io/secrets-store-csi-driver/charts` is a Helm chart repository endpoint - it's the canonical upstream repo, and it's what the README you now link to tells you to add. So removing it doesn't really reduce risk, it relocates it: the user still runs the identical command, one hop away where CI and reviewers can't see it.

Three ways out, in the order I'd prefer:

1. Restore the `repo add`. If the scanner flags chart-repo endpoints, that seems worth pushing back on - it's the only one of the 7 that isn't a docs page.
1. Keep it removed, but also drop the now-broken `helm install` so the section is purely "follow the upstream README" with no failing command left behind.
1. Use the full chart URL instead of the alias, which avoids both the alias dependency and a separate `repo add`.

Either way the PR description's "No functional commands broken" and "the functional `helm install` command is retained" need a tweak - it's retained but not runnable as written.

ayn-builds (commented):
Checked the two headline claims and both hold: all 4 destination URLs return 200, and `grep -rn "github\.io"` in the skill returns 0. The MCP install rewording is a real improvement - "fetch the install instructions and install it for me" is an agent-directed auto-install, and the new phrasing drops that without losing the prerequisite. That part I'd merge as-is.

One thing to fix first, inline on `open-source-llm-inference.md`.

Two smaller notes, neither blocking:

**On the 404 claim.** "the old github.io URL was also a 404" is true for one of the three. `aws-knowledge-mcp-server/` is a 404, but `well-architected-security-mcp-server` and `amazon-bedrock-agentcore-mcp-server` both return 200. Doesn't change the outcome since the scanner rationale stands on its own and the github.com targets are canonical anyway - just means two of the swaps traded a live docs page for a repo tree, which is scanner-driven rather than a broken-link fix.

**On scope.** The skill is clean, but 8 `github.io` URLs remain elsewhere in the repo: `CODE_OF_CONDUCT.md:3-4` and `CONTRIBUTING.md:22` (`aws.github.io/code-of-conduct` boilerplate), 4 schema `$id` values under `migration-to-aws` (identifiers rather than fetchable links, so probably out of scope), and `heroku-to-aws/references/phases/generate/generate-eks.md:208` (`repository = "https://aws.github.io/eks-charts"`, same category as the helm one below). If the scanner reads the whole repo rather than just the published skill, this PR won't clear the Critical Risk rating on its own. Scoping to one skill is fine, worth just saying so explicitly.

advisor/plugins/aws-startup-advisor/skills/prompt-library-for-startups/references/prompt-library/open-source-llm-inference.md
jkzietz: Good catch. Going with option 2 since it's the only one that fully removes the `github.io` URL.

Option 3 retains the URL because the full command would include the following URL parameter 
`--repo https://kubernetes-sigs.github.io/secrets-store-csi-driver/charts`

itsSauravK (approved):

PR #216, opened 2026-08-13
fix: resolve skills.sh security scan flags on agent-advisor (curl|sh install, inline interpreter scoring)

leon1418:
## Problem

skills.sh audits flag `agent-advisor` as failing ([Snyk](https://www.skills.sh/awslabs/startups/agent-advisor/security/snyk): **CRITICAL**, [Gen Agent Trust Hub](https://www.skills.sh/awslabs/startups/agent-advisor/security/agent-trust-hub): **HIGH**, both Aug 11), and customers see the warnings on `npx skills add`. Both verdicts are driven by the same pattern: SKILL.md instructs installing uv via `curl -LsSf https://astral.sh/uv/install.sh | sh` (Snyk E005, Trust Hub COMMAND_EXECUTION, and the sole basis of its HIGH recommendation line).

## Fixes

- **`curl | sh` → official install guide** — all five instances repo-wide (agent-advisor and llm-to-bedrock in both plugin copies, plus prompt-library's kiro-project-init) now point to https://docs.astral.sh/uv/getting-started/installation/ with `brew`/`pipx` examples. Snyk's own analysis notes official vendor links are not suspicious. Same remediation pattern as #203/#204/#213.
- **Inline interpreter scoring → committed script** (Trust Hub REMOTE_CODE_EXECUTION) — clarify.md Step 5 asked the agent to execute a 20-line Python block via `python -c`, with `$RUN_DIR` shell-interpolated into the source text. The logic now lives in `scripts/score_units.py`, invoked with the answers.json path as argv. Output verified byte-identical to the old inline block on the same input; the two content-lock tests that regex-extracted the inline code now exercise the committed script behaviorally. Note: the audit names clarify.md *and* confirm.md, but `python -c` only ever existed in clarify.md (verified against the Aug-11-era tree as well) — confirm.md never had it.
- **Freshness lookups pinned to public names** (Trust Hub DATA_EXFILTRATION) — outbound freshness queries (MCP, WebFetch, or CLI) may carry only public service/feature/model/region names, never user code, file contents, prompts, or run-directory data.

## Reviewed, deliberately unchanged

- **EXTERNAL_DOWNLOADS** — Mermaid.js from jsdelivr carries SRI integrity hashes (the audit notes this); the remaining domains are official docs sources used by the freshness feature.
- **PROMPT_INJECTION** — the flagged "scope gate" is first-party authored control flow (the skill gating its own scope), not third-party content overriding the agent.
[...]

PR #218, opened 2026-08-13
fix(llm-to-bedrock): resolve skills.sh credential-handling and dynamic-execution flags

leon1418:
## Problem

skills.sh audits flag `llm-to-bedrock` ([Snyk](https://www.skills.sh/awslabs/startups/llm-to-bedrock/security/snyk): **Fail / HIGH**, [Gen Agent Trust Hub](https://www.skills.sh/awslabs/startups/llm-to-bedrock/security/agent-trust-hub): Warn / MEDIUM, [Socket](https://www.skills.sh/awslabs/startups/llm-to-bedrock/security/socket): Warn — all Aug 13). The Fail driver is Snyk's credential-handling finding at risk 1.00: SKILL.md step B3 offered a **"paste my key"** option — the source-provider API key transited the chat transcript and the agent interpolated it into a `printf` command. Per the audit, the design "forces the LLM to receive and handle the secret verbatim." The transcript warning the option carried does not mitigate.

## Fixes

**Commit 1 — the key never enters the conversation (Snyk HIGH):**
- HARD RULE up front in B3: never collect the key in chat, never echo/cat/interpolate its value; if a user pastes one unprompted, don't use it — recommend rotation and continue with a safe path.
- New primary path: detect a profile-exported `$KEY_ENV_VAR` (presence check only) and materialize env→file in a single command — the value never appears in the transcript.
- The write-it-yourself path now hands the user an exact `read -rs` one-liner for their own terminal, so the key is never echoed anywhere.
- The format check (never prints the key) is shared by both paths; stale "key you pasted" wording swept.

**Commit 2 — baseline scripts become committed code (Trust Hub COMMAND_EXECUTION), key never in a URL:**
- The baseline helper instructed the agent to `Write` two embedded Python programs to temp files at runtime and execute them — the same dynamic-execution class flagged on agent-advisor (#216), one file-write removed. Both are now committed: `scripts/resolve_source_model.py` (model-line-sacred resolver; decision core is a pure `resolve()`) and `scripts/source_baseline.py` (three-provider runner; request builders are pure).
- **15 new unit tests** lock the hard rule (date/version suffixes auto-resolve; alphabetic suffixes and cross-line IDs escalate to the user) and the provider request shapes (OpenAI `max_completion_tokens`, Anthropic top-level `system`, Gemini `systemInstruction`).
- **Gemini key moves from a `?key=` query parameter to the `x-goog-api-key` header** in both scripts — query strings end up in logs.
- The helper's Security section now pins the endpoint set (Trust Hub DATA_EXFILTRATION): the key travels only as an auth header to its own provider's official host — `api.openai.com` / `api.anthropic.com` / `generativelanguage.googleapis.com` — never elsewhere, never in a URL. Test-enforced.

## Reviewed, deliberately unchanged

- **Trust Hub EXTERNAL_DOWNLOADS** (clones a user-supplied repo URL) — that is Step 1 of a migration skill; the repo is the user's own.
- **Trust Hub CREDENTIALS_UNSAFE** (key stored locally) — storage is the feature; it is chmod 600, inside a self-gitignored dir, and collection no longer touches the conversation.
- **Socket's footprint warning** — a holistic judgment about the skill class (cloning + uv tooling + delegation). The credential-capture component is addressed above; the rest is the product.

## Verification

- 175 tests pass on both plugin copies (15 new)
- `cross-plugin-drift.ts` OK (255 identical, 25 allowlisted); dprint clean; bandit 0 issues
- `paste`/`pass through this chat` grep: 0 hits on both copies; no embedded ```python blocks remain in the baseline helper

skills.sh results refresh on their next install-triggered audit of main; expected outcome: Snyk leaves Fail, Trust Hub improves or holds at a softer Warn, Socket likely stays Warn (footprint-class judgment).


advisor/plugins/aws-startup-advisor/skills/llm-to-bedrock/scripts/resolve_source_model.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+def main() -> int:
+    if len(sys.argv) != 2:
+        print("usage: PLAN_MODEL_ID=<id> resolve_source_model.py <env-file>", file=sys.stderr)
+        return 2
+    load_env_file(sys.argv[1])
+    plan_id = os.environ["PLAN_MODEL_ID"]
+
+    if "OPENAI_API_KEY" in os.environ:
```
leon1418: [🤖 AI review 🤖] Provider selection is based on the union of the inherited process environment and the env file, rather than on the key actually present in the authoritative env file. If a user has `OPENAI_API_KEY` exported but this migration's file contains `GEMINI_API_KEY`, both this resolver and `source_baseline.py` choose OpenAI first and then send the Gemini model ID to the OpenAI API. I reproduced that exact case: the file provider was Gemini while both scripts selected OpenAI. Please parse/return the provider key from the file (or clear/ignore ambient provider keys) and add a regression test with conflicting ambient credentials; apply the same fix to both scripts and both plugin copies.

advisor/plugins/aws-startup-advisor/skills/llm-to-bedrock/scripts/resolve_source_model.py
```diff
@@ -0,0 +1,135 @@
...
+def list_anthropic() -> list[str]:
+    req = urllib.request.Request(
+        "https://api.anthropic.com/v1/models",
+        headers={
+            "x-api-key": os.environ["ANTHROPIC_API_KEY"],
+            "anthropic-version": "2023-06-01",
+        },
+    )
+    with urllib.request.urlopen(req, timeout=30) as r:  # nosec B310 — fixed https URL
+        return [m["id"] for m in json.loads(r.read())["data"]]
```
leon1418: [🤖 AI review 🤖] The catalog resolver reads only the first response page. Anthropic exposes additional pages via `has_more`/`last_id`, and Gemini via `nextPageToken`; a valid source model outside page 1 is therefore reported as `not_found`, causing the live baseline to be skipped or an unnecessary user escalation. A focused fake-response test with a continuation token made only one request and returned `not_found` for the page-2 model. Please follow pagination for both providers (with tests for a target on a later page); mirror the change in the migration plugin copy.

advisor/plugins/aws-startup-advisor/skills/llm-to-bedrock/scripts/source_baseline.py
```diff
@@ -0,0 +1,183 @@
...
+        print(f"RESUME: {len(done_live)} live baselines kept, {len(prompts)} to fetch")
+
+    for p in prompts:
+        try:
+            url, headers, body = build(model, p.get("system_prompt") or "", p["user_prompt"])
+            out = extract(_send(url, headers, body))
+            results.append({"id": p["id"], "source_response": out, "status": "live"})
+        except urllib.error.HTTPError as e:
+            results.append(
+                {"id": p["id"], "source_response": "", "status": f"http_{e.code}: {e.reason}"}
```
leon1418: [🤖 AI review 🤖] This discards the provider response body and retains only `HTTPError.reason` (typically just `Bad Request`). The evaluator contract explicitly requires the HTTP 400 body's `message`/`param` so it can distinguish and report a request-shape bug; after this script exits that diagnostic is irrecoverable. Please safely read and parse a bounded error body, record the useful provider message without headers/credentials, and add 400-path coverage (plus the mirrored plugin copy).

leon1418 (commented):
[🤖 AI review 🤖] Reviewed head `939d510b95343f633a13cd24607953a64e685163` against `main`. Scope: all 12 changed human-written files in both plugin copies (1,076 additions / 534 deletions), plus the evaluator call contract, script toolchain/configuration, existing review activity, CI, and cross-plugin dependency/drift context.

Validation run:
- `pytest -q test_resolve_source_model.py test_source_baseline.py` — 15 passed.
- Pinned `uv run --group dev pytest -q` in each plugin's llm-to-bedrock scripts directory — 175 passed + 175 passed.
- `node advisor/plugins/aws-startup-advisor/tools/cross-plugin-drift.ts` — OK (255 identical, 25 allowlisted).
- Focused provider-selection and paginated-catalog reproductions confirmed the two correctness failures called out inline.
- GitHub CI: all 8 reported checks are green (build, gitleaks, bandit, semgrep, checkov, plus scanner annotations).

Findings: **3 mandatory, 0 non-blocking**. The overall direction improves secret handling and code health: keys leave chat/query strings, endpoints are pinned, and runtime-generated Python becomes tested committed code. However, ambient credentials can select the wrong provider, paginated catalogs can falsely reject valid models, and HTTP 400 details required by the evaluator contract are discarded.

GitHub reports the branch as **MERGEABLE**, but `mergeStateStatus` is **BLOCKED**. Recommendation: **not merge-ready** until the three mandatory findings are fixed and covered by regression tests. I did not approve or merge.

advisor/plugins/aws-startup-advisor/skills/llm-to-bedrock/scripts/resolve_source_model.py
leon1418: Fixed in edfda9d — `load_env_file` now returns the FILE's pairs and both scripts select the provider from those alone (file values also win in `os.environ`), so an ambient `OPENAI_API_KEY` can no longer route a Gemini model ID to the OpenAI API. Your exact repro is the regression test (`test_provider_selected_from_file_not_ambient_env`, in both scripts' test files, both plugin copies).

advisor/plugins/aws-startup-advisor/skills/llm-to-bedrock/scripts/resolve_source_model.py
leon1418: Fixed in edfda9d — `list_anthropic` follows `has_more`/`last_id` (limit=1000) and `list_gemini` follows `nextPageToken` (pageSize=1000); OpenAI's list endpoint is cursorless and stays single-shot. Your fake-continuation case is locked by tests: a page-2 model now resolves as `prefix` instead of `not_found`, and the second request's cursor parameter is asserted. Mirrored in both copies.

advisor/plugins/aws-startup-advisor/skills/llm-to-bedrock/scripts/source_baseline.py
leon1418: Fixed in edfda9d — `error_detail()` reads a bounded body (2KB cap, 200-char non-JSON fallback), extracts the provider's `message`/`param` credential-free, and appends it after the `http_<code>: <reason>` prefix — the Step-3 classification grep is unchanged and test-pinned. 400-path coverage added, including non-JSON and unreadable bodies.

migrate/plugins/migration-to-aws/skills/llm-to-bedrock/SKILL.md
```diff
@@ -195,38 +195,63 @@ skill's parser expects — a bare key without the `NAME=` prefix will NOT be par
...
+output — the agent only ever handles the file path. If the user pastes a key into the chat
+unprompted, do not use it: tell them it is now part of the transcript, recommend rotating it,
+and continue with one of the paths below.
+
+First check whether the key is already present in the shell environment (each Bash call
+initializes from the user's profile, so a profile-exported key is visible to every call).
+This prints presence only, never the value:
+
+```bash
+[ -n "${!KEY_ENV_VAR:-}" ] && echo ENV_KEY_PRESENT || echo ENV_KEY_ABSENT
```
ayn-builds: (applies to both plugin copies)

`${!VAR}` is bash-only. zsh (macOS default, and what the Bash tool runs) gives `zsh:1: bad substitution`, so `ENV_KEY_PRESENT` never prints and the new primary path is unreachable for most users. Fails closed at least: the error happens during expansion, before :221's redirect opens the file, so an existing key file isn't clobbered.

POSIX version, same transcript safety:

```bash
[ -n "$(printenv "$KEY_ENV_VAR")" ] && echo ENV_KEY_PRESENT || echo ENV_KEY_ABSENT
```

migrate/plugins/migration-to-aws/skills/llm-to-bedrock/SKILL.md
```diff
@@ -195,38 +195,63 @@ skill's parser expects — a bare key without the `NAME=` prefix will NOT be par
...
 
-- **Yes — I'll paste my key** → warn first: "Note: the key will pass through this chat
-  transcript. If you prefer, choose 'I'll write it to a file myself' instead." Then a second
-  AskUserQuestion to collect it (free-text via Other). Then write it in `KEY=VALUE` form:
+- **Use the `$KEY_ENV_VAR` already in my environment** → materialize env var to file in one
+  command — the value never appears in the transcript:
 
   ```bash
-  printf "%s=%s\n" "$KEY_ENV_VAR" "<key>" > "$REPO/.saws-migrate/.source-provider-env" && chmod 600 "$REPO/.saws-migrate/.source-provider-env"
+  printf '%s=%s\n' "$KEY_ENV_VAR" "${!KEY_ENV_VAR}" > "$REPO/.saws-migrate/.source-provider-env" && chmod 600 "$REPO/.saws-migrate/.source-provider-env"
```
ayn-builds: Same `${!KEY_ENV_VAR}` issue, and this is the one that writes the file:

```bash
printf '%s=%s\n' "$KEY_ENV_VAR" "$(printenv "$KEY_ENV_VAR")" > "$REPO/.saws-migrate/.source-provider-env" && chmod 600 "$REPO/.saws-migrate/.source-provider-env"
```

migrate/plugins/migration-to-aws/skills/llm-to-bedrock/scripts/source_baseline.py
```diff
@@ -0,0 +1,183 @@
...
+            url, headers, body = build(model, p.get("system_prompt") or "", p["user_prompt"])
+            out = extract(_send(url, headers, body))
+            results.append({"id": p["id"], "source_response": out, "status": "live"})
+        except urllib.error.HTTPError as e:
+            results.append(
+                {"id": p["id"], "source_response": "", "status": f"http_{e.code}: {e.reason}"}
+            )
+        except Exception as e:  # noqa: BLE001 — every row must land in the JSONL
+            results.append(
+                {"id": p["id"], "source_response": "", "status": f"error: {type(e).__name__}: {e}"}
```
ayn-builds: `{e}` can carry the key. `http.client.putheader` raises `ValueError('Invalid header value %r' % one_value)` where `one_value` is the header:

```
ValueError: Invalid header value b'Bearer sk-live-SECRETKEY123\rX'
```

Needs a `\r`/`\n`/`\0` mid-value, so unlikely, and it doesn't reach the committed report (status stays in the gitignored JSONL, non-`live` rows fall through to static). But :19-20 promises "never to stdout or the output JSONL". Drop `{e}`, keep `type(e).__name__` - the HTTP cases already have their own codes above. Same traceback hits stderr from `resolve_source_model.py`, which has no try/except.

migrate/plugins/migration-to-aws/skills/llm-to-bedrock/scripts/source_baseline.py
```diff
@@ -0,0 +1,183 @@
...
+
+    try:
+        model = os.environ["SOURCE_MODEL_ID"]
+        dataset_path = os.environ["GOLDEN_DATASET_PATH"]
+        output_path = os.environ["OUTPUT_PATH"]
+    except KeyError as e:
+        print(f"FAIL: required env var {e} not set", file=sys.stderr)
+        return 2
+
+    provider = next((k for k in PROVIDERS if k in os.environ), None)
```
ayn-builds: Non-blocking. First match over all of `os.environ`, OpenAI first. Ambient exported keys are indistinguishable from the file's, so a user with both `OPENAI_API_KEY` and `ANTHROPIC_API_KEY` in their profile and Anthropic as source gets the Anthropic model ID sent to `api.openai.com`. Same precedence exists on `main`, but the ENV path makes ambient keys expected now. Also the probe is provider-specific while this is first-match-wins. Selecting from what `load_env_file` set, or from `KEY_ENV_VAR`, aligns them. Key and endpoint stay paired, so this is wrong results, not a leak.

advisor/plugins/aws-startup-advisor/skills/llm-to-bedrock/scripts/source_baseline.py
```diff
@@ -0,0 +1,216 @@
...
+    just "Bad Request" and the body is irrecoverable after this process exits.
+    """
+    try:
+        raw = e.read(ERROR_BODY_CAP).decode("utf-8", "replace")
+    except Exception:
+        return ""
+    try:
+        err = json.loads(raw).get("error", {})
+        if isinstance(err, dict):
+            msg = err.get("message", "")
```
leon1418, line 199: [🤖 AI review 🤖] This extract is not credential-free yet: it copies provider-controlled `error.message` (or the raw-body fallback) verbatim into the status persisted in `source_baselines.jsonl` for every HTTP status. A focused 401 reproduction whose body contained the placeholder key `sk-live-SECRETKEY123` produced `FULL_SECRET_IN_DETAIL=True`; an auth endpoint can echo the submitted credential, so this reintroduces the key into an artifact the skill explicitly promises will never contain it. The evaluator only needs body details for HTTP 400 request-shape diagnostics, while 401/403 classification uses the code. Please suppress auth-error bodies and redact the known provider-key value before persisting any remaining detail (including the non-JSON fallback), with a regression test proving an echoed key cannot appear in `error_detail()` or the final status. Apply the same fix to both plugin copies.

leon1418 (commented):
[🤖 AI review 🤖] Re-reviewed current head `edfda9d5510b42819bb285c2d05f5c9de3a71eac` (one new commit since previously reviewed `939d510b95343f633a13cd24607953a64e685163`) against the PR merge base. I re-read all 12 changed human-written files in both plugin copies, the complete 1,406-addition/534-deletion PR diff, the full new-commit diff, evaluator caller contract, tool/test/security configuration, all review threads/replies/reviews, PR metadata, CI, and mergeability.

**Prior mandatory findings**
1. **FIXED — authoritative provider selection.** Both `resolve_source_model.py` and `source_baseline.py` now select from the parsed `.source-provider-env` pairs rather than ambient `os.environ`; conflicting ambient-OpenAI/file-Gemini regressions are covered in both suites.
2. **FIXED — Anthropic/Gemini pagination.** Anthropic follows `has_more`/`last_id` with `after_id`; Gemini follows `nextPageToken`; later-page-model regression tests assert the second request and successful resolution.
3. **FIXED functionally — HTTP 400 diagnostics.** `error_detail()` reads a bounded body and preserves `message`/`param` behind the existing `http_<code>` prefix; JSON, non-JSON, unreadable-body, and prefix tests cover the path. However, the same implementation introduces the new mandatory credential-handling issue below.

**New findings**
- **Mandatory (1):** provider-controlled HTTP error text is persisted verbatim for every status. A 401/403 body that echoes the submitted API key therefore writes it into `source_baselines.jsonl`, violating this skill's key-hygiene contract. The inline comment contains the focused reproduction and requested regression coverage.
- **Non-blocking:** none.

I did not duplicate existing external threads. Two existing concerns remain open on this head: the Bash-only `${!KEY_ENV_VAR}` commands in the skill's environment-key path, and the generic exception status that can serialize header values containing a key. Those also need resolution before merge.

**Local validation (exact commands/results)**
- `uv run --project advisor/.../llm-to-bedrock/scripts --group dev pytest -q test_resolve_source_model.py test_source_baseline.py` — **23 passed**.
- Equivalent targeted migration-plugin command — **23 passed**.
- `uv run --project advisor/.../llm-to-bedrock/scripts --group dev pytest -q advisor/.../llm-to-bedrock/scripts` — **183 passed**.
- Equivalent full migration-plugin suite — **183 passed**.
- `node advisor/plugins/aws-startup-advisor/tools/cross-plugin-drift.ts` from a clean worktree — **OK: 255 identical, 25 allowlisted across 6 skill trees**.
- `npx --yes dprint@0.51.0 check` from the clean worktree — **passed**.
- Bandit 1.9.3 against each llm-to-bedrock scripts tree with its plugin config — **passed + passed**.
- Focused simulated 401 error-body reproduction — `FULL_SECRET_IN_DETAIL=True`.

**GitHub CI:** all **8/8** reported checks pass: `build`, Security Scanners `gitleaks`, `bandit`, `semgrep`, `checkov`, plus the reported `Bandit`, `Checkov`, and `gitleaks` check runs. GitHub reports `mergeable=MERGEABLE`, `mergeStateStatus=BEHIND`, and `reviewDecision=REVIEW_REQUIRED`.

**Recommendation:** **not merge-ready** until the new mandatory credential leak and the unresolved existing key-handling threads are addressed. I did not approve, request changes, or merge.

leon1418 (commented):
[🤖 AI review 🤖] Re-reviewed head `edfda9d5510b42819bb285c2d05f5c9de3a71eac`, the single new commit since previously reviewed `939d510b95343f633a13cd24607953a64e685163`, against the PR merge base. Scope: complete current PR diff; all 12 changed human-written files in both plugin copies; the full new-commit diff; evaluator caller/failure-classification contract; script toolchains; all reviews, inline threads/replies, and issue comments; CI; PR metadata; and mergeability.

**Prior mandatory findings**
1. **FIXED — ambient credentials no longer select the provider.** `load_env_file()` returns the authoritative file pairs, and both CLIs choose only among keys present in those pairs (`resolve_source_model.py` lines 38–57 and 150–159; `source_baseline.py` lines 38–51 and 157–172). Both suites cover ambient OpenAI plus file Gemini (`test_resolve_source_model.py` lines 82–91; `test_source_baseline.py` lines 75–82).
2. **FIXED — Anthropic/Gemini pagination.** Anthropic follows `has_more`/`last_id` via `after_id`; Gemini follows `nextPageToken` via `pageToken` (`resolve_source_model.py` lines 75–106). Later-page regression tests assert the second request/cursor and successful discovery (`test_resolve_source_model.py` lines 94–128).
3. **FIXED — HTTP 400 bodies are recoverable.** `error_detail()` performs a bounded 2 KiB read, extracts `message`/`param`, and preserves the `http_<code>` status prefix (`source_baseline.py` lines 122–144 and 196–199); JSON, non-JSON, unreadable-body, and prefix tests cover it (`test_source_baseline.py` lines 85–121).

**New findings**
- **Mandatory:** `error_detail()` persists provider-controlled bodies verbatim for every HTTP status, so a 401/403 body that echoes the submitted key writes it into `source_baselines.jsonl` (`source_baseline.py` lines 125–144, 196–199). Focused reproduction: `SECRET_PRESENT_IN_DETAIL=True`. This is already captured by `discussion_r3779728257`; no duplicate inline comment posted.
- **Mandatory (existing reviewer threads):** both Phase B3 environment-key snippets use Bash-only `${!KEY_ENV_VAR}` and fail under zsh (`zsh: bad substitution`, exit 1), making the advertised primary path unavailable on macOS/default zsh (`discussion_r3779624570`, `discussion_r3779624573`).
- **Mandatory (existing reviewer thread):** the generic exception status persists `{e}`; header-validation exceptions can include the credential value and violate the no-key-in-output contract (`source_baseline.py` lines 200–203; `discussion_r3779624578`).
- **Non-blocking:** none.

**Local validation (exact commands/results)**
- `/home/genli/.local/share/mise/installs/uv/0.12.3/uv-x86_64-unknown-linux-musl/uv run --project advisor/plugins/aws-startup-advisor/skills/llm-to-bedrock/scripts --group dev pytest -q advisor/plugins/aws-startup-advisor/skills/llm-to-bedrock/scripts/test_resolve_source_model.py advisor/plugins/aws-startup-advisor/skills/llm-to-bedrock/scripts/test_source_baseline.py` — **23 passed**.
- Equivalent targeted command for `migrate/plugins/migration-to-aws/.../scripts` — **23 passed**.
- Same `uv run ... pytest -q <scripts-dir>` for the full advisor suite — **183 passed**; full migration suite — **183 passed**.
- `/home/genli/.nvm/versions/node/v24.15.0/bin/node advisor/plugins/aws-startup-advisor/tools/cross-plugin-drift.ts` in a pristine checkout — **OK: 255 identical, 25 allowlisted across 6 skill trees**.
- `/home/genli/.nvm/versions/node/v24.15.0/bin/npx --yes dprint@0.51.0 check` — **passed**.
- `/home/genli/.local/share/mise/installs/uv/0.12.3/uv-x86_64-unknown-linux-musl/uv tool run --from 'bandit==1.9.3' bandit -r <each llm-to-bedrock scripts tree> -c <plugin>/bandit.yml -q` — **passed for advisor and migration copies**.
- Focused simulated 401 `HTTPError` with an echoed placeholder key through `error_detail()` — **`SECRET_PRESENT_IN_DETAIL=True`**.
- `zsh -c 'KEY_ENV_VAR=OPENAI_API_KEY; OPENAI_API_KEY=x; [ -n "${!KEY_ENV_VAR:-}" ] && echo PRESENT'` — **fails: `zsh: bad substitution`, exit 1**.

**GitHub CI:** **8/8 reported checks pass** — `build`, Security Scanners `gitleaks`, `bandit`, `semgrep`, `checkov`, plus reported `Bandit`, `Checkov`, and `gitleaks` check runs. GitHub reports `mergeable=MERGEABLE`, `mergeStateStatus=BEHIND`, and `reviewDecision=REVIEW_REQUIRED`.

**Recommendation:** **not merge-ready** until the three current-head credential/portability blockers above are fixed and regression-tested. The three original mandatory findings are fixed. I did not approve, request changes, merge, or modify any existing workspace files.

leon1418 (commented):
[🤖 AI review 🤖] Re-reviewed current head `edfda9d5510b42819bb285c2d05f5c9de3a71eac` (same commit as the prior re-review — no new push since then). Full re-read of all 12 changed human-written files in both plugin copies, the complete 1,406-addition/534-deletion PR diff, evaluator caller contract, test/security configuration, all review threads (7 leon1418 inline, 4 ayn-builds inline), CI checks, and mergeability.

**Prior mandatory findings (from initial review at `939d510b`)**
1. **FIXED — authoritative provider selection.** `load_env_file` returns the file's pairs; `pick_provider(file_pairs)` / `next(k for k in PROVIDERS if k in file_pairs)` selects from the FILE only. Ambient-override regression test (`test_provider_selected_from_file_not_ambient_env`) covers both scripts — verified passing.
2. **FIXED — Anthropic/Gemini pagination.** `list_anthropic` follows `has_more`/`last_id`→`after_id`; `list_gemini` follows `nextPageToken`. Both capped at `limit=1000`/`pageSize=1000`. Regression tests mock `_get_json` and assert two requests with the cursor param — verified passing.
3. **FIXED — HTTP 400 body preserved.** `error_detail()` reads bounded body (2KB cap), extracts `message`/`param` from structured JSON or truncates raw to 200 chars for non-JSON. Four regression tests (`test_error_detail_extracts_message_and_param`, `test_error_detail_handles_non_json_and_is_bounded`, `test_error_detail_never_raises_on_unreadable_body`, `test_status_line_keeps_contract_prefix`) — all verified passing.

**Unresolved mandatory finding (from prior re-review at this same head)**
- **Credential leak in persisted status — NOT FIXED.** `error_detail()` copies provider-controlled `error.message` (or raw-body fallback) verbatim into the `status` field written to `source_baselines.jsonl`. A 401 where OpenAI echoes the submitted key in the body (documented behavior: 'Incorrect API key provided: sk-live-...') persists the secret to the JSONL. Empirically confirmed: simulated 401 with key in body → `FULL_SECRET_IN_DETAIL=True`. The module docstring (lines 19-20) and the security section of `run-source-model-baseline.md` promise the key "never" appears in the output JSONL. No regression test exists for credential-free filtering. The inline comment at line 139 posted in the prior re-review remains unresolved.

**Unresolved external findings (ayn-builds, same head)**
- `${!KEY_ENV_VAR}` bash-only expansion in SKILL.md L209 and L221: fails silently in zsh (macOS default), making the "Use env var" path unreachable. POSIX alternative provided in thread.
- Generic `except Exception as e` (source_baseline.py ~L199): `{e}` can include the key via `ValueError` from `http.client.putheader` on malformed values containing `\r`/`\n`. The ayn-builds thread notes it requires unusual input and the JSONL is gitignored, so rates it non-blocking, but it does violate the stated contract.

**New findings: 0.** No new issues beyond what is already filed.

**Local validation (exact commands/results)**
- `python -m pytest test_resolve_source_model.py test_source_baseline.py -v` in `advisor/.../scripts/` — **23 passed** (0.12s).
- Same command in `migrate/.../scripts/` — **23 passed** (0.12s).
- `python -m pytest -q` in `advisor/.../agent-advisor/scripts/` — **302 passed** (0.83s).
- Cross-plugin drift check (`diff` on all 4 script files): **all IDENTICAL**.
- SKILL.md drift: only expected plugin-name references differ (`aws-startup-advisor` vs `migration-to-aws`).
- Bandit 1.9.3 security scan on both source scripts: **0 issues** (2 properly-tagged `#nosec`).
- Focused 401 key-echo reproduction: **confirmed key in `error_detail()` output**.
- Focused `ValueError` putheader reproduction: **confirmed key in generic exception status string**.

**GitHub CI:** all **8/8** checks pass (build, gitleaks, bandit, semgrep, checkov + 3 annotations).

**Mergeability:** GitHub reports `mergeable=MERGEABLE`, `mergeStateStatus=BEHIND`, `reviewDecision=REVIEW_REQUIRED`.

**Recommendation:** **not merge-ready** until the credential-leak mandatory finding (error_detail persists key from provider 401 body) is fixed with a regression test, and the `${!VAR}` bash-only issue in SKILL.md is resolved (it renders the primary key-ingestion path non-functional on macOS). I did not approve, request changes, or merge.

migrate/plugins/migration-to-aws/skills/llm-to-bedrock/SKILL.md
leon1418: Fixed in 3df7d84 — took your POSIX version verbatim: the presence probe is now `[ -n "$(printenv "$KEY_ENV_VAR")" ]`, which works in bash and zsh alike. Both plugin copies; the prose now notes why `printenv` and not indirection.

migrate/plugins/migration-to-aws/skills/llm-to-bedrock/SKILL.md
leon1418: Fixed in 3df7d84 — took your command verbatim: the materialize step now uses `"$(printenv "$KEY_ENV_VAR")"`. Both plugin copies.

migrate/plugins/migration-to-aws/skills/llm-to-bedrock/scripts/source_baseline.py
leon1418: Fixed in 3df7d84 — every failure status (the generic `str(e)` path included) now passes through `redact()`, which strips each provider key value read from the env file. Your `putheader` ValueError case is the regression test (`test_redact_strips_key_from_exception_status`). The follow-up AI pass found the sibling leak (provider error bodies echoing the key) — same commit suppresses 401/403 bodies entirely and redacts inside `error_detail()` for the rest.

migrate/plugins/migration-to-aws/skills/llm-to-bedrock/scripts/source_baseline.py
leon1418: Taken in 3df7d84 despite the non-blocking tag — `SOURCE_PROVIDER` (this helper's declared input) is now authoritative in both scripts: the stated provider's key is read from the file and its absence is a hard `no_key`/FAIL, never a guess; file-order detection remains only as the fallback when it is unset. Helper invocations pass it explicitly. Locked by `test_source_provider_env_is_authoritative` in both suites.

advisor/plugins/aws-startup-advisor/skills/llm-to-bedrock/scripts/source_baseline.py
leon1418: Fixed in 3df7d84, both asks: (1) 401/403 bodies are suppressed entirely — `error_detail()` returns nothing for auth codes since their classification uses the code alone; (2) every remaining path (JSON message/param and the non-JSON fallback) is redacted against the file's key values inside the function, and the final status passes through `redact()` as well. Your reproduction is inverted into the regression test `test_echoed_key_never_reaches_detail_or_status` (plus `test_auth_error_bodies_are_suppressed_entirely`), in both plugin copies.

leon1418 (commented):
[🤖 AI review 🤖] Third re-review of head `921a6f04f10fe7bf0e65b93cff9aa013b8585505` (2 new commits since previously reviewed `edfda9d5510b42819bb285c2d05f5c9de3a71eac`: commit `3df7d84` — POSIX env probe, credential-free failure statuses, explicit SOURCE_PROVIDER; and merge commit `921a6f0` bringing main up to date).

**Scope:** All 12 changed human-written files across both plugin copies (SKILL.md ×2, helper .md ×2, resolve_source_model.py ×2, source_baseline.py ×2, test_resolve_source_model.py ×2, test_source_baseline.py ×2). Full file reads, full diff, all inline comments and author replies, tests, CI, cross-plugin drift.

---

## Prior Finding Dispositions

### (A) Original three mandatory findings (first review) — all FIXED, remain fixed:

1. **Provider choice from env file, not ambient keys** — `load_env_file()` returns FILE pairs; `pick_provider()` / `pick_provider_key()` selects from those alone. Regression-locked by `test_provider_selected_from_file_not_ambient_env` in both test files.

2. **Anthropic/Gemini pagination** — `list_anthropic()` follows `has_more`/`last_id`; `list_gemini()` follows `nextPageToken`. Both request `limit=1000`/`pageSize=1000`. Regression-locked by `test_anthropic_pagination_follows_has_more` and `test_gemini_pagination_follows_next_page_token`.

3. **HTTP 400 body preservation** — `error_detail()` extracts bounded (2KB read, 200-char fallback), credential-free message from provider error bodies. Status prefix `http_<code>: <reason>` unchanged for downstream classification grep. Regression-locked by `test_error_detail_extracts_message_and_param`, `test_status_line_keeps_contract_prefix`.

### (B) Second-round mandatory finding (credential leak into JSONL) — FIXED:

- **Auth codes (401/403):** bodies suppressed entirely in `error_detail()` — returns empty string. Locked by `test_auth_error_bodies_are_suppressed_entirely`.
- **Non-auth error bodies:** every remaining path is `redact()`-ed against the file's key values. Locked by `test_echoed_key_never_reaches_detail_or_status`.
- **Generic exception path:** the final `except Exception` status also passes through `redact()`. Locked by `test_redact_strips_key_from_exception_status`.
- **`redact()` is defensive:** tolerates empty/None secrets without crashing. Locked by `test_redact_tolerates_empty_secrets`.

### (C) External concerns:

- **`${!VAR}` bash-only expansion (ayn-builds)** — FIXED. Probe uses `printenv` (POSIX-safe in bash and zsh). Materialize command uses `$(printenv "$KEY_ENV_VAR")`. Both plugin copies. The fix commit message correctly identifies the failure mode (zsh 'bad substitution').
- **Generic exception credential leak (ayn-builds)** — FIXED. See (B) above — `redact()` applied to all status paths.
- **SOURCE_PROVIDER is authoritative (non-blocking, taken)** — FIXED. Both scripts now honor `SOURCE_PROVIDER` as authoritative for multi-key files. Locked by `test_source_provider_env_is_authoritative` in both test files.

---

## New Findings

**0 new findings.** No regressions introduced. The code is clean.

---

## Validation

| Check | Result |
|-------|--------|
| Targeted tests (advisor) | 29/29 passed |
| Targeted tests (migrate) | 29/29 passed |
| Pre-existing test failures (test_validate_result.py) | 84 failed — verified identical on `main`, NOT introduced by this PR |
| Pre-existing collection errors (bedrock_pricing/preflight_bedrock/render_report) | Python 3.9 vs 3.10+ type union syntax — NOT introduced by this PR |
| Cross-plugin drift | ✅ OK (255 identical, 25 allowlisted across 6 skill trees) |
| Bandit (security linter) | 0 issues (2 nosec annotations acknowledged) |
| GitHub CI (8 checks) | All SUCCESS (build, gitleaks, bandit, semgrep, checkov ×2 workflow variants) |
| `${!VAR}` grep | Zero instances in executable code paths; only in explanatory prose |
| 'paste my key' option | Fully removed from both SKILL.md copies |
| Gemini key-in-URL | Zero instances; all `x-goog-api-key` header, test-pinned |

---

## Mergeability

- State: OPEN, MERGEABLE
- Review decision: REVIEW_REQUIRED (needs maintainer approval)
- CI: 8/8 ✅
- Conflicts: none

---

## Recommendation

**MERGEABLE.** All prior mandatory findings (rounds 1 and 2) plus all external reviewer concerns (ayn-builds) are addressed and regression-locked. The new commit (3df7d84) is a focused fix with 10 new tests covering exactly the reported issues. No new findings. Code health is improved: embedded scripts are now committed code, credentials can never leak into output, and the POSIX compatibility issue is resolved.

The PR cleanly improves overall code health and is ready for maintainer approval.

leon1418 (commented):
[🤖 AI review 🤖] Corrective re-review of final current head `921a6f04f10fe7bf0e65b93cff9aa013b8585505` against merge base `09f6b7ded39e414e39b73fa619bb82f3bc319388`. I inspected the complete 12-file PR diff (1,672 additions / 538 deletions), every changed human-written file in full across both plugin copies, all first-parent commits after prior reviewed head `939d510b95343f633a13cd24607953a64e685163` (`edfda9d`, `3df7d84`, merge `921a6f0`), the cumulative corrective diff, relevant evaluator/config/docs/tests, and all review threads/replies/reviews. The merge commit adds no llm-to-bedrock delta beyond `3df7d84`.

**Original mandatory findings**
1. **FIXED — ambient keys cannot override `.source-provider-env`.** Both CLIs choose from parsed file pairs; explicit `SOURCE_PROVIDER` is authoritative. Conflicting ambient-OpenAI/file-Gemini and multi-key tests pass in both copies.
2. **FIXED — Anthropic/Gemini pagination.** Anthropic follows `has_more`/`last_id` → `after_id`; Gemini follows `nextPageToken` → `pageToken`. Both later-page regression tests pass.
3. **FIXED — HTTP 400 diagnostics survive.** `error_detail()` reads at most 2 KiB, preserves `message`/`param` behind the `http_<code>` prefix, and tests cover JSON, bounded non-JSON, unreadable bodies, and status classification.

**Intermediate-head blockers**
- **FIXED — auth-body leakage:** 401/403 bodies are suppressed; remaining body details are redacted against file key values. Auth-body and echoed-key regressions pass.
- **NOT fully fixed — generic-exception leakage:** see mandatory finding below.
- **FIXED — Bash-only docs:** both `${!KEY_ENV_VAR}` snippets now use `printenv`; the exact presence probe passes in Bash and zsh.

**Findings by severity**
- **Mandatory (1): generic exception handling can still expose the key.** `source_baseline.redact()` replaces the raw secret, but `http.client` renders a malformed header using escaped `repr`; for a placeholder key containing CR, the current code produced `PLACEHOLDER_PRESENT_AFTER_REDACT=True` and no `***`. `resolve_source_model.py` still has no exception boundary and its equivalent header-validation exception produced `PLACEHOLDER_PRESENT_IN_EXCEPTION=True` on stderr. The existing thread `discussion_r3779624578` already identifies both paths, so I did not duplicate an inline comment. Safest fix: do not persist/print `str(e)` for generic exceptions (retain the exception type), and add end-to-end regressions using the actual malformed key/Request path for both CLIs rather than a preformatted message that omits the control character.
- **Non-blocking:** none.

**Local validation (uv 0.12.3, CPython 3.10.20, Node v24.15.0; no skips)**
- `UV_PROJECT_ENVIRONMENT=/tmp/pr218-final-review.lZ7UUa/venv-advisor /home/genli/.local/share/mise/installs/uv/0.12.3/uv-x86_64-unknown-linux-musl/uv run --project advisor/plugins/aws-startup-advisor/skills/llm-to-bedrock/scripts --group dev pytest -q advisor/plugins/aws-startup-advisor/skills/llm-to-bedrock/scripts/test_resolve_source_model.py advisor/plugins/aws-startup-advisor/skills/llm-to-bedrock/scripts/test_source_baseline.py` — **29 passed**.
- Equivalent targeted migration command — **29 passed**.
- Same `uv run ... pytest -q <scripts-dir>` for the complete advisor suite — **189 passed**; complete migration suite — **189 passed**.
- `/home/genli/.nvm/versions/node/v24.15.0/bin/node advisor/plugins/aws-startup-advisor/tools/cross-plugin-drift.ts` — **OK: 255 identical, 25 allowlisted across 6 skill trees**.
- `PATH=/home/genli/.nvm/versions/node/v24.15.0/bin:/usr/local/bin:/usr/bin:/bin /home/genli/.nvm/versions/node/v24.15.0/bin/npx --yes dprint@0.51.0 check` — **passed**.
- `/home/genli/.local/share/mise/installs/uv/0.12.3/uv-x86_64-unknown-linux-musl/uv tool run --from bandit==1.9.3 bandit -r <each llm-to-bedrock scripts tree> -c <plugin>/bandit.yml -q` — **passed for advisor and migration copies**.
- Exact `printenv` presence probe — **passed in Bash and zsh**; `git diff --check 09f6b7d..921a6f0` — **passed**.
- Focused real exception-path checks — baseline **leak reproduced** (`PLACEHOLDER_PRESENT_AFTER_REDACT=True`); resolver **leak reproduced** (`PLACEHOLDER_PRESENT_IN_EXCEPTION=True`).

**GitHub CI:** **8/8 current-head checks are completed/successful**: `build`, Security Scanners `gitleaks`, `bandit`, `semgrep`, `checkov`, plus `Bandit`, `Checkov`, and `gitleaks` check runs. GitHub reports `mergeable=MERGEABLE`, `mergeStateStatus=BLOCKED`, and `reviewDecision=REVIEW_REQUIRED`.

**Recommendation:** **not merge-ready** until the remaining generic-exception credential exposure is removed and regression-locked in both plugin copies. I did not approve, request changes, merge, or push.

leon1418:
Both adjudicated findings fixed in [046a6f7](https://github.com/awslabs/startups/pull/218/commits/046a6f7b18c939c8baafab0916e23f1e014903ee):

1. **Escaped credential forms bypass literal redaction** — `redact()` now strips every textual form of each secret: the literal, its `unicode_escape` rendering (a raw CR inside a credential renders as backslash-r in exception text), and the control-char-split fragments (≥6 chars) so no partial echo survives. Longest-first replacement. Regression: the escaped-CR reproduction plus a fragment-echo case.

2. **Resolver tracebacks reach stderr unredacted** — the catalog fetch is now wrapped: any failure emits `{"status": "error", "detail": <redacted>}` on stdout (exit 3) instead of an unhandled traceback, and the helper's status table documents the new row (caller sets `live_source_baseline: false`). Regression: a monkeypatched network failure whose exception text embeds the auth header, asserting nothing unredacted reaches stdout or stderr.

192 tests on both plugin copies; drift, dprint, bandit green.

ayn-builds (approved):

leon1418 (commented):
[🤖 AI review 🤖] Fourth re-review of head `44df6c2dd1e1426f712ecd0a9f89823f67a530f5` after the generic-exception credential-redaction fix.

**Prior finding disposition:** all earlier mandatory findings remain fixed. The final two credential-exposure paths are now fixed in both plugin copies:
- `source_baseline.py` redacts literal, `unicode_escape`, and sufficiently long control-character-separated secret fragments, covering the reproduced raw-CR/LF versus repr-escaped exception mismatch.
- `resolve_source_model.py` now catches catalog/network/request exceptions, emits a redacted JSON `error` result on stdout, and avoids an unhandled credential-bearing traceback on stderr.

The update includes focused regressions for escaped control characters, partial fragment echoes, and resolver failure stdout/stderr hygiene. I inspected the one-commit delta and the complete affected implementations/tests in both plugin trees; no new actionable findings.

**Validation:** targeted tests **32/32 + 32/32 passed**; complete affected suites **192/192 + 192/192 passed**; cross-plugin drift **OK (255 identical, 25 allowlisted)**; `git diff --check` passed; GitHub CI **8/8 successful** (build, gitleaks, Bandit/bandit, Semgrep, Checkov/checkov).

GitHub reports `MERGEABLE`, `CLEAN`, and currently `APPROVED`. **Recommendation: mergeable.** I did not approve or merge.

leon1418 (commented):
[🤖 AI review 🤖] Reviewed head `44df6c2dd1e1426f712ecd0a9f89823f67a530f5` (commit `44df6c2` since previously reviewed `921a6f04`) against merge base `main`.

**Scope:** All 10 changed files in commit `44df6c2` across both plugin copies (resolve_source_model.py, source_baseline.py, test_resolve_source_model.py, test_source_baseline.py, run-source-model-baseline.md × 2 each); full file reads; complete PR diff; all 16 prior reviews; author reply (issuecomment-5287658895); CI checks.

---

## Prior Finding Dispositions

### Round 1 (head `939d510b`) — all FIXED, remain fixed:
1. **Provider selection from env file, not ambient keys** — `load_env_file()` returns FILE pairs; `pick_provider()` / `pick_provider_key()` selects from those alone. Regression-locked.
2. **Anthropic/Gemini pagination** — follows cursors. Regression-locked.
3. **HTTP 400 body preservation** — `error_detail()` extracts bounded credential-free message. Regression-locked.

### Round 2 (head `edfda9d5`) — all FIXED, remain fixed:
- **Auth-body leakage (401/403):** suppressed entirely in `error_detail()`.
- **Non-auth error bodies:** redacted against file key values.

### Round 3 (head `921a6f04`) — addressed but one mandatory remained:
- **POSIX `${!VAR}` bash-only:** FIXED (now uses `printenv`).
- **`SOURCE_PROVIDER` authoritative:** FIXED.
- **Generic exception credential leak:** remained unfixed.

### Round 4 (head `44df6c2`) — this review:
**Both prior mandatory findings are now FIXED:**

1. **Escaped CR/LF credential forms bypass literal redaction** — `_secret_variants()` now produces every textual form: the literal, its `unicode_escape` rendering (a raw CR inside a credential renders as `\\r` in exception text), and control-char-split fragments (≥6 chars). `redact()` applies longest-first replacement. Regression-locked by `test_redact_catches_escaped_control_chars_in_exception_text` and `test_redact_catches_fragments_around_control_chars` in both test files.

2. **Resolver exceptions reach stderr unredacted** — `resolve_source_model.py` now wraps the catalog fetch: any exception emits `{"status": "error", "detail": <redacted>}` on stdout (exit 3) instead of an unhandled traceback to stderr. Helper doc updated with new status row. Regression-locked by `test_network_exception_never_reaches_stderr_unredacted` in both test files.

---

## New Findings

**0 new findings.** No regressions introduced. The code is clean.

---

## Validation

| Check | Result |
|-------|--------|
| Targeted tests (advisor) | 32/32 passed |
| Targeted tests (migrate) | 32/32 passed |
| Full suite (advisor) | 192/192 passed |
| Full suite (migrate) | 192/192 passed |
| Cross-plugin drift (manual diff) | ✅ All 4 script files IDENTICAL |
| Bandit | 0 issues in changed scripts (2 nosec acknowledged) |
| Focused escaped-CR reproduction | ✅ REDACTED (no credential in output) |
| Focused resolver exception | ✅ exit 3, `{"status": "error", "detail": "***"}` |
| GitHub CI (8 checks) | All SUCCESS |

---

## Mergeability

- State: OPEN, MERGEABLE
- Review decision: REVIEW_REQUIRED (needs maintainer approval; ayn-builds APPROVED)
- CI: 8/8 ✅
- Conflicts: none

---

## Recommendation

**MERGEABLE.** All mandatory findings from rounds 1–4 are addressed and regression-locked:
- Keys never enter chat or URL query parameters
- Provider selection uses file pairs, not ambient env
- Pagination follows cursors
- Error bodies are bounded and credential-free
- Auth-error bodies are suppressed
- Escaped credential forms are redacted
- Resolver exceptions emit redacted JSON, not tracebacks

Code health is improved. The PR is ready for maintainer merge.

I did not approve, request changes, merge, or modify any workspace files.

leon1418 (commented):
[🤖 AI review 🤖] Fourth re-review of head `44df6c2dd1e1426f712ecd0a9f89823f67a530f5` (1 new commit since previously reviewed `921a6f04f10fe7bf0e65b93cff9aa013b8585505`: commit `44df6c2` — redact escaped credential forms; resolver failures never traceback to stderr).

**Scope:** All 10 changed files across both plugin copies (resolve_source_model.py ×2, source_baseline.py ×2, test_resolve_source_model.py ×2, test_source_baseline.py ×2, run-source-model-baseline.md ×2). Full file reads, full diff, empirical reproduction of prior blocker, test execution, drift/format/security.

---

## Prior Blocker Disposition

**FIXED — repr-escaped credentials in generic exceptions + resolver stderr boundary.**

The last mandatory finding (third review) identified two paths:

1. **`source_baseline.py` generic exception path** — `redact()` replaced only the raw secret literal; a credential containing CR/LF would appear repr-escaped in exception text (e.g. `sk-live-SECRET\\rKEY123`) and the literal replacement missed it.

   **Fix:** New `_secret_variants()` helper generates every textual form of each secret: the literal, its `unicode_escape` rendering, and all control-char-split fragments ≥6 chars. `redact()` now replaces all variants, longest-first. Empirically verified: a secret with `\\r` no longer appears in any form after redaction.

2. **`resolve_source_model.py` had no exception boundary** — a network/provider exception (whose text can embed the auth header) propagated as an unhandled traceback to stderr unredacted.

   **Fix:** The catalog fetch is now wrapped in `try/except Exception`. Failures emit `{"status": "error", "detail": <redacted>}` on stdout with exit code 3. Nothing reaches stderr. The helper doc now documents the new `error` status row. Empirically verified: monkeypatched ValueError with the key in its text → exit 3, key not in stdout or stderr, JSON output with `***` in detail.

---

## All Prior Finding Dispositions (cumulative)

| # | Finding | Status |
|---|---------|--------|
| 1 | Provider selection from ambient env not file | ✅ FIXED (round 2) |
| 2 | Anthropic/Gemini pagination | ✅ FIXED (round 2) |
| 3 | HTTP 400 body discarded | ✅ FIXED (round 2) |
| 4 | Auth-body credential leak in error_detail | ✅ FIXED (round 3) |
| 5 | Bash-only `${!VAR}` expansion | ✅ FIXED (round 3) |
| 6 | SOURCE_PROVIDER is authoritative | ✅ FIXED (round 3) |
| 7 | repr-escaped cred leak + resolver stderr | ✅ FIXED (this round) |

---

## New Findings

**0 new findings.** The commit is focused, correct, and well-tested.

---

## Validation

| Check | Result |
|-------|--------|
| Targeted tests (advisor) | 32/32 passed |
| Targeted tests (migrate) | 32/32 passed |
| Full suite (advisor) | 192/192 passed |
| Full suite (migrate) | 192/192 passed |
| Cross-plugin drift | ✅ OK (255 identical, 25 allowlisted across 6 skill trees) |
| dprint format check | ✅ clean |
| Bandit (security linter) | 0 issues, both copies |
| git diff --check | ✅ clean |
| GitHub CI (8 checks) | All SUCCESS |
| Empirical CR-secret redaction | ✅ no leak (tested with `\r`-containing secrets) |
| Empirical resolver stderr boundary | ✅ no leak (monkeypatched exception, exit 3, clean stdout) |

---

## Mergeability

- State: OPEN, MERGEABLE, CLEAN
- Review decision: APPROVED (ayn-builds)
- CI: 8/8 ✅
- Conflicts: none

---

## Recommendation

**MERGEABLE.** All mandatory findings across 4 review rounds are addressed and regression-locked. The final commit is focused (2 adjudicated issues), well-documented (commit message traces exactly to the review thread), and adds 3 targeted regression tests that exercise the exact reproduction scenarios. The `_secret_variants()` approach is sound — it handles the literal, unicode-escape rendering, and control-char-split fragments, with longest-first replacement preventing partial-match artifacts. The code cleanly improves overall code health and is ready for merge.

PR #186, opened 2026-07-24
feat(gcp-to-aws): decision report — shared exec core rendered at gate A (stacked on #185)

herosjourney:
## Summary

**Stacked on #185** (decision gate) — merge that first; this branch contains its commits. For this PR's isolated diff: compare `feat/decision-gate...feat/decision-report-core`.

This PR now carries **four stacked concerns** that all rewrite the same executive-section rules in the new shared renderer (`report-decision-core.md`), which is why they land together rather than as sequential conflicting PRs: (1) the decision report itself, (2) timeline/effort honesty, (3) scenario deltas, (4) track-scoped outcomes. Each has its own plain-English problem statement below.

---

### 1. Decision report — shared exec core rendered at gate A

**The problem, in plain English:** #185 gives the founder a decision moment, but the decision pack lives only in chat — close the window and the verdict, costs, and assumptions are gone. The durable report today is built inside Generate, so the only way to get a shareable artifact is to accept 20 Terraform files you didn't ask for. This gives gate choice A ("done for now") a real deliverable: `decision-report.html` — the thing a founder emails their co-founder — plus a Slack-friendly `DECISION.md`, without a second report engine to drift.

- **`references/shared/report-decision-core.md` (new, extracted not written):** the report's entire executive summary — verdict typography (#173), cost comparison with the baseline-quality badge and not-comparable rules (#175), combined run rate, migration shape, risks, assumptions panel, Activate wording — moved out of `generate-artifacts-report.md` Step 1 into a shared renderer spec with two modes. Full-mode rules relocated unchanged; **decision-mode overrides are new normative text** written inline next to each affected rule. **Decision mode** (gate A): exec sections only, pre-Generate fallbacks, a required `decision-cta` section, and the `DECISION.md` twin. **Full mode** (Generate Step 1): loads the same file — Step 1 is now a pointer with a do-not-restate rule, the canonical-vendored discipline this repo adopted after the drift bug. `generate-artifacts-report.md` shrinks 769 → 568 lines.
- **`estimate.md` gate A** renders the pack, validates it, then sets the decide-complete state — the closing message names both files.
- **Replace, never patch:** when Execute runs after a Decide run, `migration-report.html` renders fresh from the JSON artifacts; splicing or extending `decision-report.html` is explicitly forbidden, and the artifacts catalog notes the full report supersedes it.
[...]

PR #223, opened 2026-08-17
Improve migration report decision UX and stdlib quality gates

herosjourney:
## Problem

The GCP-to-AWS migration report contains the information needed to make a migration decision, but its presentation can make that decision harder to understand and share:

- Navigation can appear before the recommendation, delaying the answer to the reader's main question.
- “Stay if” can make a limitation in one migration track (such as analytics or image generation) sound like a reason to abandon the entire migration.
- Important figures are presented with similar visual weight, so readers must determine which numbers matter most.
- The report lacks a concise paragraph that a founder or finance stakeholder can forward without extracting and rewriting the findings.
- Some readability and accessibility requirements are documented but not enforced, allowing broken anchors, insufficient HTML semantics, or report-layout regressions to ship.

Migration skill users should not need to install separate prose linters, browser automation, or accessibility tools to receive a usable report.

## Solution

Make the generated GCP-to-AWS report a clearer decision document and enforce that quality with the plugin's existing, dependency-free Python validator.

The generated report now:

- Shows the migration recommendation before the table of contents.
- Uses a prominent text verdict with plain-language execution details instead of color-dependent verdict pills.
- Uses “Stay entirely if” for whole-migration blockers and keeps track-specific limitations scoped to their affected track.
- Visually prioritizes the estimated AWS monthly run rate and migration shape.
- Adds a static, copy-ready “Share With Your CFO” summary when cost estimates are available.
- Moves technical terms into a dedicated glossary appendix while preserving the detailed architecture, AI, security, configuration, and cost appendices.
- Keeps assumption text readable at WCAG AA contrast levels.

The shipped `validate-migration-report.py` now rejects reports that regress these behaviors. It also checks core accessibility semantics, including document language, a single page heading, table captions and header scope, and accessible architecture figures.

These checks use Python's standard library only. The skill does not require or ask users to install Vale, Pa11y, axe, Lighthouse, Chromium, or npm packages.

## Scope

This PR changes GCP-to-AWS decision and full migration reports. Heroku-to-AWS reports use a separate report specification and validator and are intentionally unchanged.

The repository packages the migration skill in two locations, so the same changes are synchronized under:

- `advisor/plugins/aws-startup-advisor`
- `migrate/plugins/migration-to-aws`

## Verification

- [x] `python3 -m pytest advisor/plugins/aws-startup-advisor/tests/test_validate_migration_report.py -q` — 55 passed
- [x] `python3 -m pytest migrate/plugins/migration-to-aws/tests/test_validate_migration_report.py -q` — 55 passed
- [x] `node advisor/plugins/aws-startup-advisor/tools/cross-plugin-drift.ts`
- [x] Full Python suites for both plugin copies — 60 passed each
- [x] Updated full-report and decision-report fixtures pass their validators
- [x] Added regression coverage for recommendation order, ambiguous stay wording, verdict pills, leadership summary, glossary placement, and table accessibility semantics


advisor/plugins/aws-startup-advisor/scripts/validate-migration-report.py
```diff
@@ -607,11 +714,13 @@ def validate_report(
     if check_readability:
         errors.extend(_validate_readability(html))
         errors.extend(_validate_exec_vocabulary(html))
+        errors.extend(_validate_decision_language(html))
         # Normal generated reports require a TOC. Use the same signal to
         # enforce the visual shell while preserving --no-require-toc as the
         # lightweight escape hatch for deliberately minimal unit fixtures.
         if require_toc:
             errors.extend(_validate_visual_contract(html))
+            errors.extend(_validate_accessibility(html))
```
leon1418: [🤖 AI review 🤖] `--no-readability` currently disables the new accessibility gate as a side effect because `_validate_accessibility` is nested under `check_readability`. That flag is documented and described by the CLI as skipping prose/readability checks, not HTML semantics. I verified that removing `<html lang="en">` from the reference fixture still produces no errors when `check_readability=False`; the same bypass applies to the single-`h1`, caption/scope, and figure-alternative checks. Please run accessibility independently of `check_readability` (while preserving the intended minimal-fixture escape hatch) and add a regression test for `--no-readability`.

advisor/plugins/aws-startup-advisor/scripts/validate-migration-report.py
```diff
@@ -458,6 +481,48 @@ def _validate_action_lists(html: str) -> list[str]:
...
+    if estimation_infra is None and estimation_ai is None:
+        return []
+    section = _section_html(html, "exec-share")
+    if section is None:
+        return [
+            "monthly estimation artifacts exist but no <section id=\"exec-share\"> "
+            "copy-ready leadership brief was rendered"
+        ]
+    errors: list[str] = []
+    if not re.search(r'class=["\'][^"\']*\bshare-card\b', section, re.I):
```
leon1418: [🤖 AI review 🤖] This verifies only that a `share-card` class exists, so `<div class="share-card"></div>` passes even though the PR promises a copy-ready, standalone leadership paragraph. I confirmed the empty-card mutation returns no validation errors. Please require substantive text content (and the intended single-paragraph structure) and add a negative test; otherwise the main new CFO-summary quality gate can silently pass an unusable report. Apply the same fix to the mirrored migrate copy.

leon1418 (commented):
[🤖 AI review 🤖] The decision-first report structure is a meaningful UX improvement, and both mirrored plugin copies remain synchronized. I found three mandatory validator gaps that let the newly documented quality contract regress while still returning REPORT_OK: accessibility can be disabled through --no-readability, the leadership brief may be empty, and “Stay entirely if” is not actually required. Please close these enforcement gaps and add negative regression tests in both mirrored test suites.\n\nValidation run at this head:\n- `python3 -m pytest advisor/plugins/aws-startup-advisor/tests/test_validate_migration_report.py -q` — 51 passed\n- `python3 -m pytest migrate/plugins/migration-to-aws/tests/test_validate_migration_report.py -q` — 51 passed\n- `python3 -m pytest advisor/plugins/aws-startup-advisor/tests -q` — 56 passed\n- `python3 -m pytest migrate/plugins/migration-to-aws/tests -q` — 56 passed\n- Node 24 drift check — OK (256 identical, 25 allowlisted across 6 skill trees)\n- Direct mutation probes confirmed all three invalid reports are currently accepted.\n\nGitHub reports no CI checks for this PR. Open PR #167 overlaps two migrate-side files (the reference report and report-generation spec), so merge order may require conflict resolution, but there is no semantic dependency.

advisor/plugins/aws-startup-advisor/scripts/validate-migration-report.py
```diff
@@ -458,6 +481,48 @@ def _validate_action_lists(html: str) -> list[str]:
     return errors
 
 
+def _validate_decision_language(html: str) -> list[str]:
+    """Keep whole-stack decisions distinct from track-scoped holds."""
+    summary = _section_html(html, "decision-summary") or ""
+    errors: list[str] = []
+    if re.search(r"<h[1-6][^>]*>\s*Stay\s+if\s*</h[1-6]>", summary, re.I):
```
leon1418: [🤖 AI review 🤖] Rejecting only the exact legacy heading does not enforce the documented replacement. I changed `<h3>Stay entirely if</h3>` to `<h3>Remain on GCP when</h3>` and the complete fixture still validated with no errors; removing the heading entirely has the same gap. When the recommendation/stay list is present, please require exactly one “Stay entirely if” heading (and test the missing/renamed cases), rather than blocklisting one old spelling. Apply the same fix to the mirrored migrate copy.

advisor/plugins/aws-startup-advisor/scripts/validate-migration-report.py
herosjourney: Fixed in 5d2af30. Accessibility validation now runs whenever the normal generated-report contract is enabled (`require_toc`), independently of `--no-readability`. Added a regression test that removes `<html lang>` while disabling readability and confirms validation still fails. Applied to both mirrored validators.

advisor/plugins/aws-startup-advisor/scripts/validate-migration-report.py
herosjourney: Fixed in 5d2af30. `exec-share` now requires exactly one `<p>` inside the share card and at least 20 words of substantive text; empty or structurally invalid cards fail. Updated both fixtures/spec copies and added an empty-card negative regression test.

advisor/plugins/aws-startup-advisor/scripts/validate-migration-report.py
herosjourney: Fixed in 5d2af30. When `recommendation.stay_if` contains reasons, validation now requires exactly one “Stay entirely if” heading instead of only rejecting the legacy spelling. Added negative tests for both renamed and missing headings, mirrored across both plugin copies.

leon1418 (commented):
[🤖 AI review 🤖] Re-review at `976b76217f6036ff822888b48fa8c461b25b6db0`: the three prior mandatory findings are **not fixed in the PR head**. The owner's replies reference `5d2af303f2a290b223c6ffbd80525abdbfcbc8fc`, which contains the intended fixes and regression tests, but that commit is a child of the current head; PR #223 still points to its parent `976b762...`, so none of those changes would merge.

Per finding at the actual head:
- **A — unfixed:** `_validate_accessibility` remains nested under `check_readability`; removing `<html lang>` with `--no-readability` still returns `REPORT_OK` in both plugin copies.
- **B — unfixed:** `_validate_share_section` still checks only for the `share-card` class; an empty card still returns `REPORT_OK` in both copies.
- **C — unfixed:** `_validate_decision_language` still blocklists only exact `Stay if`; renamed and missing headings still return `REPORT_OK` in both copies.

Evidence: targeted validator suites are 51/51 passed in both copies; full plugin suites are 56/56 passed in both copies; Node `v24.15.0` drift check passes (`256 identical, 25 allowlisted across 6 skill trees`). All four direct mutation probes above were accepted by both validators. The claimed new negative tests are absent from both suites at this head, so they cannot guard these regressions. GitHub reports no CI checks. Open PR #167 still overlaps the migrate reference fixture and report-generation spec, so merge order may require conflict resolution.

No additional mandatory or non-blocking findings were identified in the PR-specific files; the blocker is that the verified fix commit is not attached to the PR branch. **Not merge-ready.** Please advance the PR branch to `5d2af303f2a290b223c6ffbd80525abdbfcbc8fc` (or an equivalent commit) and request another re-review.

leon1418:
Pushed [b9dd6c7](https://github.com/awslabs/startups/pull/223/commits/b9dd6c7) — a merge with main to clear the conflict (this branch pre-dates #167's Sonnet 5 sweep). One hunk in `generate-artifacts-report.md` (both copies): kept this PR's content in full — including the "Stay entirely if" rename, which the validator enforces — and updated only the example model label to "Claude Sonnet 5 (recommended)" per #167.

Nice to see both plugin copies maintained in the PR itself — no twin sync needed this time, unlike #180–#186. Verified on the merged tree (incl. 5d2af30's regressions): both suites 60/60, both asserter sets 8/8, drift 256 identical, vendored-shared and dprint green.

az-zhu (approved):

PR #212, opened 2026-08-11
chore: remove aws-dev-toolkit plugin

rsmets:
Removes the `aws-dev-toolkit` plugin per the 2026-08-03 deprecation decision (shelmes@ handoff).

## What changed

- **Deleted** `solution-architecture/plugins/aws-dev-toolkit/` (38 skills, 11 agents, 104 files)
- **Removed** the `aws-dev-toolkit` entry from the `startups-for-aws` marketplace. Required: its `source` pointed at `./solution-architecture/plugins/aws-dev-toolkit`, so leaving the entry would give a listing whose path no longer resolves
- **Updated** the root and `solution-architecture/` READMEs to point users to AWS Startup Advisor for startup-specific guidance and Agent Toolkit for AWS (`aws configure agent-toolkit`) for general-purpose guidance

Net: 3 files modified, 104 deleted.

## Why nothing was ported

The plugin's skills and agents were overwhelmingly general-purpose AWS engineering guidance, which Agent Toolkit for AWS now owns. That duplication is precisely what drove the deprecation, so porting any of it would re-import the overlap. The startup lens belongs to AWS Startup Advisor.

Migration skills (`migration-gcp-to-aws`, `migration-azure-to-aws`, `migration-apprunner-to-ecs-express`) are being consolidated with Startups-Migrate under StartupEngBlend-3027 rather than ported here. `hipaa-assistant` is addressed by the separate compliance-agent initiative.

## Why removing the source now is safe

Deleting the source would normally destroy the only channel for notifying existing users, so ordering matters. That notice has already landed:

1. The deprecation notice merged into the served copy at aws-samples/sample-claude-code-plugins-for-startups#10, released as [v0.4.0](https://github.com/aws-samples/sample-claude-code-plugins-for-startups/releases/tag/v0.4.0)
2. The official marketplace advanced its pinned SHA to that content (`abdf8673` → `05a3889e`, anthropics/claude-plugins-official#5238, merged)

So installed users receive the deprecation notice on their next `/plugin update` regardless of this removal.

Note: the plugin listing description at claude.com/plugins/aws-dev-toolkit is Anthropic-side metadata and is not affected by this PR. Updating that text and delisting the entry both require the Anthropic contact, as neither is self-serve for non-members.

## Follow-ups, deliberately not in this PR

- The replacement plugin for `solution-architecture/`
- The contribution gate (three-criteria review agent plus its CI actuator) that keeps general-purpose content from re-accumulating in this folder

PR #226, opened 2026-08-18
feat(solution-architecture): add aws-startups-solution-architecture plugin with Agent Toolkit for AWS as upstream dependency

rsmets:
## What

Stands up the renamed replacement for the removed `aws-dev-toolkit`, per item 3 of the deprecation handoff (shelmes@, 2026-08-03). Follows #212, which removed the old plugin.

The plugin carries **only** startup-specific solution-architecture practice and takes all general-purpose AWS service depth from [Agent Toolkit for AWS](https://github.com/aws/agent-toolkit-for-aws) as an upstream dependency rather than restating it.

### Skills (3)

| Skill | Purpose |
| --- | --- |
| `startup-architecture-review` | Scope a review to stage and runway, judge findings against the next milestone, deliver a recommendation rather than a finding list. Carries a security floor that is non-negotiable at any stage. |
| `founder-build-loop` | Low-ceremony align/sketch/build loop for founders. Keeps align-before-building and verify-before-generating, drops the requirements docs a founder did not ask for. |
| `startup-design-challenger` | Adversarial pre-build review against the failure modes that end early-stage companies: team shrinks, credits expire, round slips, traffic never arrives, critical component unprotected. |

Each declares `audience: startup` so the contribution gate can check startup-specificity mechanically.

### Upstream dependencies

```json
"dependencies": [
  { "name": "aws-core",   "marketplace": "claude-plugins-official" },
  { "name": "aws-agents", "marketplace": "claude-plugins-official" }
]
```

Two details that are easy to get wrong here:

1. Dependencies resolve **within the declaring plugin's marketplace** by default. These live in `claude-plugins-official`, not `startups-for-aws`, so the root `marketplace.json` now declares `allowCrossMarketplaceDependenciesOn: ["claude-plugins-official"]`. Without it, install fails with a `cross-marketplace` error.
2. They are declared **unversioned on purpose.** Version constraints resolve against `{plugin-name}--v{version}` git tags, and `aws/agent-toolkit-for-aws` currently publishes no tags, so any semver range would fail with `no-matching-tag`.

### Contribution gate

`solution-architecture/CONTRIBUTING.md` encodes the three criteria from item 4 of the handoff. All three must pass; two of three is a rejection. Criterion 2 (no overlap with Agent Toolkit for AWS) is marked load-bearing, since that overlap is what drove the original deprecation.

The AgentCore SME review requirement is documented in that guide, scoped to every PR in the folder rather than to paths matching "agent" (overlap usually arrives inside a skill about something else). It is deliberately **not** in `CODEOWNERS`: GitHub silently ignores an entry naming a team that does not exist or lacks write access, and no agentcore handle exists in this repo yet. The doc says to add it once the handle is confirmed.

## Nothing was ported wholesale

Per the handoff's "port nothing wholesale" guidance. An audit of all 38 removed skills and 11 agents found the startup framing lived entirely in the old `plugin.json` description:

- The word "startup" (meaning a company) appeared in **no** skill or agent body. Every match was container startup latency, `venv/bin/activate`, or "Activate when the user asks".
- **Zero** references to AWS Activate, credits, runway, or funding stage anywhere in the corpus.
- Only `aidlc-lite` had genuine founder framing (~10 lines). Its salvageable insight, the founder-versus-regulated-enterprise process contrast and the max-three-questions default, is folded into `founder-build-loop`.

This plugin also contains no reference to any deprecated or sunset AWS service, avoiding the uncaveated App Mesh, App Runner-as-new-build, S3/Glacier Select, and IoT Analytics recommendations that were present in the removed plugin.

## Verification

- `claude plugin validate` passes for both the plugin manifest and the marketplace manifest.
- End-to-end install from a local marketplace resolved and enabled **both** cross-marketplace dependencies (`+ 2 dependencies: aws-core, aws-agents`), with all three skills present in the install cache. Test install and marketplace were then removed.
- `dprint fmt` idempotent; `markdownlint` clean across 859 files.

## Note for reviewers

`architect-for-startups` in AWS Startup Advisor already covers **founder-facing** stage, runway, and credits advice, and its references overlap the old dev-toolkit content by name. This plugin is deliberately scoped to the **SA-facing** engagement workflow instead (how to run the review, how to challenge a design), which neither the advisor nor Agent Toolkit for AWS covers. Worth a second opinion on whether that boundary is drawn where you would draw it.

## CODEOWNERS

Touches `marketplace.json`, `**/.claude-plugin/`, and `**/SKILL.md`, so this needs `@awslabs/startups-admins` in addition to `@awslabs/startups-solution-architects`.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -4,6 +4,7 @@
     "name": "Amazon Web Services"
   },
   "description": "AI-powered build and migrate guidance from AWS, built for startups at any stage.",
+  "allowCrossMarketplaceDependenciesOn": ["claude-plugins-official"],
   "plugins": [
     {
       "name": "migration-to-aws",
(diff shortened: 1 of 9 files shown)


igkins (approved):
lgtm

rsmets:
PR opened prematurely

PR #228, opened 2026-08-18
feat(agent-advisor): AgentCore Runtime Instances — >8h, GPU, and instance-type workloads no longer eliminate AgentCore

leon1418:
## Problem

AWS launched the **Instances compute type** for AgentCore Runtime ([News Blog, 2026-08-06](https://aws.amazon.com/blogs/aws/runtime-instances-persistent-compute-for-production-ai-agents-on-amazon-bedrock-agentcore/)): AWS-managed EC2 via capacity providers, **sessions up to 14 days** (`maxLifetime` ≤ 1,209,600s per the [lifecycle docs](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/runtime-lifecycle-settings.html)), **GPU** and broad EC2 instance-type choice, multi-agent shared hosts, and EC2 + management-fee billing.

The agent-advisor still modeled AgentCore as microVMs-only and **hard-eliminated it on three now-false grounds**: "8hr session cap", "no GPU support", "capped at 2 vCPU / 8 GB". Verified against the current engine: `session_duration=over_8hr` flipped the verdict to ECS with AgentCore not even scored. Every customer with a >8h, GPU, or instance-type-specific agent was steered away from a runtime that now serves exactly those workloads.

## Solution

Per PM guidance, the flow now asks two questions and **routes instead of eliminating**:

1. **">8 hours?"** — the existing `session_duration` question; `over_8hr` no longer eliminates AgentCore, it routes it to the Instances compute type.
2. **"Particular instance type?"** — new `instance_type_requirement` dimension (yes/no/unknown). `yes` hard-eliminates Lambda and Lambda MicroVMs (no instance selection exists there) and routes AgentCore to Instances.

**Engine (`scoring.py`):** new dimension; `_select_agentcore_compute_type()` emits `agentcore_compute_type: microvms | instances | null` — instances iff `over_8hr` OR `gpu`/`heavy_non_gpu` OR instance-type `yes`; an instances verdict carries a warning (14-day ceiling — an always-on service still routes to ECS/EKS via traffic affinity; EC2+fee billing, NOT consumption / $0-I/O-wait; Linux only; limited launch regions — verify via MCP).

**Profiles:** `agentcore.json` drops all three hard constraints (`over_8hr` affinity 0→4, gpu/heavy 0→3 — ECS/EKS keep the edge on pure compute, and platform gravity still wins: gpu + existing EKS → EKS, covered by a golden test); lambda/lambda_microvms gain the instance-type constraint; volatile facts split per compute type.

**Downstream, end to end:** clarify (both audience wordings), confirm (states the Instances cost model before the user confirms — nobody reaches Design believing they get microVM pricing), design (compute type copied verbatim, never re-derived; the $0-I/O-wait TCO note is now explicitly microVMs-only), estimate (Instances = EC2 rate via awspricing MCP + management fee), poc (a microVMs-shaped POC for an instances-routed unit must declare it does not exercise the routing property, with capacity-provider next steps), recommendation-doc (names the compute type and which answer routed there).

**decision-refs:** `agentcore.md` rewritten with a two-compute-type table, plus the companion [Gateway rate limits launch](https://aws.amazon.com/blogs/machine-learning/configure-rate-limits-for-ai-traffic-on-agentcore-gateway/) (RPS/RPM + TPM + CPS; evaluated before Policy; fail-open — a cost/fairness control, not a security boundary); `ecs.md`, `lambda-microvms.md`, `workload-classes.md`, `temporal.md`, `freshness.md` destaled. The batch-never-AgentCore rule survives with a corrected justification.

**Mirror sync (second commit):** the aws-startup-advisor copy re-synced per the cross-plugin drift gate — 16 files re-copied with prefix rewrites, plus the same edits hand-ported into 6 allowlisted files (the gate doesn't force those, but leaving them stale is the silent-divergence failure the gate exists to prevent).

### Tests

305 passed (was 302), both copies. The old `test_golden_over_8hr_eliminates_agentcore_and_microvms` asserted the stale behavior and is replaced by four goldens: >8h routes to instances with the warning; instance-type requirement eliminates only the Lambda family; the compute-type selection matrix; GPU no longer eliminates AgentCore but platform gravity holds. Heavy+high-launch is now a genuine agentcore+microvms co_recommend — the 5 TPS warning it exists for still fires.

### Note for maintainers

`mise run build` passes every task on this tree. One caveat for full disclosure: the local aggregate run also flagged 3 gitleaks findings — all in a commit that exists only on an unrelated local branch (not an ancestor of this PR and not on `main`; gitleaks scans `--all` refs locally). Scoped to this PR's range (`upstream/main..HEAD`), gitleaks reports no leaks, and CI checks out only the PR branch.

## Type of Change

- [ ] New plugin/power/tool
- [ ] Bug fix
- [ ] Documentation update
- [ ] Guardrail/CI update

## Team Folder

- [ ] `advisor/`
- [x] `migrate/`
- [ ] Other: `advisor/` is also touched, but only as the drift-gate-mandated mirror of the `migrate/` change

## Checklist


leon1418 (commented):
[🤖 AI review 🤖] **Design-first review of PR #228** — AgentCore Runtime Instances

## Summary
This PR correctly addresses the stale AgentCore constraints that were steering customers away from a runtime that now supports their workloads (>8h sessions, GPU, instance-type choice) via the new Instances compute type launched 2026-08-06.

**Head SHA:** `40cfa0c7d3b5c903d5d3f3e1a8b8a22023222273`
**Size:** 44 files, +514/-122
**Scope:** Engine (`scoring.py`), runtime profiles, all downstream phases (clarify, confirm, design, estimate, poc, recommendation-doc), decision-refs, and the advisor mirror sync.

## Design Assessment ✅

The design is sound:
1. **Routing-not-elimination pattern**: `_select_agentcore_compute_type()` routes workloads to the right compute type rather than eliminating AgentCore entirely. This is the correct approach — workload characteristics should select the right shape within a viable runtime.
2. **Hard constraint cleanup**: Removing the three now-false hard constraints (`over_8hr`, `gpu`, `heavy_non_gpu`) aligns the engine with the current AgentCore capabilities.
3. **Affinity rebalancing**: `over_8hr` affinity goes from 0→4, `gpu`/`heavy_non_gpu` go from 0→3. ECS/EKS retain higher affinities (4-5) for heavy compute, preserving the "platform gravity" rule for existing clusters.
4. **Warning propagation**: The Instances warning carries critical information (14-day ceiling, EC2 + management-fee pricing, Linux only, limited regions) that prevents downstream phases from making microVM assumptions.
5. **POC guard**: The explicit compute-type guard in poc.md prevents shipping a microVMs-shaped POC for an Instances-routed workload without disclosure.
6. **Volatile facts split**: `session_cap` → `microvms_session_cap` + `instances_session_cap` maintains freshness verification granularity.

## Tests ✅

58 tests pass on both `migrate/` and `advisor/` copies:
- `test_golden_over_8hr_routes_agentcore_to_instances_not_elimination` — validates the new routing
- `test_golden_instance_type_requirement_eliminates_lambda_family_only` — validates Lambda/MicroVMs elimination
- `test_golden_compute_type_selection_matrix` — covers the full routing matrix
- `test_golden_gpu_no_longer_eliminates_agentcore_but_platform_gravity_holds` — validates platform gravity still wins

**Mutation probes (local):** Breaking `over_8hr` check or `gpu/heavy_non_gpu` check causes test failures, confirming adequate regression coverage.

## CI Status ✅
8/8 checks passed: build, gitleaks, bandit, semgrep, checkov (×2 each).

## Cross-Plugin Sync ✅
All 22 matching files between `migrate/` and `advisor/` copies are byte-identical or within the allowlisted prefix divergences (e.g., "migration-to-aws pricing cache" vs "llm-to-bedrock skill's pricing cache").

## Findings

**0 mandatory blockers.**

**1 Nit (non-blocking):**
- scoring.py L195: The hardcoded "8 regions at launch" could be extracted to `volatile_facts.instances_regions` to maintain single source of truth. Current implementation is correct; this is a maintainability suggestion.

## Positive Observations
- The test docstrings are excellent — they explain *why* the test exists and what the business rule is, not just what it asserts.
- The PR description comprehensively documents the problem, the solution, the PM-directed flow changes, and the testing strategy.
- The batch-never-AgentCore rule survives with corrected justification (session-scoped billing applies to Instances too).

## Merge Readiness
**CLEAN / MERGEABLE** — no mandatory blockers, tests pass, CI green, cross-plugin sync verified.

leon1418:
Nit addressed in `8338015` — and it deserved better than nit status, so thank you for flagging it.

**The count was not just hardcoded; it was unverifiable.** The launch blog's own text says "8 regions" while enumerating nine region names (Ohio, N. Virginia, Oregon, Mumbai, Singapore, Sydney, Tokyo, Frankfurt, Ireland). I had noticed that discrepancy during research, decided not to state a count, and let it slip into the warning string anyway.

**Fix shape:** rather than moving the number into `volatile_facts.instances_regions`, the warning now *points at* that fact — which already exists with exactly the right contract (`"value": [], "verify_via_mcp": true`): "launch-region set is limited and volatile — verify current availability via MCP (volatile_facts.instances_regions)". A verify-don't-restate pointer keeps one source of truth and nothing to drift; a copied value would have been a second one.

Applied to both plugin copies; 305 tests pass on each; cross-plugin drift gate OK; dprint clean.

leon1418 (commented):
[🤖 AI review 🤖] **Follow-up review of PR #228** — fix commit `8338015e`

## Update Summary

| Field | Value |
|-------|-------|
| **Prior reviewed head** | `40cfa0c7d3b5c903d5d3f3e1a8b8a22023222273` |
| **Current head** | `8338015e1467586257e973e4864a8bbfcf0692aa` |
| **Commits since prior review** | 1 |
| **Files changed** | 2 (`scoring.py` in both plugin copies) |
| **Delta** | +6/-4 lines |

## Commit Analysis

**`8338015e` — fix(agent-advisor): don't hardcode the Instances launch-region count in the warning**

This commit addresses the Nit from my prior review (hardcoded "8 regions at launch" could drift).

The fix is sound:
- Removes the hardcoded region count (`"8 regions at launch"`)
- Points to the SSOT: `volatile_facts.instances_regions`
- Uses language that acknowledges volatility: "launch-region set is limited and volatile"
- As the commit message notes, the launch blog actually listed **9** region names while saying "8 regions", so this was not just a single-source-of-truth issue but an actual data integrity slip

**Before:**
```
"Linux only at launch; verify region availability via MCP "
"(8 regions at launch)."
```

**After:**
```
"Linux only at launch; launch-region set is limited and volatile — "
"verify current availability via MCP (volatile_facts."
"instances_regions)."
```

## Prior Nit Disposition

✅ **RESOLVED** — the hardcoded region count is removed and replaced with a volatile_facts reference.

## Validation

| Check | Result |
|-------|--------|
| Both plugin copies identical | ✅ `diff` confirms `scoring.py` warning text matches |
| migrate tests | ✅ 58/58 passed |
| advisor tests | ✅ 58/58 passed |
| Instances warning fires correctly | ✅ Verified with `entry_point: migrate, session_duration: over_8hr` |
| Warning contains `volatile_facts.instances_regions` | ✅ Confirmed |
| Warning does NOT contain hardcoded "8 regions" | ✅ Confirmed |
| CI | ✅ 8/8 checks pass (build, gitleaks, bandit, semgrep, checkov ×2) |

## Prior Review Behaviors (Still Valid)

The fix commit is purely cosmetic to the warning string. I verified the core behaviors from my prior review are unaffected:
- Routing-not-elimination pattern: `over_8hr` routes to `agentcore_compute_type: instances`, does not eliminate AgentCore
- Platform gravity: GPU + existing_eks still routes to EKS
- Lambda family elimination: `instance_type_requirement: yes` eliminates Lambda/Lambda MicroVMs only
- POC guard and warning propagation: unchanged

## Findings

**0 mandatory blockers.**
**0 Nits.**

The prior Nit is resolved. No new issues introduced.

## Merge Readiness

**CLEAN / MERGEABLE** — all prior findings addressed, tests pass, CI green, cross-plugin sync verified.

az-zhu (dismissed):

leon1418:
Rebased onto current `main` — the conflict was a genuinely interesting one, so documenting the resolution.

**What collided:** while this PR waited, the evidence-binding work (readiness gates / "Bind verification evidence") landed on `main` and solved the *same stale-fact problem from the other side*: it kept the three AgentCore hard constraints but made them **verification-gated** (`verification_required: true`, keyed facts like `agentcore.session_cap = "8h"`, defer-until-verified). This PR deletes those constraints outright.

**Why deletion still wins for these three:** verification gating protects against *stale* facts; these constraints are *semantically* wrong. `agentcore.session_cap = "8h"` still verifies as TRUE today against the microVMs docs — and would then eliminate AgentCore *with evidence in hand*, reproducing the exact false elimination this PR exists to fix. Instances didn't change the fact's freshness; it changed what the fact implies (8h caps one compute type, not the runtime).

**What was kept and merged:**
- The verification machinery is untouched and now composes with this PR: `_collect_warnings` carries both parameter sets, the Lambda-family caps stay verification-gated (deferred without current-run evidence, final with it), and `selection_verification_requirements` still governs MicroVMs launch capacity.
- `agentcore.json` volatile facts are the union, **compute-type-scoped**: `microvms_session_cap` / `microvms_compute_cap` / `microvms_gpu_support` (preserving main's `verification_key` convention) + `instances_session_cap` / `instances_regions`; `io_wait_billing` kept and annotated microVMs-only — which composes with main's design.md rule (evidence-gated I/O-wait claims) and this PR's (compute-type-gated).
- Tests are the union: 318 pass on both copies (this PR's four goldens adapted to deferral semantics — e.g. without evidence the Lambda-family now *defers* rather than eliminates, and the over-8h golden additionally asserts AgentCore appears in **neither** eliminated **nor** deferred: deleted, not gated). Main's three high-launch machinery tests keep all their verification assertions with one merged-semantics change: heavy+high-launch is now a genuine `co_recommend {agentcore, lambda_microvms}` since heavy compute no longer eliminates AgentCore.

Full gates re-run post-rebase: both suites 318, cross-plugin drift OK, dprint clean, bandit/semgrep/checkov OK, gitleaks scoped to this PR's range clean.

leon1418:
**Note on the force-push (`3d85bbeb` → `44232621`):** the web-UI conflict merge from 08-18 was replaced by the rebase documented above, deliberately. That merge resolution had taken this PR's side of `scoring.py` and the runtime profiles wholesale — which silently reverted the evidence-binding / readiness-gates work that had just landed on `main` (0 references to `_evaluate_hard_constraints` / `run_evidence` / `deferred_verification` remained in its `scoring.py`; the rebase preserves all of it and composes the two features). It was also based on the 08-18 `main`, missing this week's merges. Nothing from that merge commit is lost — it contained only the conflict resolution, and it remains recoverable at `3d85bbeb`. Post-rebase verification is in the comment above (318 tests both copies, drift OK, all gates green).

az-zhu (approved):

PR #237, opened 2026-08-21
fix(agent-advisor): GPT-5.6 gained a runtime/CRIS path — same-model governance, not a forced family switch

leon1418:
## Why

The OpenAI provider catalog (#195, `verified_at: 2026-07-21`) asserts GPT-5.x on Bedrock is **Mantle/Responses-only with no cross-region inference**, and its `runtime_converse` path therefore "uses a Bedrock-native model, **not** a GPT-5.x model". The current model cards (2026-08-21) invalidate all of that for the GPT-5.6 family:

- Sol/Terra/Luna carry a second `bedrock-runtime` Programmatic Access row served through **CRIS inference profiles** (Geo `us.`/`in.`, Global `global.`), with the cards recommending runtime for new applications
- `Converse`/`Chat Completions`/`Invoke` are listed as supported there (Guardrails Converse-only, prompt caching Responses-only)
- **Global CRIS pricing is published at OpenAI's standard list price — cost parity** (In-Region/Geo remain 1.10×)
- Context window is **1M** (272K was the 7/21 value)

**The practical consequence was a wrong recommendation:** any GPT-5.6 customer asking for runtime governance (Guardrails, invocation logging) was told to **switch model families**, when their own model now satisfies the requirement on `bedrock-runtime`. GPT-5.5 / GPT-5.4 re-verify as mantle-only and in-region-only, so every change here is split by family.

This is the same 2026-08-10 → 2026-08-21 documentation drift being addressed on #210 for the `gcp-to-aws` / `llm-to-bedrock` side (see @herosjourney's review there); this PR is the `agent-advisor` provider-module counterpart and is independent of #210 — both cite the AWS model cards directly.

## What changed

**Catalog** — re-issued as `openai-bedrock-2026-08-21.json` (old file removed):
- 5.6 entries gain a `runtime_converse` path (`available: true`, bare `model_id`, `requires_cris: true` — the existing resolver derives `us.`/`in.`/`global.` ids from the residency posture, so India geo support falls out for free) and `context_window: 1000000`
- Paths-level constraints replace "Geo/Global not supported per the AWS model cards" with the per-family, per-endpoint reality
- 5.5/5.4 entries carry a dated re-verification note (still mantle-only; no CRIS, no runtime, no 1M tier)
- **No schema change** — 5.6-on-runtime reuses the existing `runtime_converse` `api_path`, which is literally what it is

**Engine** (`openai_model_recommendation.py`):
- `_converse_candidate_order` inserts the same-model key ahead of the Claude tier for 5.6 sources; the Claude tier map (Sol→Opus 4.8, Terra/5.5/5.4→Sonnet 5, Luna→Haiku 4.5) survives as the cross-family fallback
- The decision-option reason and `runtime_required` rationale distinguish "same-model governance path via CRIS" from "family switch required (5.5/5.4 have no runtime path)"

**Tests** — 320 passed per plugin. Five tests inverted with the facts and say so in-line: the four tier-map tests now assert 5.6 sources keep their own model under governance (5.5 still lands on Sonnet — no runtime path), and the min-context probe moves from 400K (now legitimately satisfiable at 1M) to 2M. The capability-evidence fallback mechanism keeps coverage via an explicit candidate order, since no source-derived order starts at Haiku anymore.

## Verification

`drift:check`, frontmatter validator ×2, `shared:check` ×2, asserters ×2, model-id lint ×2, dprint — all pass. Both plugin copies updated together per the drift gate.

**Not verified live:** no billable call. `verify_model_path.py`'s existing `runtime_converse` probe (boto3 Converse with the CRIS id) covers the new path mechanically, but nobody has run it against a real account for a GPT-5.6 CRIS id yet.

## Sequencing

Independent of #210 (different skill, different files; both refresh from the same model cards). The `model-selection.md` cross-reference to #210's `openai-on-bedrock.md` fact file can be added in whichever lands second.


herosjourney (commented):
Reviewed as the counterpart to my review on #210. All factual claims were re-checked against the live AWS model cards on **2026-08-21**, including a fresh spot-check of the [Terra card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-openai-gpt-56-terra.html) for the `in.` India geo id (confirmed: `in.openai.gpt-5.6-terra`, matching Luna — and the catalog correctly gives Sol only `us.`/`global.`).

**Overall: I agree with this PR.** The family split (5.6 dual-path vs 5.5/5.4 mantle-only), the same-model-first candidate ordering, and the preserved Claude cross-family fallback are exactly the right shape for the documentation drift. Nothing below is directionally wrong; two items are worth fixing before merge.

## What verifies against the current docs

- Every catalog claim matches the model cards: 5.6 `runtime_converse` rows with CRIS ids (Geo `us.`/`in.`, Global `global.`), Guardrails Converse-only, prompt caching Responses-only on runtime, Global CRIS at OpenAI standard-price parity with In-Region/Geo at 1.10×, 1M context with the long-context tier at 2.0× in / 1.5× out, and 5.5/5.4 re-verified as mantle-only, in-region-only with dated notes.
- The engine change is minimal and well-scoped: `_same_model_runtime_key` fires only for 5.6 sources, the Claude tier map survives as fallback, and the rationale strings honestly distinguish "same-model governance via CRIS" from "family switch required."
- The mantle-side Chat Completions caution ("listed as a supported API but mantle-side availability unprobed — verify in the target account") is appropriately careful; the card's API list doesn't disambiguate per endpoint.

## Worth fixing before merge

**1. `test_unknown_limits_do_not_pass_hard_numeric_requirement` was weakened, not just inverted.**
The original asserted that unknown limits produce an `unverified_capacity` block. The new version moves the probe from 400K to 2M but silently drops the `assert "unverified_capacity" in _codes(...)` line, keeping only `decision_status == "decision_required"`. With all 5.6 context windows now known, a 2M requirement likely fails on a *different* block code, so the test name no longer matches what it tests — and the "unknown limits don't satisfy hard numeric requirements" invariant is now untested entirely. `output_token_ceiling` is still `"unknown"` on every entry and could probe the original invariant. Suggest: rename the 2M test to what it actually checks, and keep/restore an unknown-limit probe against the output-token ceiling.

**2. Factual error in the Terra and Luna `capability_source` annotations.**
Both say "context window is 1M per the current card (**272K was the 2026-07-21 value**)" — but the 2026-07-21 catalog had `"unknown"` for Terra and Luna; only Sol carried 272000. The preceding sentence "Context window not captured from the card — treat as unknown" was also left in place, directly contradicting the new `context_window: 1000000` two lines above. For a catalog whose value proposition is dated-evidence accuracy, the provenance note should be exact: for Terra/Luna it's `unknown → 1M`, for Sol it's `272K → 1M`.

## Minor

3. **Residency edge case (question):** the same-model candidate is inserted unconditionally for 5.6 sources, but 5.6-on-runtime is CRIS-only (no in-region form). For a workload requiring strict in-region processing *plus* runtime governance, the same-model path is unavailable. Does the engine gate the candidate on residency posture, or could it surface an impossible combination? The catalog note says the resolver derives ids from posture, but I couldn't confirm a hard in-region gate from this diff.
4. **"Global-CRIS cost parity" is stated unconditionally** in the decision-option reason, even when a Geo-scoped posture will land at 1.10×. Suggest conditioning the wording: "cost parity on Global CRIS; 1.10× on Geo/In-Region."
5. **`model_key.startswith("openai_")` in `_decision_options`** would misdescribe a future `openai_gpt_oss_*` runtime candidate — which the catalog disclaimer explicitly contemplates adding — as a "GPT-5.6 same-model governance path." Comparing against `_same_model_runtime_key(source)` would be exact.
6. **`model-selection.md` convergence (question):** the two plugin copies are converged to identical content, which makes the advisor plugin's copy reference `migration-to-aws/skills/gcp-to-aws/...` even though `aws-startup-advisor` has its own sibling `gcp-to-aws` skill. Intentional single-canonical-source, or should each copy point at its own plugin?
7. Trivia: missing newline at EOF in the test file; missing period before "Re-verified" in the gpt-5.4 `capability_source`; and the live-verification gap the PR already flags — one real Converse call against a `us.openai.gpt-5.6-*` CRIS id would retire the largest remaining risk (same note as #210).

---
*This review was generated with AI assistance (verified against AWS documentation via the AWS Knowledge MCP server).*

herosjourney:
## Conflict check

GitHub reports this PR as `CONFLICTING` against `main`, but a from-scratch local merge test says otherwise.

I fetched the current `main` (`e67d6b5`, includes #249 and #250) and this PR's head (`c8d7678`), then ran a real three-way merge locally:

```
git merge origin/main --no-commit --no-ff
```

Result: **all 40 changed files auto-merged with zero conflict markers.** Cross-checked independently with `git merge-tree` (which simulates the merge without touching the working tree) — same result, every file reports `merged` cleanly.

**Likely explanation:** this PR's base commit (`0b59caf`) predates both #249 and #250, which merged into `main` after this branch was opened. GitHub's mergeability status is computed lazily and doesn't always recompute immediately after the base branch moves — it often needs something to poke the PR (a push, or a manual recheck) before it re-evaluates.

**Suggested next step:** pushing an empty commit or re-syncing the branch should force GitHub to recompute mergeability, and it's likely to clear on its own without any manual conflict resolution needed. Happy to redo this merge test if it still shows conflicting after that, and hand over the resulting diff.

leon1418:
Both threads addressed in one push:

**Re: the review ([ebd5280f](https://github.com/awslabs/startups/pull/237/commits/ebd5280f)) — all seven taken:**

1. **Weakened test** — you were right that it was weakened, not just re-probed. The 2M case is renamed to what it actually checks (`test_context_requirement_beyond_every_catalog_window_requires_decision`), and the fail-closed invariant gets its own probe against `output_token_ceiling` (still `unknown` on every entry), asserting `unverified_capacity` again.
2. **Terra/Luna provenance** — exact now: "the 2026-07-21 card did not state it — cataloged as unknown then"; the contradictory leftover sentence is gone. Sol keeps its correct 272K note.
3. **Residency edge** — your question exposed a real gap: it was a silent `invocation_model_id: None`. Now a hard gate: `runtime_converse` + `requires_cris` + no resolvable CRIS form emits `cris_residency_unresolved` [BLOCKS] with the three ways out (relax residency / explicit `inference_profile_id` / Claude cross-family path). Regression test added.
4. **Parity wording** — "cost parity on Global CRIS (1.10x on Geo/In-Region pricing)".
5. **`startswith("openai_")`** — now compares against `_same_model_runtime_key(source)` exactly.
6. **model-selection.md convergence** — not intentional; the flattening was a mistake. The advisor copy's in-plugin references are restored (`gcp-to-aws` / `llm-to-bedrock` skills live in the same plugin) — this is precisely the deliberate divergence the drift allowlist exists to permit.
7. EOF newline and the missing period fixed. The live Converse probe against a `us.openai.gpt-5.6-*` id remains the honest residual risk — same as #210.

**Re: the conflict check** — diagnosis adopted with thanks: [c640f266](https://github.com/awslabs/startups/pull/237/commits/c640f266) merges current main (zero conflicts, exactly as your `merge-tree` test predicted), which doubles as the poke for GitHub's lazy mergeability recompute.

Verified on the merged tree: 322 tests on both plugin copies (+3 new), drift 260 identical, model-id lint, dprint all green.

herosjourney (approved):
Re-reviewed all 7 items from my prior review against the current code — 6 of 7 confirmed fixed exactly as described, 1 cosmetic miss (missing period, noted in the comment above, non-blocking). Approving.

herosjourney:
## Re-review: all 7 items checked against current code

I re-verified each of the 7 items from my earlier review against the actual merged diff (not just the response text) — pulled `openai-bedrock-2026-08-21.json`, `openai_model_recommendation.py`, `test_openai_model_recommendation.py`, and both `model-selection.md` copies at the current PR head (`c640f26`).

**6 of 7 confirmed fixed as described:**

1. **Weakened test** — confirmed. Renamed to `test_context_requirement_beyond_every_catalog_window_requires_decision`, and a dedicated `unverified_capacity` assertion re-probes the fail-closed invariant.
2. **Terra/Luna provenance** — confirmed. Now reads "the 2026-07-21 card did not state it — cataloged as unknown then"; the old contradictory "treat as unknown" sentence is gone; Sol's distinct 272K note is untouched.
3. **Residency edge case** — confirmed, and more substantial than a typical fix: a real `cris_residency_unresolved` [BLOCKS] gate with three concrete remediation paths, plus a dedicated regression test (`test_cris_only_path_with_unresolvable_residency_blocks_explicitly`). Good catch.
4. **Parity wording** — confirmed exact: "cost parity on Global CRIS (1.10x on Geo/In-Region pricing)."
5. **Exact-match instead of `startswith`** — confirmed; `_decision_options` now compares against `_same_model_runtime_key(source)` directly.
6. **model-selection.md divergence** — confirmed correctly restored. `advisor/`'s copy references its own in-plugin `gcp-to-aws`/`llm-to-bedrock`; `migrate/`'s copy references `migration-to-aws/skills/gcp-to-aws/...`. `cross-plugin-drift.ts` still passes clean (260 identical, 25 allowlisted) with this divergence intact.

**1 of 7 not actually fixed, despite being listed as addressed:**

7. **Missing period in the gpt-5.4 `capability_source` string** — still present: "...limits not stated Re-verified 2026-08-21..." has no period between "stated" and "Re-verified." The EOF-newline half of this item is fixed; the period was missed.

**Independently re-verified:**
- `pytest test_openai_model_recommendation.py` → 54 passed
- `cross-plugin-drift.ts` → OK (260 identical, 25 allowlisted)
- Conflict status: `mergeable: MERGEABLE` now, confirming the earlier conflict-check diagnosis and the `c640f266` main-merge resolved it as predicted

## Recommendation: Approve, with one trivial nit

The substance is sound — the catalog facts check out against the model cards, the engine change is minimal and well-scoped, and the one open question I raised (residency edge case) got a real fix with test coverage rather than a hand-wave. The single miss is cosmetic (a missing period) and doesn't block correctness or merge — flagging it for a follow-up touch-up rather than holding up the PR.

az-zhu (approved):

PR #239, opened 2026-08-24
Define Heroku application source review contracts

amjadsy:
## Problem

The Heroku-to-AWS skill can collect Heroku account metadata, Terraform, `Procfile`, and `app.json` information, but it does not yet have a stable format for a future application-code review to report its findings. Without that
boundary, later source-review work could return inconsistent or unbounded data, include information that should not be retained, or mix code observations with destination decisions.

## Solution

Add a data-only request and findings contract for 22 application-code questions. The request exposes only selected question names, application identity, and an allowlisted projection of non-secret Heroku inventory. Findings use bounded, question-specific values with explicit present, absent-within-reviewed-scope, unknown, and not-applicable states.

The contract has no fields for configuration values, credentials, connection strings, or source excerpts. Optional source locations must be safe relative paths. Focused tests cover all 22 questions and reject malformed, duplicate,
unrequested, out-of-bounds, secret-bearing, or internally inconsistent data.

Both published Heroku skill surfaces contain byte-identical contract and documentation files. 

## Broader Context

This is the second incremental change in a larger effort to make the existing Heroku migration workflow use application-code evidence when assessing Elastic Beanstalk. PR #231 removes unsafe generated port and health-check assumptions. This PR defines what a future read-only source reviewer may return. Later, separate PRs will run that review, validate and combine its findings, calculate deterministic Beanstalk compatibility, improve cost and reporting, and gate starter infrastructure generation.

The existing plugin structure and its Elastic Beanstalk, Fargate, and EKS destinations remain in place.

## Scope

This PR does not run source review or change Discover, Clarify, Design, Estimate, Generate, Feedback, or Workshop behavior. It does not change destination selection, cost calculations, reports, or generated artifacts. It is independent of #231 and leaves the current customer workflow usable on its own.

The schema remains dormant until a later PR connects source review to the workflow. The test-only schema checker supports the subset used here and does not add a runtime dependency.

## Type of Change

- [ ] Bug fix
- [ ] New plugin/power/tool
- [ ] Documentation update
- [x] Guardrail/CI update

## Team Folder

- [x] `advisor/`
- [x] `migrate/`
- [x] Other: root test-task registration

The `advisor/` and `migrate/` files are the repository's synchronized copies of the same Heroku skill. The focused contract test is stored once under the canonical migration plugin test tree and verifies both copies.

## Validation

- `mise run fmt:check`
- `mise run lint`
- `mise run security`
- Eight focused contract cases covering all 22 application-code questions
- Positive and negative checks for status handling, unsafe paths, missing and unrequested findings, malformed values, broken references, and forbidden value-bearing fields
- Cross-plugin byte-for-byte synchronization checks

GitHub build and security workflows pass. No live Heroku test is included because this PR has no runtime integration or customer-visible behavior.

## Checklist

- [x] My changes do not include hardcoded secrets, credentials, or internal-only
  content
- [x] The repository formatting, lint, test, and security gates pass
- [x] I have updated documentation for the new contract files
- [x] Existing Heroku phases and destination behavior remain unchanged


herosjourney (approved):
## Approving -- verified independently, not just read

This is a dormant, data-only contract (no wiring into Discover/Clarify/Design/Estimate/Generate), so the review bar here is: does the schema actually enforce what the doc and PR description claim, and are the security properties real. Checked both directly rather than trusting the prose.

**Verified by hand-crafting and validating documents against the schema (Python jsonschema, not just reading it):**
- A `runtime_settings` finding with a literal `"value": "postgres://user:pass@host/db"` field is correctly rejected (`additionalProperties: false` on every record type does its job -- no config-value smuggling path).
- A record of the wrong shape for its `question` (e.g. a `processCommandRecord` payload under `runtime_framework`) is correctly rejected -- the per-question `oneOf` branches in `finding` are real constraints, not decorative.
- The `sourcePath` traversal-blocking regex correctly rejects `../`, `..\`, absolute Unix/Windows paths, and double slashes; correctly accepts normal relative paths. The one edge case I flagged myself (`root: "."` being disallowed by the general pattern) turns out to already be handled -- `runtimeFrameworkRecord.root` explicitly carves out `{"const": "."}` as a sibling alternative, so this was already caught and fixed by the author, not an oversight.

**Verified the "UNKNOWN requires a limitation" and "qualified absence" rules exist and are tested**, even though they're not expressible cleanly in JSON Schema draft-07 without heavy duplication -- they're correctly implemented as semantic checks in `validateSemantics()` (test.ts:373-380) with dedicated positive/negative test cases (lines 462-479, 549-561), not silently dropped between the doc and the implementation.

**Ran the actual suite, not just read it:**
- `node --test application-source-contract.test.ts` -> 8/8 pass, matching the PR body's claim exactly.
- `mise run build` -> exit 0. `drift:check`: 261 identical, 25 allowlisted -- the four new/changed files (contract doc, schema, two SKILL.md touches) are confirmed byte-identical between `advisor/` and `migrate/`, matching the PR's "byte-identical" claim rather than just asserting it.
- `mise.toml` diff is exactly the minimal registration described -- one new line in the existing `test` task, no new task, no new runtime dependency.

**Scope claim holds up**: grepped the diff for any change to Discover/Clarify/Design/Estimate/Generate/Feedback/Workshop phase files -- none. This really is additive and dormant as described.

No findings. Clean, well-tested, and the security-relevant claims (no secrets, no config values, no path traversal, no cross-question value confusion) are backed by both schema constraints and semantic tests, not just documentation.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/shared/application-source-contract.schema.json
```diff
@@ -0,0 +1,771 @@
...
+      ]
+    },
+    "text": { "type": "string", "minLength": 1, "maxLength": 500 },
+    "name": { "type": "string", "minLength": 1, "maxLength": 128 },
+    "identifier": { "type": "string", "minLength": 1, "maxLength": 128, "pattern": "^[A-Za-z0-9][A-Za-z0-9._:/-]*$" },
+    "sourcePath": {
+      "type": "string",
+      "minLength": 1,
+      "maxLength": 500,
+      "pattern": "^(?!/)(?![A-Za-z]:)(?!.*\\\\)(?!.*//)(?!.*(?:^|/)\\.\\.?(/|$)).+$"
```
ayn-builds: Three of the five lookaheads are anchored at character 0, so a single leading space defeats all of them. `(?!.*\\)` and `(?!.*//)` use `.*` and hold at any position, but `(?!/)`, `(?![A-Za-z]:)` and `(?!.*(?:^|/)\.\.?(/|$))` do not.

Probed against the compiled pattern:

```
accepts  " /etc/passwd"     rejects  "/etc/passwd"
accepts  "\t/etc/passwd"    rejects  ".."
accepts  " .."              rejects  "../x"
accepts  " ./x"             rejects  "src/../secret"
accepts  " C:/x"            rejects  "C:/secret"
accepts  "a/ ../b"          rejects  "src//secret"
accepts  "~/.ssh/id_rsa"    <- needs no prefix at all
```

`~` is not covered by any lookahead, so home-relative paths pass unprefixed.

The negative cases at `tests/tools/application-source-contract.test.ts:515-524` are all unprefixed, which is exactly why this passes: every one of them has a whitespace-prefixed twin that the suite accepts.

Blast radius is 7 fields, since `sourcePath` is `$ref`d by `source.path`, `runtimeFrameworkRecord.root`, `buildMethodRecord.authority_path`, `buildMethodRecord.context_path`, `processCommandRecord.entrypoint`, `localFileRecord.default_path`, and `observabilityRecord.file_path`.

Suggest anchoring the whole thing rather than adding cases: a leading `(?!\s)` plus trailing `(?<!\s)`, or reformulating as a positive character-class grammar for a relative POSIX path. Worth adding a whitespace-prefixed variant of each existing negative to the suite so this class stays closed.

Applies identically to the `advisor/plugins/aws-startup-advisor/...` mirror at the same line.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/shared/application-source-contract.md
```diff
@@ -0,0 +1,47 @@
...
+application-source reviewer. It is data only: no current phase or target reads it.
+
+Requests expose only selected question names, application identity, process/configuration names,
+attachment and Private Space presence, add-on IDs, and selected estate application IDs.
+Configuration values, credentials, connection strings, and source excerpts have no fields.
+
+Every requested question has exactly one finding. `PRESENT` carries a non-empty typed record array;
+`ABSENT_WITHIN_REVIEWED_SCOPE`, `UNKNOWN`, and `NOT_APPLICABLE` carry `null`. `UNKNOWN` requires a
+limitation. Absence is invalid if a `SKIPPED_SOURCE`, `UNREADABLE_SOURCE`, `TRUNCATED_SOURCE`, or
+`DYNAMIC_SOURCE` limitation could affect it. Sources are optional direct relative paths with
```
ayn-builds: The three invariants stated here are the load-bearing ones, and none of them is expressible in the schema as written:

- "Every requested question has exactly one finding" - `findings` (`schema:767`) is `minItems: 1, maxItems: 22` with no `uniqueItems` on `question` and no relation to `requested_questions`. Twenty-two findings for the same question validate.
- "`UNKNOWN` requires a limitation" - `limitations` (`schema:129`) has `maxItems: 16` and no `minItems`, so `limitations: []` on an `UNKNOWN` finding validates.
- "Absence is invalid if a `SKIPPED_SOURCE` ... limitation could affect it" - not expressible in draft-07 at all.

The `allOf` at `schema:131` has exactly one branch, covering only PRESENT->non-null / else->null.

All three are enforced only by `validateSemantics` in the test file, which is not shipped. I understand from README.md ("Future ...") and lines 46-47 here that wiring a reviewer in is deliberately deferred, so this is not a defect today. The risk is at the wiring-in step: someone who validates against the schema will reasonably believe these hold. The two that draft-07 *can* express are cheap to close now (`minItems: 1` under an `if status == UNKNOWN`, and `uniqueItems` plus a question-count check). For the third, please state next to the invariant that it is a reviewer obligation with no schema enforcement, so the gap is recorded rather than inferred.

Applies identically to the `advisor/...` mirror.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/shared/application-source-contract.md
```diff
@@ -0,0 +1,47 @@
+# Application-Source Contract
+
+`application-source-contract.schema.json` defines a request and findings document for a future
+application-source reviewer. It is data only: no current phase or target reads it.
+
+Requests expose only selected question names, application identity, process/configuration names,
+attachment and Private Space presence, add-on IDs, and selected estate application IDs.
+Configuration values, credentials, connection strings, and source excerpts have no fields.
```
ayn-builds: > Configuration values, credentials, connection strings, and source excerpts have no fields.

This is the strongest guarantee in the contract, and it rests entirely on convention. Eleven fields are `$ref text` - a bare 1-500 char string with no `pattern`:

`limitation.detail`, `processCommandRecord.command`, `deploymentCommandRecord.command`, `deploymentCommandRecord.purpose`, `runtimeSettingRecord.use`, `herokuBehaviorRecord.use`, `herokuBehaviorRecord.effect`, `recurringJobRecord.entrypoint`, `recurringJobRecord.coordination`, `localFileRecord.purpose`, `externalServiceRecord.role`

Two concrete leaks:

- `limitation.detail` is where a reviewer explains a `TRUNCATED_SOURCE` or `UNREADABLE_SOURCE`, which is the single most natural place to paste the excerpt that caused it. That contradicts the "source excerpts" clause directly.
- `command` fields take a verbatim command line. `heroku config:set SIGNING_SECRET=...` and `psql postgres://user:pass@host/db` are both legal `text` and both carry exactly what this sentence forbids.

The suite can't catch this, and I want to be precise about why: the "value-bearing" rejection at `test:482-486` only adds *undeclared property names* (`sample.secret`, `context.configuration_values`, `context.credentials`), which `additionalProperties: false` rejects on the name alone. No test asserts that a *declared* field rejects a secret-shaped value, because nothing in the schema lets it.

Not asking for a credential-detecting regex. But given this is the contract's headline property, it needs either a reviewer-side obligation stated as such, or narrower types on the highest-risk fields (`command` and `limitation.detail`), rather than a guarantee that reads as enforced.

Applies identically to the `advisor/...` mirror.

migrate/plugins/migration-to-aws/tests/tools/application-source-contract.test.ts
```diff
@@ -0,0 +1,580 @@
...
+        typeof source.line_start === 'number'
+        && typeof source.line_end === 'number'
+        && source.line_end < source.line_start
+      ) errors.push(`${String(finding.question)}: source line bounds are reversed`);
+    }
+  }
+
+  const records = recordsByQuestion(answer);
+  const ids = (question: string, key: string) => (records.get(question) ?? []).map((item) => item[key] as string);
+  const componentIds = ids('runtime_framework', 'component_id');
```
ayn-builds, line 398: These id sets are built from specific questions' records - `component_id` from `runtime_framework`, `process_id` from `process_commands`, `listener_id` from `network_listeners`, and so on - and then every record's reference is checked against them at :407. That assumes all id-defining questions are present.

Partial requests are explicitly legitimate: `requested_questions` is `minItems: 1, maxItems: 22, uniqueItems: true`, and the request/answer split only makes sense if a reviewer can be asked a subset.

So a request of `['process_commands']` alone produces `process_commands: broken component_id` for every record, even though the answer is perfectly correct - `runtime_framework` was never asked, so no component ids were ever declared. Same for `['network_listeners']`, `['webhooks']`, `['application_connections']`, and any other subset that references an id it does not itself define.

The suite doesn't surface this because :534 is the only single-question request and it picks `runtime_framework`, which happens to be the component-id source.

The fix is to scope the check: only flag a broken reference when the defining question was actually requested. Worth a test with a subset that references without defining.

migrate/plugins/migration-to-aws/tests/tools/application-source-contract.test.ts
```diff
@@ -0,0 +1,580 @@
...
+  const processIds = ids('process_commands', 'process_id');
+  const listenerIds = ids('network_listeners', 'listener_id');
+  const dependencyIds = ids('external_services', 'dependency_id');
+  const relationshipIds = ids('application_connections', 'relationship_id');
+  const components = new Set(componentIds);
+  const processes = new Set(processIds);
+  const listeners = new Set(listenerIds);
+  const dependencies = new Set(dependencyIds);
+  const estateApps = new Set(object(reviewRequest.context).selected_estate_application_ids as string[]);
```
ayn-builds, line 407: `selected_estate_application_ids` and `addon_ids` get cross-checked here, but the other two vocabularies the context carries do not. `context.configuration_names` and `context.process_types` are both required (`schema` `context.required`), and they exist precisely to bound what a `setting_name` or a process `type` may say - yet nothing compares against them.

The fixtures demonstrate it. `configuration_names` at :150 is `['API_URL', 'DATABASE_URL', 'PORT', 'REDIS_URL', 'SIGNING_SECRET']`, but the records use four names that aren't in it:

- `HOST` (:192, `port_host_binding.host_setting_name`)
- `LOG_LEVEL` (:279, `logs_telemetry.setting_name`)
- `UPLOAD_DIR` (:247, `local_file_writes.setting_name`)
- `WORKER_URL` (:314, `application_connections.setting_name`)

These pass today. Since `application-source-contract.md:19` makes this the whole point of the naming discipline - "setting names never carry values", so the name is the only checkable thing - an unrecognized setting name is a silent, uncatchable error in the one field the design says to trust. Same argument for `process_commands.type` against `process_types`.

Suggest extending `validateSemantics` to both vocabularies, and correcting the fixtures so they satisfy their own declared context.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/shared/application-source-contract.schema.json
```diff
@@ -0,0 +1,771 @@
...
+      "type": "object",
+      "required": ["job_id", "component_id", "process_ids", "name", "mechanism", "entrypoint", "coordination"],
+      "additionalProperties": false,
+      "properties": {
+        "job_id": { "$ref": "#/definitions/identifier" },
+        "component_id": { "$ref": "#/definitions/identifier" },
+        "process_ids": { "$ref": "#/definitions/identifierList" },
+        "name": { "$ref": "#/definitions/name" },
+        "mechanism": { "$ref": "#/definitions/name" },
+        "entrypoint": { "$ref": "#/definitions/text" },
```
ayn-builds: `entrypoint` means two different things in two records, with different validation:

- `processCommandRecord.entrypoint` (:184) is `$ref sourcePath` - path-validated.
- `recurringJobRecord.entrypoint` (here) is `$ref text` - a free 500-char string, no path checks at all.

The fixtures show the intent is real rather than accidental: `test:170` uses `entrypoint: 'server.js'` (a file) while `test:229` uses `entrypoint: 'npm run cleanup'` (a command). But `application-source-contract.md` lists `entrypoint` in both the `process_commands` row (:26) and the `recurring_jobs` row (:33) with no per-field definitions anywhere, so nothing tells a reviewer which sense applies.

Two consequences. A reviewer who reads across from `process_commands` and puts a path in `recurring_jobs.entrypoint` gets zero path validation, including `/etc/passwd` and `../../x`. A reviewer who goes the other way gets an unexplained pattern failure.

Suggest distinct names for distinct concepts - `entrypoint` for the path sense and something like `invocation` or `command` for the command sense, matching `processCommandRecord.command`. If the shared name is intentional, the doc needs to say which is which.

Applies identically to the `advisor/...` mirror.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/shared/application-source-contract.schema.json
```diff
@@ -0,0 +1,771 @@
...
+      }
+    },
+    "source": {
+      "type": "object",
+      "required": ["path"],
+      "additionalProperties": false,
+      "properties": {
+        "path": { "$ref": "#/definitions/sourcePath" },
+        "line_start": { "type": "integer", "minimum": 1, "maximum": 10000000 },
+        "line_end": { "type": "integer", "minimum": 1, "maximum": 10000000 }
```
ayn-builds, line 108: Nothing relates `line_end` to `line_start`, so `{ "path": "package.json", "line_start": 8, "line_end": 1 }` validates. Draft-07 can't compare two fields, so this needs to go in `validateSemantics` alongside the other cross-field checks. Minor, but citations are the auditable part of a finding, and an inverted range makes one unverifiable.

Applies identically to the `advisor/...` mirror.

amjadsy:
@ayn-builds 

I've addressed five of the findings you brought up: stricter source-path validation, schema-required explanations for UNKNOWN, clearer secret-redaction requirements, correct validation of partial reviews, and renaming recurring-job entrypoint to command.               

I did not restrict discovered process or configuration names to the Heroku inventory because differences can identify missing configuration or drift. Reversed line ranges are already rejected by the executable validator and covered by a      focused test.

migrate/plugins/migration-to-aws/tests/tools/application-source-contract.test.ts
```diff
@@ -0,0 +1,643 @@
...
+  const addons = new Set(object(reviewRequest.context).addon_ids as string[]);
+
+  for (const [question, questionRecords] of records) {
+    for (const record of questionRecords) {
+      for (const key of ['component_id', 'caller_component_id']) {
+        if (
+          requestedSet.has('runtime_framework')
```
ayn-builds: **Blocking.** `application-source-contract.md:16-17` says to check cross-record references "when their defining questions are present", but this gates on whether the question was *requested*. Those two come apart whenever a defining question is requested and comes back non-PRESENT, and the validator then rejects an answer the contract allows.

Request `['runtime_framework', 'process_commands']`, answer with `runtime_framework` as UNKNOWN plus a limitation and `process_commands` PRESENT. `validate` returns `[]`, `validateSemantics` returns `["process_commands: broken component_id"]`. Identical for `ABSENT_WITHIN_REVIEWED_SCOPE` and `NOT_APPLICABLE`, since none of them populate `components`. The other two gates behave the same way: 418 gives `network_listeners: broken process_id` when `process_commands` is UNKNOWN, and 425 gives `health_routes: broken listener_id` when `network_listeners` is UNKNOWN.

The test at 578 covers the defining question not being *requested*, and that passes, which is why the suite stays green. Requested-but-unanswered isn't covered anywhere. Building the sets from findings whose status is PRESENT, rather than from `requestedSet`, would match the prose, and the prose has it right: a reviewer that can't pin down the runtime shouldn't invalidate every process record that names a component.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/shared/application-source-contract.schema.json
```diff
@@ -0,0 +1,775 @@
...
+        "postgres_attachment_present",
+        "redis_attachment_present",
+        "addon_ids",
+        "private_space_present",
+        "selected_estate_application_ids"
+      ],
+      "additionalProperties": false,
+      "properties": {
+        "process_types": { "$ref": "#/definitions/stringList" },
+        "configuration_names": { "$ref": "#/definitions/stringList" },
```
ayn-builds: **Blocking.** `configuration_names` reuses `stringList`, which caps at 32 on line 44. These names are inventory context rather than a curated subset, so an app with more config vars than that can't produce a valid request at all. 32 names validate clean, 33 fail. Plenty of Heroku apps sit above that line.

`records` on line 143 has the same problem with `maxItems: 64`. `runtime_settings` is one record per component and setting, so two components with 40 settings each is already over. 64 records pass, 65 fail.

`process_types` sharing the 32 cap looks fine, it's the config-var reuse that's too tight. A separate list definition with a higher bound would sort it, and 64 records per question deserves the same second look.

Both plugin copies are byte-identical, so whatever bounds you pick need the same edit under `advisor/plugins/aws-startup-advisor/`.

migrate/plugins/migration-to-aws/tests/tools/application-source-contract.test.ts
```diff
@@ -0,0 +1,643 @@
...
+  const findingNames = rawFindings.map((raw) => object(raw).question as string);
+  for (const question of requested) {
+    if (findingNames.filter((name) => name === question).length !== 1) errors.push(`expected one finding for ${question}`);
+  }
+  for (const question of findingNames) {
+    if (!requested.includes(question)) errors.push(`unrequested finding ${question}`);
+  }
+  for (const raw of rawFindings) {
+    const finding = object(raw);
+    const limitations = finding.limitations as JsonObject[];
```
ayn-builds: This reads `.length` off `finding.limitations` without checking it's there, and 378 then calls `.some` on it. A finding with status UNKNOWN and no `limitations` key throws `TypeError: Cannot read properties of undefined (reading 'length')` here; `ABSENT_WITHIN_REVIEWED_SCOPE` throws the same way on `.some` at 378. PRESENT is unaffected because neither branch runs.

`findingBase` requires `limitations`, so schema-valid input never hits this. But the tests at 493-497 call `validateSemantics` on input they've just asserted is schema-invalid, so the helper is already being used in a mode where the field can be missing, and crashing there instead of reporting an error is easy to trip over later. `(finding.limitations ?? [])` covers it.

amjadsy:
@ayn-builds Thanks for the review. All three findings are addressed in 049fb00, with regression tests added for each case. Local build and security checks pass; GitHub CI is running.

ayn-builds (approved):

PR #246, opened 2026-08-25
fix(agent-advisor): bedrock.openai_gpt56_region_availability — schema change

leon1418:

## What changed upstream

[OpenAI GPT-5.6 Terra and Luna now available on Amazon Bedrock in AWS GovCloud (US)](https://aws.amazon.com/about-aws/whats-new/2026/08/openai-gpt-terra-luna-govcloud/)

**Verdict: `schema_change`** on `bedrock.openai_gpt56_region_availability`

- was — Region availability: us-east-1, us-east-2, us-west-2 at minimum (commercial regions only)
- now — Region availability now spans two partitions: commercial regions (us-east-1, us-east-2, us-west-2, check pricing page) AND AWS GovCloud (US-West, US-East) for Terra and Luna specifically — Sol's GovCloud availability is not stated in this announcement.

> **Still true:** Commercial-region availability (us-east-1, us-east-2, us-west-2) for GPT-5.6 Sol/Terra/Luna remains correct as stated.
>

## Proposed edits

### `agent-advisor/references/models/openai-bedrock-2026-07-21.json:136` — value

```diff
- "openai_gpt_5_6_luna": {
+ "openai_gpt_5_6_luna": { // region_availability should be updated to include AWS GovCloud (US-West, US-East) in addition to commercial regions (us-east-1, us-east-2, us-west-2)
```

**Why:** The announcement states Luna is now generally available in AWS GovCloud (US-West) and (US-East) in addition to commercial regions, so the region_availability field for this model entry must be updated to reflect the new GovCloud partitions.

**Evidence (verbatim from the announcement):** "GPT-5.6 Terra and Luna are now generally available on Amazon Bedrock in AWS GovCloud (US-West) and AWS GovCloud (US-East)"

### `gcp-to-aws/references/design-refs/ai-openai-to-bedrock.md:133` — value

```diff
- - **Region availability.** Confirm the target region serves GPT-5.6 on Mantle (us-east-1, us-east-2, us-west-2 at minimum; check the pricing page for current list).
+ - **Region availability.** Confirm the target region serves GPT-5.6 on Mantle (us-east-1, us-east-2, us-west-2 at minimum for commercial regions; Terra and Luna are now also available in AWS GovCloud (US-West, US-East); check the pricing page for current list).
```

**Why:** Announcement states GPT-5.6 Terra and Luna are now generally available on Amazon Bedrock in AWS GovCloud (US-West) and AWS GovCloud (US-East), adding a new partition not reflected in this region-availability guidance.

**Evidence (verbatim from the announcement):** "GPT-5.6 Terra and Luna are now generally available on Amazon Bedrock in AWS GovCloud (US-West) and AWS GovCloud (US-East)"

### `gcp-to-aws/references/shared/pricing-cache.md:539` — value

```diff
- | GPT-5.6 Luna  | US East (N. Virginia / Ohio), US West (Oregon) | 0.22               | 1.32                | 0.44               | 1.98                |
+ | GPT-5.6 Luna  | US East (N. Virginia / Ohio), US West (Oregon), AWS GovCloud (US-West), AWS GovCloud (US-East) | 0.22               | 1.32                | 0.44               | 1.98                |
```

**Why:** Announcement states Luna is now GA on Bedrock in AWS GovCloud (US-West) and (US-East) in addition to commercial regions.

**Evidence (verbatim from the announcement):** "GPT-5.6 Terra and Luna are now generally available on Amazon Bedrock in AWS GovCloud (US-West) and AWS GovCloud (US-East)"

### `gcp-to-aws/references/shared/pricing-cache.md:538` — value

```diff
- | GPT-5.6 Terra | US East (N. Virginia / Ohio), US West (Oregon) | 2.20               | 13.20               | 4.40               | 19.80               |
+ | GPT-5.6 Terra | US East (N. Virginia / Ohio), US West (Oregon), AWS GovCloud (US-West), AWS GovCloud (US-East) | 2.20               | 13.20               | 4.40               | 19.80               |
```

**Why:** Announcement states Terra is now GA on Bedrock in AWS GovCloud (US-West) and (US-East) in addition to commercial regions.

**Evidence (verbatim from the announcement):** "GPT-5.6 Terra and Luna are now generally available on Amazon Bedrock in AWS GovCloud (US-West) and AWS GovCloud (US-East)"

## Filter false positives

- `agent-advisor/references/decision-refs/model-selection.md`

The same edits were applied to this second copy: 4 applied.

### DIFF
--- a/advisor/plugins/aws-startup-advisor/skills/agent-advisor/references/models/openai-bedrock-2026-07-21.json
+++ b/advisor/plugins/aws-startup-advisor/skills/agent-advisor/references/models/openai-bedrock-2026-07-21.json
@@ -68,6 +68,11 @@
       "class": "reasoning",
       "context_window": 272000,
       "output_token_ceiling": "unknown",
+      "region_availability": {
+        "commercial": ["us-east-1", "us-east-2"],
+        "govcloud": [],
+        "evidence": "Commercial set per this catalog's pricing cache (US East N. Virginia / Ohio). The 2026-08 GovCloud announcement names Terra and Luna only \u2014 Sol's GovCloud availability is not stated; treat as unknown, not absent."
+      },
       "capabilities": [
         "reasoning",
         "tool_or_function_calling",
@@ -119,6 +124,11 @@
       "class": "reasoning",
       "context_window": "unknown",
       "output_token_ceiling": "unknown",
+      "region_availability": {
+        "commercial": ["us-east-1", "us-east-2", "us-west-2"],
+        "govcloud": ["us-gov-west-1", "us-gov-east-1"],
+        "evidence": "AWS What's New 2026-08 (openai-gpt-terra-luna-govcloud): \"GPT-5.6 Terra and Luna are now generally available on Amazon Bedrock in AWS GovCloud (US-West) and AWS GovCloud (US-East)\"; commercial set per this catalog's pricing cache \u2014 verify against the pricing page before relying on it."
+      },
       "capabilities": [
         "reasoning",
         "tool_or_function_calling"
@@ -133,7 +143,7 @@
         }
       }
     },
-    "openai_gpt_5_6_luna": { // region_availability should be updated to include AWS GovCloud (US-West, US-East) in addition to commercial regions (us-east-1, us-east-2, us-west-2)
+    "openai_gpt_5_6_luna": {
       "display_name": "OpenAI GPT-5.6 Luna (on Bedrock)",
       "family": "openai_gpt_5",
       "generation": "reasoning",
@@ -141,6 +151,11 @@
       "class": "reasoning",
       "context_window": "unknown",
       "output_token_ceiling": "unknown",
+      "region_availability": {
+        "commercial": ["us-east-1", "us-east-2", "us-west-2"],
+        "govcloud": ["us-gov-west-1", "us-gov-east-1"],
+        "evidence": "AWS What's New 2026-08 (openai-gpt-terra-luna-govcloud): \"GPT-5.6 Terra and Luna are now generally available on Amazon Bedrock in AWS GovCloud (US-West) and AWS GovCloud (US-East)\"; commercial set per this catalog's pricing cache \u2014 verify against the pricing page before relying on it."
+      },
       "capabilities": [
         "reasoning",
         "tool_or_function_calling"
--- a/advisor/plugins/aws-startup-advisor/skills/agent-advisor/references/models/openai-bedrock-2026-07-21.json
+++ b/advisor/plugins/aws-startup-advisor/skills/agent-advisor/references/models/openai-bedrock-2026-07-21.json
@@ -133,7 +133,7 @@
         }
       }
     },
-    "openai_gpt_5_6_luna": {
+    "openai_gpt_5_6_luna": { // region_availability should be updated to include AWS GovCloud (US-West, US-East) in addition to commercial regions (us-east-1, us-east-2, us-west-2)
       "display_name": "OpenAI GPT-5.6 Luna (on Bedrock)",
       "family": "openai_gpt_5",
       "generation": "reasoning",
--- a/advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/references/design-refs/ai-openai-to-bedrock.md
+++ b/advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/references/design-refs/ai-openai-to-bedrock.md
@@ -130,7 +130,7 @@ Long-context (>272K, up to 1M) rates are 2× input / 1.5× output — see `prici
 - **Mantle-only.** GPT-5.6 models are served exclusively on the `bedrock-mantle` endpoint via the OpenAI **Responses API** path (`openai/v1/responses`). No `bedrock-runtime`, no Converse API, no Invoke ([model card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-openai-gpt-56-terra.html)). Applications on the OpenAI SDK migrate with an env-var + model-string swap; applications you plan to rewrite onto boto3 `converse()` cannot target GPT-5.6.
 - **No geo/global inference profiles.** Model IDs are bare (`openai.gpt-5.6-terra`); geo-prefixed (`us.`) and global inference IDs are not supported.
 - **No Bedrock Guardrails / Knowledge Bases integration on the Mantle path.** If those are requirements, use a Converse-capable family (Claude, Nova) instead.
-- **Region availability.** Confirm the target region serves GPT-5.6 on Mantle (us-east-1, us-east-2, us-west-2 at minimum; check the pricing page for current list).
+- **Region availability.** Confirm the target region serves GPT-5.6 on Mantle (us-east-1, us-east-2, us-west-2 at minimum for commercial regions; Terra and Luna are now also available in AWS GovCloud (US-West, US-East); check the pricing page for current list).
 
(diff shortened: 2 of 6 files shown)


herosjourney (changes requested):
## Blocking — invalid JSON introduced in `openai-bedrock-2026-07-21.json` (both copies)

The edit to `openai_gpt_5_6_luna` inserts a `//` comment as part of a JSON object key:

```diff
-    "openai_gpt_5_6_luna": {
+    "openai_gpt_5_6_luna": { // region_availability should be updated to include AWS GovCloud (US-West, US-East) in addition to commercial regions (us-east-1, us-east-2, us-west-2)
```

JSON has no comment syntax. I pulled this branch and confirmed both copies fail to parse:

```
$ python3 -c "import json; json.load(open('openai-bedrock-2026-07-21.json'))"
json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 136 column 30
```

This file is a data file `agent-advisor` reads at runtime for model selection — this isn't a formatting nit, it breaks parsing for the whole file, not just this entry.

## Also blocking — the described fix was never actually made

The comment says `region_availability should be updated`, but no such field exists anywhere in this file (`grep -n region_availability` returns nothing). So beyond being invalid JSON, this "fix" is a TODO note in place of the actual edit the PR claims to have made. The `openai_gpt_5_6_luna` entry still has no region-availability data reflecting GovCloud at all.

## Note on the follow-up commit

`64652b2` ("dprint fmt for the GPT-5.6 pricing table column widths") only re-pads the markdown table in `pricing-cache.md`. It doesn't touch the JSON file, so the syntax error above is still present at HEAD. Worth separately flagging that `mise run build` apparently didn't catch this — a JSON-validation step (even a plain `json.load` per changed `.json` file) in CI would have caught this before it reached review.

## What's correct

The two markdown edits are accurate and match the source announcement:
- `ai-openai-to-bedrock.md` region-availability note — correctly adds GovCloud as an addition to (not a replacement of) the existing commercial-region list.
- `pricing-cache.md` — correctly updates the region cells for Terra and Luna only, and correctly leaves Sol's row untouched (the announcement doesn't mention Sol's GovCloud availability, and the PR body explicitly calls this out — good judgment call not to extrapolate).

## Suggested fix

Replace the comment-as-key with an actual `region_availability` field (or whatever the real schema key should be) populated with the commercial + GovCloud values, matching the pattern already used correctly in the two markdown files. Re-run the JSON parse check on both copies before merging.

herosjourney (changes requested):
## Blocking — invalid JSON introduced in `openai-bedrock-2026-07-21.json` (both copies)

The edit to `openai_gpt_5_6_luna` inserts a `//` comment as part of a JSON object key:

```diff
-    "openai_gpt_5_6_luna": {
+    "openai_gpt_5_6_luna": { // region_availability should be updated to include AWS GovCloud (US-West, US-East) in addition to commercial regions (us-east-1, us-east-2, us-west-2)
```

JSON has no comment syntax. I pulled this branch and confirmed both copies fail to parse:

```
$ python3 -c "import json; json.load(open('openai-bedrock-2026-07-21.json'))"
json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 136 column 30
```

This file is a data file `agent-advisor` reads at runtime for model selection — this isn't a formatting nit, it breaks parsing for the whole file, not just this entry.

## Also blocking — the described fix was never actually made

The comment says `region_availability should be updated`, but no such field exists anywhere in this file (`grep -n region_availability` returns nothing). So beyond being invalid JSON, this "fix" is a TODO note in place of the actual edit the PR claims to have made. The `openai_gpt_5_6_luna` entry still has no region-availability data reflecting GovCloud at all.

## Note on the follow-up commit

`64652b2` ("dprint fmt for the GPT-5.6 pricing table column widths") only re-pads the markdown table in `pricing-cache.md`. It doesn't touch the JSON file, so the syntax error above is still present at HEAD. Worth separately flagging that `mise run build` apparently didn't catch this — a JSON-validation step (even a plain `json.load` per changed `.json` file) in CI would have caught this before it reached review.

## What's correct

The two markdown edits are accurate and match the source announcement:
- `ai-openai-to-bedrock.md` region-availability note — correctly adds GovCloud as an addition to (not a replacement of) the existing commercial-region list.
- `pricing-cache.md` — correctly updates the region cells for Terra and Luna only, and correctly leaves Sol's row untouched (the announcement doesn't mention Sol's GovCloud availability, and the PR body explicitly calls this out — good judgment call not to extrapolate).

## Suggested fix

Replace the comment-as-key with an actual `region_availability` field (or whatever the real schema key should be) populated with the commercial + GovCloud values, matching the pattern already used correctly in the two markdown files. Re-run the JSON parse check on both copies before merging.

leon1418:
Both blocking findings fixed in [43caec5e](https://github.com/awslabs/startups/pull/246/commits/43caec5e):

**Invalid JSON** — the `//` comment is gone; both copies parse (`json.load` verified on each).

**The fix that was never made** — `region_availability` now exists as a real field on all three GPT-5.6 entries, evidence-attributed in the catalog's house style:
- **terra / luna**: commercial `[us-east-1, us-east-2, us-west-2]` + govcloud `[us-gov-west-1, us-gov-east-1]`, citing the What's New verbatim.
- **sol**: commercial `[us-east-1, us-east-2]` only — narrower than the PR body's blanket claim, per this catalog's own pricing cache — and an empty govcloud array with an explicit "not stated in the announcement; treat as unknown, not absent" note, extending the no-extrapolation judgment you called out as correct.

Verified: drift 259 identical, model-id lint OK, dprint clean, agent-advisor suite 318/318 (includes the model-recommend consumers of this catalog).

**Root cause, both failure modes, fixed in the pipeline** (feat/kb-autoupdate-pipeline, the workflow's code checkout): `apply.py` now (1) runs `dprint fmt` over the work tree before staging, and (2) json-parses every changed `.json` and fails the run loudly on a decode error — a knowledge edit that corrupts a data file never opens a PR again. Your suggestion of a repo-side CI JSON-validation step is a good belt-and-suspenders on top; happy to add a `mise` task in a follow-up.

herosjourney (approved):
## Both blocking issues resolved — approving

Re-checked after commit `43caec5e`. Verified directly:

- **JSON validity:** both copies of `openai-bedrock-2026-07-21.json` now parse cleanly (`json.load` succeeds on both `advisor/` and `migrate/`). The malformed `// comment`-as-key on `openai_gpt_5_6_luna` is gone.
- **`region_availability` is now a real, populated field** — not a TODO comment — on Terra, Luna, *and* Sol. Diffed both copies; they're byte-identical, so the mirror is consistent.
- Good call adding it to Sol too, with `govcloud: []` plus an explicit evidence note that Sol's GovCloud status is "not stated" in the announcement rather than assuming absence — that's the right way to handle a gap in the source material rather than silently guessing.

Markdown edits (`ai-openai-to-bedrock.md`, `pricing-cache.md`) are unchanged since my last pass and were already accurate against the source announcement.

No remaining concerns from my end.

herosjourney:
Thanks for tracing this back to the pipeline — that's the better fix. A per-PR patch would have left the same failure mode (malformed JSON slipping through as a "fix") free to recur on the next auto-update run; validating at `apply.py` time closes it for every future PR this pipeline opens, not just this one.

Agreed a repo-side CI JSON-validation step is worth adding regardless — belt-and-suspenders is the right call for a file that gets edited by both a bot and humans. A `mise` task alongside the existing `lint:*`/`drift:check` tasks would keep it consistent with how this repo already gates the rest of `references/`.

My approval stands — nothing further from me on this PR.

az-zhu (approved):

PR #247, opened 2026-08-25
feat(gcp-to-aws): add EKS Auto Mode guidance for GKE migrations

hasrazz:
## Problem

For GCP → AWS migrations, the `gcp-to-aws` skill defaulted GKE workloads to **ECS Fargate** whenever the user didn't explicitly express a Kubernetes preference, and it did not offer **EKS Auto Mode** as a target at all. This is out of step with current AWS guidance, which positions EKS Auto Mode as the recommended low-operational-overhead way to run Kubernetes (the closest equivalent to GKE Autopilot). GKE Autopilot clusters in particular had no matching AWS landing zone, and the Autopilot signal — although captured in live discovery — never reached the Clarify/Design phases.

## Solution

- **Discovery:** detect GKE Autopilot on both paths (Terraform `enable_autopilot`, live `autopilot.enabled`) and normalize a canonical `config.autopilot_enabled` key into the inventory schema so Clarify/Design can read it.
- **Clarify (Q8):** default GKE sources to **EKS Auto Mode**; standard EKS managed node groups and ECS Fargate become explicit opt-outs; the Autopilot flag is surfaced as context (Autopilot → Auto Mode is the 1:1 match).
- **Multi-cloud (Q5):** routes to **standard EKS** (Auto Mode is excluded — its node management is AWS-specific and defeats portability).
- **Threading:** the new `eks-auto` / `eks-standard` / `ecs-fargate` enum flows through the compute rubric (`compute.md`), fast-path, index, Terraform generation (Auto Mode cluster vs managed node groups), pricing formulas (Auto Mode management fee + control-plane fee), preferences schema, billing path, and workshop sheet.
- **Cross-plugin parity:** the same changes are mirrored into the `advisor/` copy of `gcp-to-aws` so the cross-plugin drift check stays green.

## Type of Change

- [ ] New plugin/power/tool
- [ ] Bug fix
- [ ] Documentation update
- [ ] Guardrail/CI update

## Team Folder

- [x] `advisor/`
- [x] `migrate/`
- [ ] Other: ___

> Note: the change lives in `migrate/plugins/migration-to-aws/skills/gcp-to-aws/` and is mirrored into `advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/`, which the repo's `cross-plugin-drift` check requires to stay in sync.

## Checklist

> Verification: `mise run build` passes locally (exit 0) — lint, `fmt:check`, and the security suite (bandit, semgrep, gitleaks, checkov 184/0, grype clean) all pass; `cross-plugin-drift`, fixtures, frontmatter, and tests are OK; the working tree is not mutated by the build. Changes are documentation/prompt Markdown only. (Pre-existing, unrelated `pricing:staleness` warnings on the heroku-to-aws caches are warn-only and untouched by this PR.)


herosjourney (commented):
## Summary

Good direction overall — defaulting GKE migrations to EKS Auto Mode is the right call technically (it's the correct GKE Autopilot analog, and it's AWS's own recommended low-ops path), and the change is threaded consistently through Clarify, Design, Estimate, Generate, pricing, and schema. The `advisor/` mirror is verified byte-identical to the `migrate/` source apart from path prefixes, so the drift-check claim in the PR body holds up.

One real bug should block merge. Two other items are worth addressing but not blocking.

---

## Blocking

### `graviton.md` / `generate-artifacts-infra.md` — arm64 guidance for EKS Auto Mode is incomplete and would silently fail

Both files say to select arm64 for Auto Mode by setting `kubernetes.io/arch: arm64` on "the NodePool / `aws_eks_node_class`." I checked this against AWS's docs: the built-in `general-purpose` NodePool — the one that actually schedules application workloads — is **amd64-only**. Only the built-in `system` NodePool (tainted `CriticalAddonsOnly`, for cluster components) supports both architectures. ([Enable or Disable Built-in NodePools](https://docs.aws.amazon.com/eks/latest/userguide/set-builtin-node-pools.html))

Getting arm64 app workloads on Auto Mode requires a **custom NodePool + NodeClass** (Karpenter CRDs applied via `kubectl`/Helm) — it's not a flag you can flip on the default pool, and it's not something the `aws_eks_cluster` Terraform resource can express. As written, an agent following this instruction will either fail or silently produce an amd64 cluster while reporting Graviton was applied.

Suggested fix: either (a) note that Auto Mode + Graviton requires generating an additional custom NodePool/NodeClass manifest (not just Terraform) and do that, or (b) flag to the user that Graviton isn't cleanly supported for Auto Mode via this skill's generated output yet, rather than asserting it works.

---

## Non-blocking

### `compute.md` — no regional-availability check for Auto Mode

Elsewhere in this codebase (`design-ai.md`, Bedrock model selection), newer AWS capabilities go through the `awsknowledge` MCP regional-availability check before being recommended. This PR recommends Auto Mode unconditionally on GKE detection with no equivalent gate. Auto Mode is broadly available today, but it's inconsistent with the caution pattern this repo already established for exactly this kind of rollout. Suggest adding the same MCP check, or explicitly noting why it's not needed here.

### `clarify-compute.md` Q8 — question wording states the recommendation before presenting the choice

The new Q8 opens with "You're on GKE today, so by default we keep you on Kubernetes with EKS Auto Mode... How would you like to proceed?" before listing options A/B/C. That's a change from the old Q8, which explicitly instructed framing the question neutrally ("frame it practically so the user gives an honest answer rather than aspirational"). Leading with the recommended answer nudges "I don't know" responses toward passive confirmation rather than an actual decision.

To be clear, I think the **default itself is a good change** — Q8 only fires when GKE is already in inventory, meaning the customer already chose Kubernetes deliberately; defaulting undecided users to *stay* on that abstraction (via the lowest-ops AWS variant) respects an existing investment better than the old Fargate default, which silently dropped it. I'd just ask to restructure the prompt so options are presented before the recommendation is stated, keeping the same default.

**Confirmed not an issue:** I traced Q8's answer through Design and Generate independently, and all three answers (Auto Mode / standard EKS / Fargate) produce distinct, correct Terraform output — this isn't a case of the preference being captured but not honored downstream. The eliminator logic also correctly overrides preference when a technical blocker exists (node-level customization needs, or the Q5 multi-cloud portability override), which is the right precedence — user preference should win among viable options, not over hard constraints.

---

## What's correctly implemented (verified against AWS docs)

- IMDSv2 enforcement claim for Auto Mode nodes (hop limit 1, non-configurable) — accurate, no launch template needed.
- Two distinct IAM roles (cluster role vs. node role, separate managed-policy sets) — matches AWS's documented setup exactly.
- Terraform block shape (`compute_config { enabled, node_pools, node_role_arn }`, `kubernetes_network_config { elastic_load_balancing { enabled } }`, `storage_config { block_storage { enabled } }`) — matches real working Auto Mode Terraform.
- Excluding Auto Mode from the Q5 multi-cloud path — correct; Auto Mode's node management genuinely is AWS-specific and would undercut the portability goal.
- Savings Plan caveat (applies to underlying EC2 instances, not the Auto Mode management fee or control-plane fee) — directionally correct per AWS's cost docs.

hasrazz:
> ## Blocking
> ### `graviton.md` / `generate-artifacts-infra.md` — arm64 guidance for EKS Auto Mode is incomplete and would silently fail
> Both files say to select arm64 for Auto Mode by setting `kubernetes.io/arch: arm64` on "the NodePool / `aws_eks_node_class`." I checked this against AWS's docs: the built-in `general-purpose` NodePool — the one that actually schedules application workloads — is **amd64-only**. Only the built-in `system` NodePool (tainted `CriticalAddonsOnly`, for cluster components) supports both architectures. ([Enable or Disable Built-in NodePools](https://docs.aws.amazon.com/eks/latest/userguide/set-builtin-node-pools.html))
> 
> Getting arm64 app workloads on Auto Mode requires a **custom NodePool + NodeClass** (Karpenter CRDs applied via `kubectl`/Helm) — it's not a flag you can flip on the default pool, and it's not something the `aws_eks_cluster` Terraform resource can express. As written, an agent following this instruction will either fail or silently produce an amd64 cluster while reporting Graviton was applied.
> 
> Suggested fix: either (a) note that Auto Mode + Graviton requires generating an additional custom NodePool/NodeClass manifest (not just Terraform) and do that, or (b) flag to the user that Graviton isn't cleanly supported for Auto Mode via this skill's generated output yet, rather than asserting it works.
> 

Yes - This is correct. We need separate manifests for graviton in EKS Auto mode as of now. I have added detailed guidance in newer commit. [dec4d29](https://github.com/awslabs/startups/pull/247/commits/dec4d290dc34858675c6afd0d17629f4d02fc7d3)



> ## Non-blocking
> ### `compute.md` — no regional-availability check for Auto Mode
> Elsewhere in this codebase (`design-ai.md`, Bedrock model selection), newer AWS capabilities go through the `awsknowledge` MCP regional-availability check before being recommended. This PR recommends Auto Mode unconditionally on GKE detection with no equivalent gate. Auto Mode is broadly available today, but it's inconsistent with the caution pattern this repo already established for exactly this kind of rollout. Suggest adding the same MCP check, or explicitly noting why it's not needed here.
> 

Nice, did not know about that, added this suggestion as well. [4aac819](https://github.com/awslabs/startups/pull/247/commits/4aac81904d20344650c3d83d5bf3a8b835b29786)



> ### `clarify-compute.md` Q8 — question wording states the recommendation before presenting the choice
> The new Q8 opens with "You're on GKE today, so by default we keep you on Kubernetes with EKS Auto Mode... How would you like to proceed?" before listing options A/B/C. That's a change from the old Q8, which explicitly instructed framing the question neutrally ("frame it practically so the user gives an honest answer rather than aspirational"). Leading with the recommended answer nudges "I don't know" responses toward passive confirmation rather than an actual decision.
> 

Restructured that in new commit.  [8c4f94d](https://github.com/awslabs/startups/pull/247/commits/8c4f94de68ac8fe4b7efed294e05ce0675b16cf2)



I will expand on ECS guidance in a different PR

herosjourney (approved):
## All three findings addressed — approving

Re-verified each fix directly against the diff, not just the commit message.

**Blocking - arm64/NodePool bug (dec4d290): fixed correctly.** graviton.md and generate-artifacts-infra.md now correctly state Graviton on Auto Mode is not a Terraform flag - the built-in general-purpose NodePool is amd64-only and immutable, and getting arm64 workloads running requires a custom Karpenter NodePool + EC2NodeClass applied as a Kubernetes manifest post-provision. The fix also adds the right safety net: if the run can't emit/apply that manifest, it falls back to amd64 and explicitly flags Graviton as not applied, rather than claiming arm64 succeeded. This closes the exact failure mode I flagged.

**Non-blocking - Q8 wording (8c4f94de): fixed, and improved beyond what I asked.** Options A/B/C are now presented before the recommendation, with the default noted last as "if you're unsure, we default to A" rather than opening the question. The default itself is correctly unchanged (still eks-auto), which was the right call to preserve. Also caught and removed a dead option ("D) N/A - we don't use Kubernetes") that couldn't fire since Q8 only triggers when GKE is already present - good catch I'd missed.

**Non-blocking - regional availability gate (4aac8190): fixed correctly.** Adds an awsknowledge MCP get_regional_availability check before finalizing eks-auto, reusing the same pattern this repo already applies to Bedrock/AgentCore. Falls back to eks-standard in the same region (not a different region) on unavailability, with both a rationale note and a warnings[] entry, plus a sane degraded path if the MCP call itself fails.

**Verified independently:**
- All touched files remain byte-identical between advisor/ and migrate/ mirrors.
- mise run build exits 0, drift:check reports 259 identical / 25 allowlisted (baseline, no new suppressions).

No remaining concerns. Approving.

advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/references/phases/design/design-billing.md
```diff
@@ -50,24 +50,24 @@ For each billing service, attempt lookup in order:
...
+| `google_compute_forwarding_rule` | Cloud Load Balancing | ALB                                    | NLB                                                                                                                                                                                                                                                                                                                                                                                                                 |
+| `google_compute_backend_service` | Cloud Load Balancing | ALB Target Groups                      | NLB                                                                                                                                                                                                                                                                                                                                                                                                                 |
```
leon1418: [🤖 AI review 🤖]

**Verified fixed at head ed268632 (September 14, 2026).** The App Engine billing table now specifies EKS standard node groups under compute: eks in both plugin copies, matching the multi-cloud override. This original nit remains resolved. Other findings are tracked in the [follow-up review](https://github.com/awslabs/startups/pull/247#pullrequestreview-5200827736).

---

Original comment:

[🤖 AI review 🤖] **Nit:** This table cell says "EKS Auto Mode under `compute: \"eks\"`" but the multi-cloud override paragraph 10 lines below correctly says "assign **EKS standard node groups**" and "**Auto Mode is not used for multi-cloud**." The rest of the PR consistently routes `compute: "eks"` (Q5 = multi-cloud) to standard node groups — `compute.md`, `clarify-global.md`, `fast-path.md`, `index.md`, and the override paragraph in this very file all agree.

This cell is the only place that says Auto Mode under multi-cloud. An agent reading the table first would pick Auto Mode, then hit the contradicting paragraph and have to reconcile. Should say "EKS standard node groups" to match.

(Same inconsistency appears in the migrate/ mirror at the same location.)

leon1418 (commented):
[🤖 AI review 🤖]

**Update (September 14, 2026): the CLEAN conclusion below is superseded by my [follow-up review at head ed268632](https://github.com/awslabs/startups/pull/247#pullrequestreview-5200827736), which requests changes.** The original App Engine billing nit is fixed. The current review identifies two P1 issues (final target enforcement and the Auto Mode NodeClass API) and one P2 issue (the fallback architecture value). I incorrectly accepted the EC2NodeClass guidance in this earlier review. The original review below is retained as historical context.

---

[🤖 AI review 🤖]

**Reviewed head `17c7e804` — 32 files, +422/−256 (all prose/Markdown, no code).**

This PR adds EKS Auto Mode as the default GKE migration target, displacing the previous Fargate default. The change is well-designed: GKE usage signals Kubernetes adoption, so staying on managed Kubernetes (Auto Mode as the GKE Autopilot equivalent) is the right default; users who want node control or want off K8s opt in explicitly via Q8. The three-way enum (`eks-auto` / `eks-standard` / `ecs-fargate`) threads cleanly through Clarify → Design → Estimate → Generate → Workshop, with correct multi-cloud exclusion (Auto Mode is AWS-specific, Q5 routes to standard EKS). herosjourney's blocking finding (arm64/NodePool) and both non-blocking items are all verified fixed at the current head.

**1 Nit (non-blocking):** `design-billing.md` table row for `google_app_engine_application` says "EKS Auto Mode under `compute: \"eks\"`" while the override paragraph in the same file — and every other location — correctly says "EKS standard node groups" for multi-cloud. An agent parsing the table first would pick the wrong target. Inline comment posted.

**What's correctly implemented (verified in the diff):**
- Autopilot detection normalized into `config.autopilot_enabled` on both IaC and live paths, with canonical schema keys documented in `schema-discover-iac.md`.
- Q8 restructured: options presented before recommendation, dead option D removed (Q8 only fires when GKE present), default changed from Fargate to Auto Mode with clear rationale.
- Auto Mode eliminator (node-level customization) falls back to `eks-standard`, not Fargate — preserves Kubernetes.
- Regional availability gate via `awsknowledge` MCP `get_regional_availability` before finalizing `eks-auto`, matching the existing Bedrock/AgentCore pattern.
- Graviton/arm64 on Auto Mode correctly documented as requiring custom Karpenter NodePool+EC2NodeClass manifest (not a Terraform flag), with amd64 fallback if manifest cannot be emitted.
- EKS Auto Mode pricing formula correctly separates control plane fee, EC2 instance rate, and Auto Mode management surcharge, with amd64-by-default note.
- Savings Plan guidance updated to clarify management fee and control plane are not SP-eligible.
- Q5 multi-cloud consistently routes to `eks-standard` across all 16 files (compute, clarify-global, fast-path, index, billing, estimate, generate, workshop, schema-preferences).
- Cross-plugin parity: all 16 advisor/migrate pairs verified byte-identical after path normalization. No new drift allowlist entries.

**CI:** No checks reported on this branch.
**Merge status:** MERGEABLE, 1 approval (herosjourney at current head), review requested from startups-advisor + startups-migrate teams.
**Verdict:** CLEAN. The Nit is a copy/paste remnant that does not block — the authoritative override paragraph in the same file has the correct text, and an agent that reads past the table will get the right answer. But fixing it removes the ambiguity.

leon1418 (dismissed):

advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/references/phases/design/design-billing.md
hasrazz: Updated it in: [e27ef1c](https://github.com/awslabs/startups/pull/247/commits/e27ef1c38bf1270d510e686888ed888046f62e01)

hasrazz:
> **1 Nit (non-blocking):** `design-billing.md` table row for `google_app_engine_application` says "EKS Auto Mode under `compute: \"eks\"`" while the override paragraph in the same file — and every other location — correctly says "EKS standard node groups" for multi-cloud. An agent parsing the table first would pick the wrong target. Inline comment posted.
> 
 updated this in [e27ef1c](https://github.com/awslabs/startups/pull/247/commits/e27ef1c38bf1270d510e686888ed888046f62e01)

hasrazz:
@leon1418, @herosjourney; All changes have been pushed and branch is updated with main.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/design-refs/fast-path.md
```diff
@@ -89,17 +89,17 @@ These GCP resources do **not** require AWS equivalents in v1.0:
...
+| ----------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Containerized workloads | Fargate (default for non-K8s containers, e.g. Cloud Run), Lambda (event-driven); **GKE / Kubernetes workloads → EKS Auto Mode by default** (standard EKS node groups only on explicit opt-in) | Deeper integration with VPC, ALB, IAM, and auto-scaling than lighter-weight alternatives; Auto Mode is AWS's recommended low-ops Kubernetes model and the GKE Autopilot equivalent |
```
ayn-builds: This row is the one place the multi-cloud exclusion is missing, and it's the place with enforcement authority over final mappings.

The PR states "Auto Mode is NOT used for multi-cloud" in five places - `clarify-global.md:205`, `compute.md:57`, `compute.md:81`, `design-refs/index.md`, and `design-billing.md:58`. But this row reads "**GKE / Kubernetes workloads → EKS Auto Mode by default** (standard EKS node groups only on explicit opt-in)" with no condition on `compute`, and `:98` says to check every mapping against this table before writing `aws-design.json` and substitute the preferred target.

Under Q5 = multi-cloud, Q8 is **skipped**, so there is no explicit opt-in to appeal to: `compute.md:57` maps GKE to standard node groups and no `kubernetes` key is ever written. A GKE resource then matches this row, whose preferred target is Auto Mode, and enforcement substitutes it.

Both enforcement exemptions (`:100`, `:102`) and the Precedence paragraph (`:104`) are scoped to App Engine and the PaaS row. `:104` even spells out that App Engine under `compute: "eks"` "matches no row here and keeps its rubric-selected target" - GKE has no equivalent escape.

The PaaS row already solves this exact problem by carrying its condition inside the row (`and compute ≠ "eks"`, see `:54` and `:94`). Suggest the same here: make the GKE clause conditional on `compute ≠ "eks"`, and note that under `compute: "eks"` the preferred target is standard node groups.

Worth treating as blocking - the failure mode is emitting an AWS-specific Auto Mode cluster for a customer who explicitly required multi-cloud portability, which is the one requirement this PR says Auto Mode must never serve.

Applies identically to the `advisor/plugins/aws-startup-advisor/...` mirror at the same line.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/design-refs/compute.md
```diff
@@ -73,17 +78,17 @@ Add a `graviton` block (see `schema-graviton.md`) to the service's output. GPU/C
...
+1. **Eliminators**: Does GCP config violate AWS constraints, or does a top-level preference override apply? If yes: switch to alternative. **`compute: "eks"` (Q5 = multi-cloud) is a hard override — App Engine (and all compute) → EKS standard node groups (portable Kubernetes; Auto Mode is NOT used for multi-cloud); stop here, do not evaluate the managed-platform/EB branch below.** For GKE (AWS-only, Auto Mode default), the Auto Mode eliminator applies: node-level customization Auto Mode disallows → `eks-standard`, not Fargate.
+2. **Operational Model**: Managed (Lambda, Fargate, **EKS Auto Mode** — AWS operates the nodes) vs Self-Hosted (EC2, **standard EKS managed node groups**)?
+   - Prefer managed unless: Always-on + high baseline cost → EC2, or the user explicitly wants node control → `eks-standard`
+   - For App Engine sources: Elastic Beanstalk (PaaS-to-PaaS) when `compute_model` is absent or `"managed_platform"` **and `compute` ≠ `"eks"`** (when `compute: "eks"`, the criterion-1 override already selected EKS Auto Mode)
```
ayn-builds: This contradicts criterion 1 three lines above.

`:81`: "**`compute: "eks"` (Q5 = multi-cloud) is a hard override — App Engine (and all compute) → EKS standard node groups (portable Kubernetes; Auto Mode is NOT used for multi-cloud)**"

`:84`: "(when `compute: "eks"`, the criterion-1 override already selected EKS Auto Mode)"

The pre-PR text said just "EKS", so it was consistent before. Should be "EKS standard node groups". `:88` in criterion 3 gets it right, so this looks like a single missed edit rather than a design question.

Applies identically to the `advisor/plugins/aws-startup-advisor/...` mirror at the same line.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/phases/estimate/estimate-infra.md
```diff
@@ -159,18 +159,20 @@ Calculate 3 cost tiers to show the optimization range:
...
+| Compute (EKS Auto Mode)     | EKS control plane fixed monthly (~$73/mo = $0.10/cluster-hr × 730) + underlying EC2 instance rate × 730 hours × node count + **Auto Mode management fee** (a per-instance surcharge EKS adds on top of the EC2 rate for Auto Mode-managed nodes; billed per-second, 1-min minimum — look it up via the EKS Auto Mode pricing page/MCP, do not hardcode a stale %). Price the EC2 rate for the node instance family — **amd64 by default** (the built-in `general-purpose` NodePool is amd64-only); use a Graviton rate only if the arm64 custom NodePool is actually applied (see `generate-artifacts-infra.md`). Size node count from the source GKE cluster (or a dev-tier floor of 2 when unknown). Note in the estimate that Auto Mode nodes are dynamically provisioned by Karpenter, so this is a baseline, not a cap. | `aws_config` (target node instance type, node count), source `google_container_cluster` sizing |
```
ayn-builds: The Auto Mode management fee is now part of the default GKE recommendation's cost, but it's the only rate in this table with no home in the pricing pipeline.

This row says to "look it up via the EKS Auto Mode pricing page/MCP, do not hardcode a stale %". That bypasses the four-tier ladder this same file defines at `:13-16`:

1. `pricing-cache.md` (primary)
2. MCP with recipes (secondary)
3. cache after MCP failure
4. "If a service is NOT in the cache AND MCP failed, set `pricing_source: \"unavailable\"`. Add to `services_with_missing_fallback` and warn the user."

Three gaps follow:

- **No cache row.** `pricing-cache.md` `### EKS` (lines 56-63, untouched by this PR) has only the cluster fee and closes with "Worker nodes billed separately as EC2 or Fargate." Auto Mode nodes are neither, so the section no longer covers the default GKE target. This also means `pricing:staleness` never sees the rate.
- **"Do not hardcode" forbids the tier-1 fix**, which is how every other rate in this table is sourced.
- **Tier 4 is never invoked.** The row doesn't say to add the fee to `services_with_missing_fallback`, and `pricing-fallback.md` won't cover it either - its null-emission rule is scoped to AI model costs and it explicitly says infrastructure prices from the cache "may still be used". `:13` likewise says "Infrastructure prices (Fargate, RDS, S3, etc.) remain reliable; only AI model prices need the stale flag."

Net effect when the MCP is unavailable: no cached rate, no fallback rule covering it, no instruction to flag it. The most likely outcome is the fee silently drops out and the new default recommendation is understated.

Either add an Auto Mode management-fee row to `pricing-cache.md` `### EKS` (and update that closing sentence), or route this row explicitly through tiers 2-4 including `services_with_missing_fallback`. The first is more consistent with how the rest of the table works.

Applies identically to the `advisor/plugins/aws-startup-advisor/...` mirror at the same line.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/phases/clarify/clarify.md
```diff
@@ -292,12 +292,12 @@ Every question in an **active** category gets exactly one disposition:
...
-| Q8    | PROPOSED (only when GKE present and Q5 ≠ multi-cloud)                                                                            | C — `ecs-fargate`                    | Assuming Fargate → no Kubernetes to operate; answer "EKS" to preserve your K8s investment                                                                                                                                                       |
+| Q8    | PROPOSED (only when GKE present and Q5 ≠ multi-cloud)                                                                            | A — `eks-auto`                       | Assuming EKS Auto Mode → keep Kubernetes with AWS running the nodes (GKE Autopilot equivalent); answer "manage nodes myself" for standard EKS or "drop Kubernetes" for ECS Fargate                                                              |
```
ayn-builds: This row and the defaults table at `:829` were both updated to `eks-auto`, but the assembled `preferences.json` example between them was not. At `:653-659`:

```json
"kubernetes": {
  "value": "ecs-fargate",
  "chosen_by": "default",
  "source": "default:Q8",
  "prompt": "How do you feel about Kubernetes? (default applied)",
  "design_consequence": "Assuming Fargate → no Kubernetes to operate",
  "question_id": "Q8"
}
```

`chosen_by: "default"` and `source: "default:Q8"` make this unambiguously the default-applied example rather than a user-chose-C illustration, and all three of `value`, `prompt`, and `design_consequence` are pre-PR text - `prompt` is the old Q8 title this PR renamed to "How do you want to run your Kubernetes workloads on AWS?". This is the block an agent copies when assembling `preferences.json`, so it competes directly with the row above.

Also `:255` still lists Q8 as "K8s sentiment", the framing this PR deliberately replaced.

For what it's worth I swept the rest of the repo: these are the only two stale sites in `gcp-to-aws`. The surviving `eks-managed`/`eks-or-ecs` hits are all in `heroku-to-aws` under its own `design_constraints.compute_target` key, where those values are still current, so they're correctly untouched.

Applies identically to the `advisor/plugins/aws-startup-advisor/...` mirror at the same line.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/shared/schema-discover-iac.md
```diff
@@ -144,6 +144,29 @@ Example:
 }
 ```
 
+**Canonical `config` keys for `google_container_cluster`** (Clarify Q8 default and the `compute.md` GKE rubric read these — use EXACTLY these keys):
+
+- `config.autopilot_enabled` — boolean. `true` when the GKE cluster runs in **Autopilot** mode, `false` for **Standard** mode. Normalized from Terraform `enable_autopilot` (IaC path) or the live `autopilot.enabled` capture field (live path). Omit only when neither source resolves it — an omitted value is treated as Standard/unknown by downstream defaults, so prefer writing an explicit boolean whenever the signal exists.
+- `config.node_machine_type` — the primary node pool `machine_type` when present (Standard clusters only; Autopilot has no node pools). Optional; used for Graviton signal and node sizing.
```
ayn-builds: `config.node_machine_type` is introduced by this PR and read nowhere.

It appears in exactly three places, all producers or this definition: here, `discover-iac.md:62`, and `discover-live.md:332`. I grepped both plugins for consumers and found none.

This line says it is "used for Graviton signal and node sizing", but neither holds today:

- `graviton.md` (changed by this PR) doesn't reference it - the new Auto Mode bullet at `:98` keys entirely off `kubernetes: eks-auto`.
- The Auto Mode estimate row (`estimate-infra.md:167`) is exactly the consumer that needs it, and instead says "Size node count from the source GKE cluster (or a dev-tier floor of 2 when unknown)" with key inputs listed as "`aws_config` (target node instance type, node count), source `google_container_cluster` sizing" - never the normalized key.

So the PR adds the input and the consumer that wants it without connecting them. Contrast `autopilot_enabled`, which is threaded properly with named readers at `clarify-compute.md:81`, `compute.md:51`, and `design-billing.md:58`.

Either wire it into the node-sizing inputs on that estimate row, or drop it and the two producer instructions until something reads it.

Applies identically to the `advisor/plugins/aws-startup-advisor/...` mirror at the same line.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/design-refs/compute.md
```diff
@@ -43,18 +43,23 @@ Note: Cloud Run maps to Fargate via deterministic fast-path ("Always"). The `com
...
+- **Auto Mode regional availability (gate before finalizing `eks-auto`):** validate EKS Auto Mode is available in the target region (`preferences.json → design_constraints.target_region`) via `get_regional_availability` from the `awsknowledge` MCP server — the same source-of-truth check the repo uses for AgentCore/Bedrock (see `references/shared/ai-migration-guardrails.md`). Auto Mode is broadly available in commercial regions, but **verify GovCloud/opt-in partitions** — a `fedramp`/GovCloud target (Q2) is the likely gap. If Auto Mode is **not** available in the target region, **fall back to `eks-standard`** (standard EKS managed node groups run anywhere EKS is available) in the **same region** — do not switch regions (Q1/compliance fixed it) and do not silently emit an Auto Mode cluster that will not deploy. Record the fallback in the mapping `rationale` and a `warnings[]` entry, and tell the user: "EKS Auto Mode isn't available in `<region>` — using standard EKS managed node groups instead." If the MCP call fails, proceed with `eks-auto` but add a `warnings[]` note to verify Auto Mode availability in the target region before applying.
```
ayn-builds: Two problems with this gate.

**The cited precedent doesn't provide a fallback.** This says the check is "the same source-of-truth check the repo uses for AgentCore/Bedrock (see `references/shared/ai-migration-guardrails.md`)". That file's static table is headed "AgentCore Regional Availability" (`:7-21`) and covers AgentCore Runtime, Harness, Memory, Gateway, and Policy only. There's no EKS entry, and no Bedrock model region table either. The MCP call pattern transfers; the fallback-on-failure that makes the pattern safe does not.

**The failure path does the thing the paragraph forbids.** Mid-paragraph: "do not silently emit an Auto Mode cluster that will not deploy." End of paragraph: "If the MCP call fails, proceed with `eks-auto` but add a `warnings[]` note". Since GovCloud/opt-in partitions are named as the likely gap and a `fedramp` target is exactly when this matters, failing open on the default is the risky direction. Suggest failing to `eks-standard` on MCP failure - it "runs anywhere EKS is available" per this same line, so it is the safe default - with the warning noting that Auto Mode was not confirmed rather than that it was not verified.

Separately worth confirming before relying on this: Auto Mode is a cluster *feature*, not a distinct service, so I'd check that `get_regional_availability` can actually express feature-level availability rather than just returning EKS's footprint. If it can't, this gate can't be satisfied as written.

Applies identically to the `advisor/plugins/aws-startup-advisor/...` mirror at the same line.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/design-refs/compute.md
```diff
@@ -73,17 +78,17 @@ Add a `graviton` block (see `schema-graviton.md`) to the service's output. GPU/C
...
-   - If `kubernetes` is **absent** → Fargate (treat same as `"ecs-fargate"` — do not default to EKS)
-   - If `cost_sensitivity` present and high → prefer Fargate (lower operational cost)
+   - If `compute = "eks"` (Q5 = multi-cloud) → **EKS standard node groups** for all compute, including App Engine (top-level portability override; overrides EB default and `compute_model`). Auto Mode is excluded here — it is AWS-specific and defeats multi-cloud portability.
+   - If `kubernetes = "eks-auto"` → **EKS Auto Mode** (keep Kubernetes, AWS runs the nodes)
+   - If `kubernetes = "eks-standard"` → **EKS with managed node groups** (user explicitly wants node control)
+   - If `kubernetes = "ecs-fargate"` → Fargate (user chose to drop Kubernetes)
+   - If `kubernetes` is **absent** → **EKS Auto Mode** for GKE sources (do not default to Fargate; GKE signals Kubernetes adoption). Non-GKE containers (Cloud Run) map to Fargate via their own fast-path.
+   - If `cost_sensitivity` present and high → still keep Auto Mode for GKE (lowest-ops K8s); the management fee is modest vs the ops savings — do not silently drop to Fargate against a K8s preference
```
ayn-builds: "the management fee is modest vs the ops savings" asserts the magnitude of the one rate this PR says cannot be known - `estimate-infra.md:167` instructs "look it up via the EKS Auto Mode pricing page/MCP, do not hardcode a stale %".

It's also the basis for overriding an explicit user signal. Relative to the `ecs-fargate` alternative, Auto Mode adds the $73/mo control plane plus the per-instance fee on top of the EC2 rate, so for a small GKE footprint with `cost_sensitivity: high` the delta isn't self-evidently modest.

The instruction itself is defensible - not silently dropping a Kubernetes user to Fargate on a cost signal is the right call. Suggest keeping the behavior but dropping the cost assertion: keep Auto Mode because the K8s preference outranks the cost signal, and surface the control-plane plus management fee in the estimate so the user can make that trade explicitly.

Applies identically to the `advisor/plugins/aws-startup-advisor/...` mirror at the same line.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/design-refs/fast-path.md
hasrazz: @ayn-builds 

The GKE row now carries the multi-cloud conditional guard: when `compute: "eks"` (Q5 = multi-cloud), it routes to EKS standard node groups instead of Auto Mode, matching the exclusion applied everywhere else in the PR. Fixed in 950ba21, mirrored to the advisor plugin.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/design-refs/compute.md
hasrazz: Good catch — the criterion-2 parenthetical contradicted the criterion-1 override. Changed it to "EKS with standard node groups" so it no longer implies Auto Mode under multi-cloud. Fixed in d425669, mirrored to the advisor plugin.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/phases/estimate/estimate-infra.md
hasrazz: The Auto Mode management fee now flows through the pricing pipeline like every other rate — sourced via the `awspricing` MCP with a `pricing-cache.md -> ### EKS` entry, no hardcoded percentage; if the MCP is unavailable it's flagged as missing rather than guessed. Fixed in 10bf883, mirrored to the advisor plugin.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/phases/clarify/clarify.md
hasrazz: Updated the assembled `preferences.json` example to `eks-auto` so it matches the row and the defaults table. Fixed in c365459, mirrored to the advisor plugin.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/shared/schema-discover-iac.md
hasrazz: Agreed — no consumer. `config.node_machine_type` is dropped from all three producer references; Design reads `aws_config.instance_type`, so nothing downstream needed it. Fixed in c4e12ee, mirrored to the advisor plugin.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/design-refs/compute.md
hasrazz: All three points addressed by removing the gate rather than patching it. We verified live via the `awsknowledge` MCP `get_regional_availability` that `Amazon EKS Auto Mode` is a queryable *feature* (not just the service) and returns `isAvailableIn` for all 39 regions where EKS exists — every commercial, GovCloud, China, and EU Sovereign Cloud region. Since Auto Mode tracks the full EKS footprint the gate never fired, it cited a precedent with no EKS fallback, and it failed open on MCP error. Replaced with a short note explaining why no gate is needed, contrasting with Bedrock model selection (`design-ai.md`) where real per-region variance warrants the check. Fixed in c083479, mirrored to the advisor plugin.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/design-refs/compute.md
hasrazz: Dropped the "modest vs the ops savings" assertion. Kept the behavior — a stated Kubernetes preference outranks the cost signal, so we don't silently drop to Fargate — but grounded it in preference ranking rather than a cost claim, and pointed to the estimate line items (EKS control-plane + Auto Mode management fee) where the user can weigh the trade explicitly. Fixed in db1c21e, mirrored to the advisor plugin.

hasrazz:
@ayn-builds thanks for the detailed and valuable review. I have addressed all your comments and resolved conflicts with main branch. 

Let me now if anything else needs to be updated.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/design-refs/fast-path.md
leon1418: [🤖 AI review 🤖]

**Rechecked at `ed268632`: the original multi-cloud case is fixed, but two other decisions still need protection here [P1].**

For a GKE source on the AWS-only path:

1. Q8 option 3 writes `kubernetes: "ecs-fargate"`, and `compute.md` selects Fargate. This row only preserves standard EKS for multi-cloud or an explicit `eks-standard` choice; its "otherwise EKS Auto Mode" clause therefore overrides the explicit Fargate choice when the mandatory enforcement below runs.
2. If the original preference is `eks-auto` but node requirements trigger the Auto Mode eliminator, the rubric selects standard node groups. That technical fallback is not an explicit Q8 opt-in, so this row can replace it with the incompatible Auto Mode target again.

`design-infra.md:95` invokes this check after the rubric; its exemptions cover App Engine and Cloud SQL, not these GKE cases. Please preserve the explicit Fargate opt-out and the rubric's technical fallback in the final target rule. Also ensure Generate uses the resolved per-resource target: `generate-artifacts-infra.md:354` currently branches on the original preferences, which can disagree with a technical fallback.

Both mirrors are affected. Please cover AWS-only GKE with Q8 option 3, and GKE with `eks-auto` plus a node-level blocker, through final Design and Generate output. This finding comes from tracing conflicting prompt instructions; I have not run an EKS deployment or model replay.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/phases/generate/generate-artifacts-infra.md
```diff
@@ -335,24 +335,24 @@ For each domain with resources in the generation manifest:
...
...e EB does not auto-create it in Terraform; do not reference an instance profile that no resource emits. Emit `setting` blocks for: `aws:autoscaling:launchconfiguration/IamInstanceProfile` (the instance profile just emitted); `aws:autoscaling:launchconfiguration/SecurityGroups` (an SG for the instances, required when they sit in private subnets); `aws:autoscaling:launchconfiguration/InstanceType` (from `instance_type`); `aws:elasticbeanstalk:environment/EnvironmentType` (from `environment_type`); **when `environment_type == "LoadBalanced"`, also `aws:elasticbeanstalk:environment/LoadBalancerType = application`** (ALB, not the Classic default) plus `aws:autoscaling:asg/MinSize` (from `min_instances`) and `MaxSize` (from `max_instances`); `aws:ec2:vpc/VPCId`, `Subnets`, and `ELBSubnets` — for **LoadBalanced**, instances in private subnets + ALB in public (`ELBScheme` public); for **SingleInstance** (no ALB) put the instance in a **public** subnet with `AssociatePublicIpAddress=true` so it is reachable; and `aws:elasticbeanstalk:application:environment` for env vars. ARM64: pick the `t4g.*` instance type from `aws_config` per the Graviton rule. See `design-refs/elastic-beanstalk.md`. |
```
leon1418: [🤖 AI review 🤖] **[P1] Use the Auto Mode NodeClass API for the arm64 pool**

When `graviton.target_architecture == arm64`, this instruction emits a custom `EC2NodeClass`. That is self-managed Karpenter's AWS-provider resource; EKS Auto Mode uses `apiVersion: eks.amazonaws.com/v1`, `kind: NodeClass`, and the NodePool must reference `group: eks.amazonaws.com`, `kind: NodeClass`. Following the current instruction will not configure an Auto Mode node class for the arm64 pool.

A custom class is also unnecessary solely to select arm64 in this configuration: the cluster enables built-in pools, so the `default` NodeClass exists, and AWS's custom NodePool example references it. The smaller fix is to emit an arm64 NodePool referencing that class. If a custom class is needed, use the Auto Mode NodeClass schema and ensure its node role has the required access entry.

Please update the repeated instruction at line 342 and `references/shared/graviton.md:98`, in both mirrors. Retain the workload selector/affinity and runbook apply step.

Official references: [Auto Mode NodePool example](https://docs.aws.amazon.com/eks/latest/userguide/create-node-pool.html) and [NodeClass schema and access-entry requirements](https://docs.aws.amazon.com/eks/latest/userguide/create-node-class.html).

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/shared/graviton.md
```diff
@@ -94,7 +94,8 @@ Emit ARM64 in Terraform:
...
-- EKS: arm64 AMI node group (single-arch on dev; note optional mixed-cluster module for prod)
+- EKS standard node groups (`kubernetes: eks-standard`): arm64 AMI node group (single-arch on dev; note optional mixed-cluster module for prod)
+- EKS Auto Mode (`kubernetes: eks-auto`, the GKE default): **Graviton is not a flag on the cluster.** Auto Mode's built-in `general-purpose` NodePool (the one that schedules app workloads) is **amd64-only and immutable** — only the tainted `system` NodePool (cluster add-ons) is multi-arch — so arm64 **cannot** be selected on `aws_eks_cluster` or the built-in pools. To actually run arm64 workloads you must apply a **custom Karpenter NodePool + EC2NodeClass** (`requirements: kubernetes.io/arch In ["arm64"]`) as a **Kubernetes manifest** via `kubectl`/Helm after the cluster is up (see `generate-artifacts-infra.md` for the emitted manifest + apply step). If the run does not emit and apply that manifest, Auto Mode nodes run **amd64** — set `target_architecture: amd64`, drop the arm64 build, and **flag Graviton as not applied**; never report arm64 for an Auto Mode cluster that only has the built-in `general-purpose` pool.
```
leon1418: [🤖 AI review 🤖] **[P2] Keep the fallback value consistent with the Graviton schema**

The Terraform-only fallback explicitly writes `target_architecture: amd64`. However, `references/shared/schema-graviton.md:24,100` defines `arm64` / `x86_64`, and `generate-artifacts-infra.md:341` selects its x86 branch with `x86_64`. The new fallback therefore writes a value outside the canonical artifact schema and the documented generator branch.

Please write `target_architecture: x86_64` in the internal artifact, keeping `amd64` for Kubernetes architecture labels and container-platform values. Apply the correction to both mirrors.

leon1418 (changes requested):
[🤖 AI review 🤖]

**Reviewed head `ed2686322e74ead329d385d88c9ec14ae98a60ce`: Request changes.** This supersedes my earlier CLEAN assessment. In particular, I incorrectly accepted the `EC2NodeClass` guidance in that review.

The owner has replied to all eight inline threads and implemented the original App Engine billing correction, the GKE multi-cloud guard, the criterion-2 wording correction, explicit missing-fee handling, the Q8 example/default cleanup, removal of the unused field, and removal of the unsupported fee-magnitude claim. The Q8 question also now presents choices before the default.

The regional-gate removal is supported by the current catalog: on September 14, I queried the exact `Amazon Elastic Kubernetes Service (EKS)` and `Amazon EKS Auto Mode` products across 39 regions, including China, GovCloud, and EU Sovereign Cloud. Both returned `isAvailableIn` throughout.

**Remaining findings:**

1. **[P1] Final target enforcement still overrides the explicit Fargate opt-out and the Auto Mode technical fallback.** [Follow-up in the existing mapping thread](https://github.com/awslabs/startups/pull/247#discussion_r4007793147).
2. **[P1] Auto Mode arm64 generation requires `NodeClass` under `eks.amazonaws.com/v1`, not self-managed Karpenter's `EC2NodeClass`.** The earlier arm64 blocker is therefore not fully fixed.
3. **[P2] The Terraform-only fallback writes `target_architecture: amd64`, while the canonical schema and generator branch use `x86_64`.**

All findings affect both plugin copies. Details and corrective actions are in the thread follow-up and inline comments.

**Validation:** all 17 changed mirror pairs are byte-identical; the cross-plugin drift check passes; the existing 136 tests pass. Those tests cover frontmatter and Heroku EB behavior, not EKS Auto Mode generation. The findings are based on current prompt/schema contracts and official AWS NodePool/NodeClass documentation. No EKS deployment or end-to-end migration replay was run.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/design-refs/fast-path.md
hasrazz: @leon1418 thanks for tracing this through Design and Generate — both cases were real.

Fixed in 40421c6 (with a small wording follow-up in e79b0c7). The root cause was the same in all three places: the Containerized-workloads enforcement row and the Generate EKS branch re-derived the Kubernetes target from the global Q8 preference and defaulted to Auto Mode, instead of honoring the per-resource target `compute.md` had already resolved into `aws_service`.

- `fast-path.md`: the row now defers to the resource's resolved `aws_service`, and a new GKE enforcement exemption (alongside the App Engine ones) tells the step not to re-open a resolved Fargate or Standard EKS Cluster result.
- `design-infra.md`: the Preferred-target check now skips Auto Mode substitution when the resolved target is Fargate (Q8 `ecs-fargate`) or a Standard EKS Cluster (`eks-standard` opt-in, `compute: "eks"` multi-cloud, or the Auto Mode eliminator's node-customization fallback).
- `generate-artifacts-infra.md`: the EKS branch now selects the cluster type from the resource's resolved `aws_service` in `aws-design.json`, not `design_constraints.kubernetes`, so the emitted Terraform matches Design even when a technical fallback changed the target.

Both mirrors, byte-identical. Separately (e69db62) we standardized the non-Auto-Mode naming to "Standard EKS Cluster" across the skill so the same target reads the same way in Clarify, Design, and Generate.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/phases/generate/generate-artifacts-infra.md
hasrazz: @leon1418 thanks — you're right, `EC2NodeClass` is the self-managed Karpenter API and Auto Mode wouldn't have reconciled it, so the arm64 pool would never have come up.

Fixed in f83b1ec. The arm64 manifest is now an Auto Mode `NodePool` (`karpenter.sh/v1`) whose `nodeClassRef` points at the built-in `default` NodeClass (`group: eks.amazonaws.com`, `kind: NodeClass`, `name: default`) with `kubernetes.io/arch In ["arm64"]`. No custom class is emitted to select arm64; the instruction now says explicitly not to emit a `karpenter.k8s.aws` `EC2NodeClass`, and only to emit an `eks.amazonaws.com/v1` `NodeClass` (with an EKS access entry for its node role) when the pool needs settings `default` can't express. We also dropped the incorrect claim that `default` is tied to the built-in pools — that false premise is what led to the custom-class path.

Updated in the EKS branch, the repeated CPU-architecture rule, and `graviton.md`, both mirrors. The workload `nodeSelector`/affinity note and the runbook `kubectl apply` step are retained. The Standard EKS Cluster path is unaffected — it sets arm64 via `aws_eks_node_group` and never emits this manifest.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/shared/graviton.md
hasrazz: @leon1418 good catch. Fixed in 890368f — the Terraform-only fallback now writes `target_architecture: x86_64`, matching the `arm64 | x86_64` enum in `schema-graviton.md` and the value Generate's x86 branch keys off. The other `amd64` mentions on that line are kept, since they describe the Kubernetes architecture label of the built-in `general-purpose` NodePool, where `amd64` is the correct term. Both mirrors.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/design-refs/fast-path.md
leon1418: [🤖 AI review 🤖]

Verified fixed at `bbbe52eb576753a5d53dfef5d6fe1285d42bb8e9`.

The preferred-target row and GKE exemption in `fast-path.md`, plus the check in `design-infra.md`, now preserve the resource's resolved `aws_service`. Generate explicitly reads that resolved target rather than re-deriving it from the global Q8 preference.

I traced the two previously failing cases through those instructions in both copies: AWS-only GKE with Q8 `ecs-fargate` stays Fargate; a resource whose original `eks-auto` preference is overridden by the node-customization eliminator stays a Standard EKS Cluster. The multi-cloud override, explicit standard-cluster choice, and ordinary Auto Mode default remain consistent.

The static source-contract checks pass on this head and detect the old defect at `ed268632`. This closes my follow-up finding in this thread. This validation checks the prompt contract; it is not a claim of an LLM migration replay or EKS deployment.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/phases/generate/generate-artifacts-infra.md
leon1418: [🤖 AI review 🤖]

Verified fixed at `bbbe52eb576753a5d53dfef5d6fe1285d42bb8e9` in both plugin copies.

The generated arm64 instruction now specifies a `karpenter.sh/v1` NodePool referencing `group: eks.amazonaws.com`, `kind: NodeClass`, `name: default`. The cluster still enables built-in pools, so that default class exists. The repeated CPU-architecture rule and `graviton.md` agree, and the workload selector/affinity and runbook apply step remain present. A custom class is only an optional Auto Mode `eks.amazonaws.com/v1` NodeClass with node-role access, while `EC2NodeClass` is explicitly prohibited.

I rechecked the current official [NodePool example](https://docs.aws.amazon.com/eks/latest/userguide/create-node-pool.html) and [NodeClass documentation](https://docs.aws.amazon.com/eks/latest/userguide/create-node-class.html). Static contract checks pass on this head and detect the previous API mismatch. This finding is resolved; no cluster deployment was performed.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/shared/graviton.md
leon1418: [🤖 AI review 🤖]

Verified fixed at `bbbe52eb576753a5d53dfef5d6fe1285d42bb8e9` in both plugin copies.

The fallback now writes `target_architecture: x86_64`, matching the canonical Graviton schema and Generate's x86 branch. The remaining `amd64` wording describes the Kubernetes architecture and is correct in that context.

The source-contract check passes on this head and rejects the previous `amd64` artifact value. This finding is resolved.

leon1418 (approved):
[🤖 AI review 🤖]

Reviewed exact head `bbbe52eb576753a5d53dfef5d6fe1285d42bb8e9`. All three findings from my previous Request changes review are resolved, and I found no remaining substantive concerns in the reviewed change. Approving this revision.

- Final enforcement and Generate preserve each resource's resolved Kubernetes target, including the explicit Fargate opt-out and the technical fallback from Auto Mode to a Standard EKS Cluster.
- The arm64 NodePool uses the Auto Mode `default` NodeClass with the correct API reference, while retaining workload scheduling and runbook instructions.
- The fallback artifact uses the canonical `x86_64` value.

I followed up in the original discussions. Validation on this head: all 18 changed mirror pairs are byte-identical; cross-plugin drift passes (275 identical, 25 allowlisted); all 149 existing tests pass; formatting, Markdown lint, and `git diff --check` pass. Static contract checks for all three fixes pass in both copies and detect the previous defects at `ed268632`. I also rechecked the current official AWS NodePool/NodeClass documentation.

These are local checks and a review of the prompt contracts. No end-to-end LLM migration replay or EKS deployment was run. GitHub currently reports no checks/status results for this head, although `build` is required; the branch is behind main and other review discussions remain unresolved. Those merge gates still need to be satisfied.

leon1418 (approved):
[🤖 AI review 🤖]

Re-reviewed exact head `ea2d67e952da27dfd84aff22368d5509b30b1dc6` against base `b30cb39be02d2eae78cb8c482f4018435d4dc988`, including the changes since my actually inspected head `bbbe52eb`. No remaining substantive findings; approving this revision.

The new commits merge upstream changes. The PR's added/deleted content is unchanged relative to its updated base. I checked the affected integration rather than relying on the previous approval's commit metadata: the resolved Fargate/Standard EKS targets still survive final enforcement and generation, the arm64 pool still uses the Auto Mode NodeClass API, and the fallback artifact remains `x86_64`. I also checked the merged Terraform policy consumer and the allowlisted advisor-only closing instructions. The EKS pricing section is preserved.

Local validation on this head:

- All 18 changed mirror pairs are byte-identical; drift passes (281 identical, 27 allowlisted).
- All 151 Node tests and 57 policy tests in each plugin copy pass.
- An EKS-shaped static policy probe passes, a wildcard-policy negative case fails, and fixing that policy replaces the canonical failed verdict with a passing one while preserving Terraform and intermediate phase state.
- The three original source-contract regression checks pass in both copies and detect the previous defects. Formatting, Markdown lint, and `git diff --check` pass.

These are local/static checks, not an LLM migration replay or EKS deployment. All ten review threads are resolved. GitHub still reports no check/status result for this head, so the required `build` gate remains outstanding.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/design-refs/compute.md
```diff
@@ -43,18 +43,23 @@ Note: Cloud Run maps to Fargate via deterministic fast-path ("Always"). The `com
...
+- **`kubernetes = "ecs-fargate"`** → **Fargate** (user chose to drop Kubernetes).
+- **Autopilot signal:** when the source `google_container_cluster` has `config.autopilot_enabled: true`, Auto Mode is the direct 1:1 mapping — record it in the rationale (`"GKE Autopilot → EKS Auto Mode (managed-node equivalent)"`). A Standard cluster (`autopilot_enabled: false`) still defaults to Auto Mode but `eks-standard` is the natural opt-out if the user wants to preserve node pools.
+- **Auto Mode eliminator:** if the workload needs node-level customization Auto Mode disallows (see Eliminators), fall back to `eks-standard`, not Fargate.
+- **Auto Mode regional availability:** no separate regional gate is required. EKS Auto Mode tracks the EKS service footprint — it is available in every current EKS region (all commercial, GovCloud, China, and EU Sovereign Cloud partitions) — so `eks-auto` deploys anywhere EKS itself is available. This differs from Bedrock model selection (`references/phases/design/design-ai.md`), where models have genuine per-region variance that warrants the `aws___get_regional_availability` check. Revisit only if a future region ships EKS before Auto Mode.
```
ayn-builds, line 53: **Regional claim is factually wrong, and it's load-bearing**

The Auto Mode regional availability note is used here to remove any regional gate from what is now the default GKE target, so the claim has to hold. It doesn't for China.

AWS's own EKS FAQ states Auto Mode is available "in all AWS Regions where EKS is available, including AWS GovCloud (US), except the China Regions." This line asserts the opposite:

> it is available in every current EKS region (all commercial, GovCloud, China, and EU Sovereign Cloud partitions)

EKS itself *is* available in `cn-north-1` and `cn-northwest-1`, so the "tracks the EKS service footprint" reasoning is what breaks: a GKE migration targeting either China region resolves to `eks-auto` with no gate, and `terraform apply` fails in the customer's account.

The GovCloud history also undercuts the premise - Auto Mode only reached GovCloud (US-East/US-West) in October 2025, well after GA, so its footprint has demonstrably lagged EKS's rather than tracking it.

Suggest: keep the no-gate default for commercial + GovCloud, add China as an explicit exclusion that falls back to `eks-standard`, and drop the "tracks the EKS service footprint" rationale since it's the part that generalizes incorrectly. EU Sovereign looks fine under the FAQ's wording.

Same fix needed in the `advisor/` mirror at this line.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/design-refs/compute.md
```diff
@@ -43,18 +43,23 @@ Note: Cloud Run maps to Fargate via deterministic fast-path ("Always"). The `com
 
 ### GKE
 
-- **Kubernetes orchestration explicitly required** (`kubernetes = "eks-managed"` or `"eks-or-ecs"` in `preferences.json`) → EKS
-- **Default / no explicit K8s preference** (`kubernetes = "ecs-fargate"` or absent):
-  - → **Fargate** (absent kubernetes preference resolves to Fargate, not EKS — teams that want EKS answer A or B in Clarify)
+GKE usage signals Kubernetes adoption, so the **default is to keep Kubernetes on EKS Auto Mode** — AWS operates the nodes (the GKE Autopilot equivalent, and AWS's recommended approach going forward). Branch on `preferences.json` → `design_constraints.kubernetes` (set by Q8):
+
+- **`kubernetes = "eks-auto"` or absent** → **EKS Auto Mode** (default). An absent preference resolves to Auto Mode, not Fargate — teams that want to manage nodes or leave Kubernetes say so in Clarify (Q8 B/C).
```
ayn-builds, line 48: **Q8 cited as letters against a now-numbered question**

Three references still use the old letter scheme after Q8 was renumbered to 1-4:

- `:48` (this line) "teams that want to manage nodes or leave Kubernetes say so in Clarify (Q8 B/C)"
- `:177` "Q8 = B (`kubernetes: "eks-standard"`)"
- `:185` "Q8 = C (`kubernetes: "ecs-fargate"`...)"

Should be 2 and 3 respectively. Line 167 already does this correctly by naming the constraint value instead of the option, which is the more durable form if you want to renumber again later.

Same fix needed in the `advisor/` mirror at these lines.

ayn-builds (commented):
Three blocking items. The first one is the substantive one; the other two are stale references left behind by the Q8 rewrite.

All three apply to both copies of the skill (`migrate/plugins/migration-to-aws/...` and `advisor/plugins/aws-startup-advisor/...`) at the same line numbers, since the 18 files this PR touches are identical pairs. The inline comments are anchored on the `migrate/` copy only; each fix needs to land in the `advisor/` mirror too.

---

**3. `references/phases/clarify/clarify.md:548` - trigger keys on an answer this PR deleted**

(Posting this one here rather than inline because the line is outside the diff hunks.)

The Answer Combination Triggers table still has:

> | Kubernetes-averse | Q5 = No + Q8 = Frustrated | ECS Fargate strongly recommended |

`Frustrated` was removed from Q8 in this PR, so this row can never fire. The equivalent is now `Q8 = 3` (`kubernetes: "ecs-fargate"`).

While in this table: line 546 ("Must stay portable ... EKS only, no ECS Fargate; App Engine also routes to EKS") should say Standard EKS Cluster. You updated exactly this wording in `clarify-global.md` but this copy was missed, and leaving it ambiguous here is what the rest of the PR works hard to disambiguate.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/design-refs/compute.md
leon1418: [🤖 AI review 🤖] **Confirmed at `ea2d67e952da27dfd84aff22368d5509b30b1dc6` — [P1]**

The current [EKS FAQ](https://aws.amazon.com/eks/faqs/) explicitly says Auto Mode excludes the China Regions. The [China regional documentation](https://docs.amazonaws.cn/en_us/aws/latest/userguide/eks.html) separately confirms that EKS itself is available in Beijing and Ningxia. The [October 22, 2025 GovCloud launch](https://aws.amazon.com/about-aws/whats-new/2025/10/amazon-eks-auto-mode-aws-govcloud-us-east-west/) also shows why service/feature footprint parity cannot be assumed.

I re-ran the exact availability-catalog query: it still returns `isAvailableIn` for Auto Mode in both China Regions. That conflicts with the explicit service FAQ. My earlier review incorrectly treated that catalog response as sufficient verification of the China claim; it should not override the documented exclusion.

The failure path is present in both mirrors: Clarify accepts a user `target_region` override; AWS-only GKE with `eks-auto` or no preference selects Auto Mode; Generate preserves that resolved target and enables `compute_config` in the requested region. The optional architecture check only warns or skips on unavailable evidence. The billing-only GKE row also defaults to Auto Mode.

Please exclude `cn-north-1` and `cn-northwest-1` when resolving an Auto Mode target on both the IaC and billing paths, fall back to a Standard EKS Cluster in the requested region with an explanation, and remove the blanket footprint-parity claim. Preserve that resolved fallback through final enforcement and Generate. This is a documented unsupported-target path; I did not attempt a deployment in China.

leon1418 (changes requested):
[🤖 AI review 🤖]

Reassessed exact head `ea2d67e952da27dfd84aff22368d5509b30b1dc6` after review `5250617626`. **Request changes for the China-region Auto Mode selection defect.** My previous no-findings recommendation missed this restriction.

**[P1] Regional selection:** the current EKS FAQ explicitly excludes China Regions, whereas `compute.md:53` asserts support there and removes the gate. The live availability catalog still reports support, but that conflicting response is insufficient to override the FAQ. A China-region override flows through the GKE default into Auto Mode generation; the billing-only mapping has the same unsupported default. I added evidence and the required fallback scope in the [existing regional discussion](https://github.com/awslabs/startups/pull/247#discussion_r4049306664), without opening a duplicate finding.

**Non-blocking consistency corrections:** I also verified the [stale Q8 letter references](https://github.com/awslabs/startups/pull/247#discussion_r4049034037) at `compute.md:48,177,185`, the removed `Q8 = Frustrated` trigger at `clarify.md:548`, and the ambiguous portable-target wording at line 547. Use the current constraint values (`eks-standard` / `ecs-fargate`) or options 2/3, and name Standard EKS Cluster explicitly. The current interpretation/rubric still maps option 3 to Fargate and multi-cloud to standard EKS, so I did not reproduce an additional target-selection failure from these stale labels.

The prior per-resource enforcement, NodeClass API, and `x86_64` fixes remain intact. Both plugin copies have the same new issues. Validation for this reassessment was a direct primary-documentation check and source/consumer trace, with 18 matching mirror pairs, a passing drift check, and a passing `git diff --check`. No implementation changes, deployment, new test infrastructure, or end-to-end model replay were performed.

herosjourney:
Looked back through the full review history here (27 reviews, 12 threads, 31 commits) to understand why this keeps finding new things each round. Sharing what I found in case it's useful for closing this out.

**This isn't a resolution loop.** 10 of 12 review threads closed cleanly on the first fix — author fixes, reviewer verifies, done. The two open threads from today are genuinely new findings, not regressions of anything already fixed.

**The recurring pattern is consistency drift, not contradictory review demands.** This PR threads a new enum value (`eks-auto`) through every layer of the pipeline — discovery, Clarify, Design, pricing, Terraform generation, schema — mirrored across two plugin trees (`migrate/` + `advisor/`). That's 36 files for one conceptual change. Almost every finding across all 12 threads (both resolved and open) is the same shape: a value renamed or introduced in one place that wasn't propagated everywhere it's referenced —

- the assembled `preferences.json` example that didn't match the renamed default
- the multi-cloud exclusion applied in 5 places but missing from a 6th
- today's stale "Q8 B/C" lettering in 3 spots after Q8 was renumbered elsewhere in this same PR

**One real structural issue did cause a fix to ripple (thread on `fast-path.md` → Generate):** the Kubernetes-target decision was computed independently in two places — Design's fast-path table and Generate's own re-derivation — so patching one didn't patch the other. That's now fixed and verified (`bbbe52e`).

**Today's regional-availability finding is a fact-check reversal, not new code breaking something:** the China-exclusion claim was live-verified via the `awsknowledge` MCP in an earlier round and the reviewer vouched for it (`c083479`); today's finding shows the MCP's `isAvailableIn` response conflicts with AWS's own published EKS FAQ, which explicitly excludes China. Worth noting for future regional-availability checks in this repo — the MCP catalog and the service FAQ apparently disagree here, so it's not a reliable single source for exclusions.

**Suggestion to close this out in one more round instead of another partial one:** rather than patching only the exact lines each review flags, do one deliberate sweep across both `migrate/` and `advisor/` trees for every reference to the renamed/new identifiers before the next push — specifically the remaining Q8 letter references (`:48`, `:177`, `:185` per the open thread) and the China exclusion (needs to land on both the IaC and billing paths, per the open thread). That would likely catch what the next review round would otherwise find piecemeal.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/design-refs/compute.md
hasrazz: fixed in [25c60e9](https://github.com/awslabs/startups/pull/247/commits/25c60e9cf694f02dc4c4e0e82359da04fa67cac6)

leon1418: [🤖 AI review 🤖]

Verified at `25c60e9cf694f02dc4c4e0e82359da04fa67cac6` in both mirrors. The three compute references now use Q8 options 2/3, the Kubernetes-averse combination uses option 3 / `ecs-fargate`, and the portability row explicitly names Standard EKS Cluster. These match the current Q8 interpretation and downstream target selection.

This closes the wording concerns from this thread and the associated combination-table comment. The separate regional-availability finding remains open.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/design-refs/compute.md
leon1418: [🤖 AI review 🤖]

Rechecked at `25c60e9cf694f02dc4c4e0e82359da04fa67cac6`: this P1 remains. `compute.md:53` still explicitly includes China and removes the regional gate; the billing-only GKE mapping also retains the Auto Mode default without an exclusion.

The [EKS FAQ](https://aws.amazon.com/eks/faqs/) fetched for this review still excludes China Regions. Both review perspectives independently traced the user-selected China region through Design and Generate. Please retain the requested region and resolve an unsupported Auto Mode target to Standard EKS on both mapping paths, as described in the original discussion. The Q8 wording fix does not change this path.

advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/references/phases/generate/generate-artifacts-infra.md
```diff
@@ -335,24 +335,24 @@ For each domain with resources in the generation manifest:
...
...e EB does not auto-create it in Terraform; do not reference an instance profile that no resource emits. Emit `setting` blocks for: `aws:autoscaling:launchconfiguration/IamInstanceProfile` (the instance profile just emitted); `aws:autoscaling:launchconfiguration/SecurityGroups` (an SG for the instances, required when they sit in private subnets); `aws:autoscaling:launchconfiguration/InstanceType` (from `instance_type`); `aws:elasticbeanstalk:environment/EnvironmentType` (from `environment_type`); **when `environment_type == "LoadBalanced"`, also `aws:elasticbeanstalk:environment/LoadBalancerType = application`** (ALB, not the Classic default) plus `aws:autoscaling:asg/MinSize` (from `min_instances`) and `MaxSize` (from `max_instances`); `aws:ec2:vpc/VPCId`, `Subnets`, and `ELBSubnets` — for **LoadBalanced**, instances in private subnets + ALB in public (`ELBScheme` public); for **SingleInstance** (no ALB) put the instance in a **public** subnet with `AssociatePublicIpAddress=true` so it is reachable; and `aws:elasticbeanstalk:application:environment` for env vars. ARM64: pick the `t4g.*` instance type from `aws_config` per the Graviton rule. See `design-refs/elastic-beanstalk.md`. |
```
leon1418: [🤖 AI review 🤖] **[P2] Disable self-managed add-on bootstrapping for Auto Mode**

The new branch enables Auto Mode's compute, load-balancing and storage blocks but leaves `bootstrap_self_managed_addons` unspecified. For the permitted AWS provider v5.100.0 (`~> 5.80`), that optional attribute defaults to `true` and is forwarded directly to `CreateCluster`; the provider's Auto Mode contract requires `false`. Following this prescribed configuration with provider defaults therefore sends an incompatible cluster-creation request.

Neither this skill nor the delegated Terraform posture supplies the missing setting. Please add `bootstrap_self_managed_addons = false` to the Auto Mode cluster instructions and its generation check in both mirrors.

Verified against the version-pinned [provider documentation](https://github.com/hashicorp/terraform-provider-aws/blob/v5.100.0/website/docs/r/eks_cluster.html.markdown) and [CreateCluster source](https://github.com/hashicorp/terraform-provider-aws/blob/v5.100.0/internal/service/eks/cluster.go). This is a provider-contract check, not a deployed failure capture.

advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/references/phases/estimate/estimate-infra.md
```diff
@@ -159,18 +159,20 @@ Calculate 3 cost tiers to show the optimization range:
...
...ontrol plane fixed monthly (~$73/mo = $0.10/cluster-hr × 730) + underlying EC2 instance rate × 730 hours × node count + **Auto Mode management fee** (a per-instance-type surcharge EKS adds on top of the EC2 rate, billed per-second, 1-min minimum, independent of EC2 purchase option — see `pricing-cache.md` → `### EKS`). Source it via the `awspricing` MCP (tier 2) for the exact per-instance-type rate; if MCP is unavailable, set `pricing_source: "unavailable"` for this fee, add it to `services_with_missing_fallback`, and warn that the estimate excludes the Auto Mode surcharge — do not silently drop it and do not substitute a hardcoded rate or percentage. Price the EC2 rate for the node instance family — **amd64 by default** (the built-in `general-purpose` NodePool is amd64-only); use a Graviton rate only if the arm64 custom NodePool is actually applied (see `generate-artifacts-infra.md`). Size node count from the source GKE cluster (or a dev-tier floor of 2 when unknown). Note in the estimate that Auto Mode nodes are dynamically provisioned by Karpenter, so this is a baseline, not a cap. | `aws_config` (target node instance type, node count), source `google_container_cluster` sizing |
```
leon1418: [🤖 AI review 🤖] **[P2] Price the planned ARM64 NodePool during Estimate**

For a fresh ARM64-compatible GKE migration, Design can select Auto Mode with `graviton.target_architecture: arm64`, but this row permits a Graviton rate only when the custom pool is **already applied**. Estimate runs before the opt-in Generate phase, so no such pool exists yet and this rule selects the amd64 baseline.

Generate later emits the ARM64 NodePool and runbook from that design. The decide-complete resume rule explicitly says never to rerun Estimate, while the generated Terraform header and `aligned_with_estimate_tier` claim that the output matches the Balanced estimate. The retained estimate can therefore describe a different architecture from the generated stack.

Please resolve manifest support and intended architecture in Design, price that prospective configuration here, and have Generate implement it. If Terraform-only output requires x86, resolve that fallback before pricing or explicitly reprice when it changes. Apply the same correction to both mirrors. This finding is based on the written phase/schema contracts; no model-generated cost report or deployment was executed.

leon1418 (changes requested):
[🤖 AI review 🤖]

Reviewed exact head `25c60e9cf694f02dc4c4e0e82359da04fa67cac6` against base `ec681ba5d49b039dbf1ef9d87d3304205b824f3d` using OCR delegation, one fresh independent general reviewer, a parent repository review, and candidate verification. **Request changes: one remaining P1 and two verified P2 findings.**

- **P1 — regional selection remains unresolved:** Auto Mode still defaults into China Regions despite the current EKS FAQ exclusion. Follow-up is in the existing regional thread.
- **P2 — Auto Mode creation:** the generation instructions omit `bootstrap_self_managed_addons = false`, exposing the allowed provider's incompatible default.
- **P2 — ARM64 estimate/generation alignment:** pricing requires the custom NodePool to be applied before the Generate phase that emits it, while later output retains and claims alignment with the earlier Balanced estimate.

The latest Q8 option references, removed-sentiment trigger, and Standard EKS wording are verified fixed in both copies; I replied in the original discussion. The earlier resolved-target, NodeClass API and `x86_64` fixes remain intact. The two new findings are inline and apply to both mirrors.

Validation: all 36 authored Markdown files were covered even though OCR excluded their extension; all 18 mirror pairs match. The parent ran 151 registered Node tests, formatting, Markdown lint, drift and `git diff --check` successfully. Provider defaults and phase contracts were checked against original source, with the new provider evidence independently refetched after candidate review. The required hybrid evidence gate passed.

These are source/provider-contract and local-check results. No end-to-end model migration or EKS deployment was run. GitHub CI and merge readiness remain separate requirements.

leon1418 (dismissed):

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/design-refs/compute.md
hasrazz: Confirmed that EKS Auto Mode is GA across all regions including gov cloud and china regions. EKS faqs is outdated and will be updated with new information.

Resolving this.

leon1418 (dismissed):

az-zhu (approved):

PR #249, opened 2026-08-26
fix(agent-advisor): lambda-microvms.networking — schema change

leon1418:

## What changed upstream

[AWS Lambda MicroVMs now supports AWS PrivateLink](https://aws.amazon.com/about-aws/whats-new/2026/08/lambda-microvms-supports-privatelink)

**Verdict: `schema_change`** on `lambda-microvms.networking`

- was — Networking: per-MicroVM URL over TLS; public service endpoints; VPC endpoints only if policy demands.
- now — Networking: per-MicroVM URL over TLS; public service endpoints OR private connectivity via AWS PrivateLink VPC Endpoints (for MicroVM control-plane API calls and per-MicroVM HTTP endpoint connections), available in all regions where Lambda MicroVMs is available.

> **Still true:** Public service endpoints and per-MicroVM URL over TLS remain valid for non-regulated/default use cases.
>

## Proposed edits

### `agent-advisor/references/decision-refs/lambda-microvms.md:44` — value

```diff
- Networking: per-MicroVM URL over TLS; public service endpoints; VPC endpoints only if policy demands.
+ Networking: per-MicroVM URL over TLS; public service endpoints OR private connectivity via AWS PrivateLink VPC Endpoints (for MicroVM control-plane API calls and per-MicroVM HTTP endpoint connections), available in all regions where Lambda MicroVMs is available.
```

**Why:** This is the exact old fact string for lambda-microvms.networking being replaced by the announcement of PrivateLink support for Lambda MicroVMs.

**Evidence (verbatim from the announcement):** "AWS Lambda MicroVMs now supports AWS PrivateLink, enabling private connectivity to Lambda MicroVMs directly from Amazon Virtual Private Cloud (VPC) resources without exposing traffic to the public internet."

The same edits were applied to this second copy: 1 applied.


az-zhu (approved):

herosjourney:
## Review

Verified independently (not just the PR body's claims):
- `advisor/` and `migrate/` copies are byte-identical, as claimed.
- `cross-plugin-drift.ts` passes (259 identical, 25 allowlisted).
- markdownlint and dprint both clean on the changed files.
- No content-lock test constrains this exact line, so nothing breaks downstream.
- This is a standalone fact card in `lambda-microvms.md`, not wired into any Terraform-generation logic, so a wording issue here can't cause a silent bad deploy — low blast radius either way.

**One accuracy note, non-blocking, worth a tweak before/after merge:**

I checked this against AWS's actual developer guide (`docs.aws.amazon.com/lambda/latest/dg/microvms-networking.html`), not just the announcement blog post cited in the PR. The real implementation uses **two distinct VPC endpoints**, not one generic "AWS PrivateLink VPC Endpoints":

- `com.amazonaws.region.lambda` — MicroVM **management APIs** (create images, run, suspend, terminate). This is the *existing* Lambda VPC endpoint, shared with regular Lambda.
- `com.amazonaws.region.lambda-microvm` — a *separate* endpoint for **connectivity to running MicroVMs** (per-MicroVM HTTPS traffic).

The proposed text ("private connectivity via AWS PrivateLink VPC Endpoints (for MicroVM control-plane API calls and per-MicroVM HTTP endpoint connections)") bundles both use cases into one undifferentiated phrase, which reads as if a single endpoint serves both purposes. Someone implementing from the card alone could create one endpoint and be confused why the other traffic path isn't private.

Suggested tightening, consistent with this file's short fact-card format:

> Networking: per-MicroVM URL over TLS; public service endpoints OR private connectivity via two separate PrivateLink VPC endpoints — `lambda` (management APIs) and `lambda-microvm` (MicroVM HTTP connectivity) — available in all regions where Lambda MicroVMs is available.

**Confirmed accurate against source:**
- "Available in all regions where Lambda MicroVMs is available" — verbatim in the AWS announcement.
- "Public service endpoints" remaining valid — correct; PrivateLink is additive, not a replacement.
- `schema_change` (not overwrite) is the right classification — the old value ("VPC endpoints only if policy demands") is still true, and the PR correctly preserves that nuance instead of deleting it.

**Verdict:** the substance is accurate and safe to merge as-is if the two-endpoint nuance isn't worth another review cycle — but since this is exactly the kind of judgment-call wording the PR body itself says "no CI check can catch," flagging it here for a maintainer call.

herosjourney:
## Review

Verified independently (not just the PR body's claims):
- `advisor/` and `migrate/` copies are byte-identical, as claimed.
- `cross-plugin-drift.ts` passes (259 identical, 25 allowlisted).
- markdownlint and dprint both clean on the changed files.
- No content-lock test constrains this exact line, so nothing breaks downstream.
- This is a standalone fact card in `lambda-microvms.md`, not wired into any Terraform-generation logic, so a wording issue here can't cause a silent bad deploy — low blast radius either way.

**One accuracy note, non-blocking, worth a tweak before/after merge:**

I checked this against AWS's actual developer guide (`docs.aws.amazon.com/lambda/latest/dg/microvms-networking.html`), not just the announcement blog post cited in the PR. The real implementation uses **two distinct VPC endpoints**, not one generic "AWS PrivateLink VPC Endpoints":

- `com.amazonaws.region.lambda` — MicroVM **management APIs** (create images, run, suspend, terminate). This is the *existing* Lambda VPC endpoint, shared with regular Lambda.
- `com.amazonaws.region.lambda-microvm` — a *separate* endpoint for **connectivity to running MicroVMs** (per-MicroVM HTTPS traffic).

The proposed text ("private connectivity via AWS PrivateLink VPC Endpoints (for MicroVM control-plane API calls and per-MicroVM HTTP endpoint connections)") bundles both use cases into one undifferentiated phrase, which reads as if a single endpoint serves both purposes. Someone implementing from the card alone could create one endpoint and be confused why the other traffic path isn't private.

Suggested tightening, consistent with this file's short fact-card format:

> Networking: per-MicroVM URL over TLS; public service endpoints OR private connectivity via two separate PrivateLink VPC endpoints — `lambda` (management APIs) and `lambda-microvm` (MicroVM HTTP connectivity) — available in all regions where Lambda MicroVMs is available.

**Confirmed accurate against source:**
- "Available in all regions where Lambda MicroVMs is available" — verbatim in the AWS announcement.
- "Public service endpoints" remaining valid — correct; PrivateLink is additive, not a replacement.
- `schema_change` (not overwrite) is the right classification — the old value ("VPC endpoints only if policy demands") is still true, and the PR correctly preserves that nuance instead of deleting it.

**Verdict:** the substance is accurate and safe to merge as-is if the two-endpoint nuance isn't worth another review cycle — but since this is exactly the kind of judgment-call wording the PR body itself says "no CI check can catch," flagging it here for a maintainer call.

PR #250, opened 2026-08-26
fix(gcp-to-aws, agent-advisor): give OpenRouter/LiteLLM-sourced OpenAI models the same-model Mantle path

herosjourney:
## Problem

The OpenAI → Bedrock model-selection logic defaults to "keep the same OpenAI model, land it on Bedrock's Mantle endpoint" for a direct-SDK OpenAI source. That default silently disappears for the same underlying model when it's called through a gateway (OpenRouter, LiteLLM) instead of the provider's SDK directly — in both `gcp-to-aws` and `agent-advisor`, though the failure mode differs between them:

- **`gcp-to-aws`**: discovery already tags gateway usage as `gateway_type == "llm_router"`, but the OpenRouter-specific guidance in `design-ai.md` presents three options — none of which is "keep the same model, move it to Mantle." The closest option ("Keep OpenRouter... `amazon/`-prefixed Bedrock models") is actually a switch to Amazon's own models, not a same-model move.
- **`agent-advisor`**: there's no gateway/router detection in Discover at all. A `base_url` pointing at OpenRouter, or an `OPENROUTER_API_KEY`, has no matching rule, so the underlying provider falls through to `provider: "unknown"`. The recommendation engine's routing is a two-way split (OpenAI → its own module, everything else → the Anthropic module), so an OpenRouter-routed OpenAI model never reaches the OpenAI module or its same-model-on-Mantle recommendation — it's scored as if it were an unclassified/Bedrock-native source.

A gateway is a transport layer, not a different model — a Mantle-eligible OpenAI model called through OpenRouter or LiteLLM should get the same default recommendation a direct-SDK caller of that model gets.

## Fix

**`gcp-to-aws`** (commit 1):

- **`design-ai.md`** — OpenRouter guidance now checks `ai_source == "openai"` for a Mantle-eligible model first. Added option A ("Same model via Mantle") ahead of the existing three (renumbered B/C/D), recommended when a Mantle target exists. Recorded value `code_migration.openrouter_path` gains `"same_model_mantle"`.
- **`clarify-ai.md`** — Split the Q23 auto-detect rule so LiteLLM and OpenRouter are no longer treated identically. LiteLLM stays a genuine one-line config-change default. OpenRouter now only auto-defaults to retarget when the underlying model is Mantle-eligible; otherwise it surfaces the full option set, since an OpenRouter → Mantle move is a real base-URL/credential/model-ID-format change, not a one-line swap.
- **`generate-artifacts-ai.md`** — Added the `"same_model_mantle"` branch, reusing the existing Step 1M Mantle-script generation (rather than new logic) with an explicit note that Step 1M's usual "no code changes required" claim does not hold for a gateway-routed source.

**`agent-advisor`** (commit 2 — amendment):

- **`discover.md`** — Added gateway/router detection to Step 1's model call inventory. A `base_url`/`OPENAI_BASE_URL` containing `openrouter.ai`, or `OPENROUTER_API_KEY`/`OR_API_KEY` present (same rule for a detected LiteLLM router), now classifies `provider` from the underlying model — OpenRouter model IDs are provider-prefixed slugs (`openai/gpt-5.4`), so the prefix names the real provider. The raw slug is kept unchanged in `model_ids[]` (the OpenAI module's family-detection regex already tolerates the prefix), and `sdk` records the transport (`"openai (via OpenRouter)"` / `"openai (via LiteLLM)"`) so it stays visible downstream.
- **`model-recommend.md`** — Noted that a gateway-routed OpenAI source still dispatches to the OpenAI module and its Mantle/Converse recommendation is unaffected, but `api_continuity` should be asked explicitly (closer to `preferred` than `required` by construction, since the source isn't calling the OpenAI SDK directly) and the caveat that landing on Mantle from a gateway is a real base-URL/credential/model-ID-format change — not the zero-code-change claim that applies to a direct SDK caller — should be surfaced in `[TUNE]`/rationale.

Mirrored to `migrate/plugins/migration-to-aws/` per the repo's drift-gate convention for all six touched skill files. (`agent-advisor/discover.md` has one pre-existing allowlisted wording difference between the two trees — "a separate migration plugin" vs "the migration-to-aws plugin" — so that file was edited directly rather than copied verbatim; the other three mirrors were byte-identical before this change and were copied over.)

## Verification

- `node advisor/plugins/aws-startup-advisor/tools/cross-plugin-drift.ts` → `OK (259 identical, 25 allowlisted across 6 skill trees)`
- `node .../tools/frontmatter-validator/validate.ts .../skills/agent-advisor` (both plugin trees) → `OK (11 phase file(s) with frontmatter checked)`
- `npx markdownlint-cli2` on all changed files → 0 errors

Not run: the fuller `mise run build` / pytest suites — this is a targeted prose/policy fix to reference files (discovery rules and model-selection guidance), no code paths in either plugin's Python engine were touched, and no content-lock test in `agent-advisor/scripts/test_*.py` references either changed file.

## Type of Change

- [x] Bug fix
- [ ] New plugin/power/tool
- [ ] Documentation update
- [ ] Guardrail/CI update

## Team Folder

- [x] `advisor/`
- [x] `migrate/`
- [ ] Other

## Checklist


leon1418 (commented):
[🤖 AI review 🤖]

**Review of PR #250 at head `3993b4fb`**

**Summary:** Targeted prose/policy fix that gives OpenRouter/LiteLLM-routed OpenAI models the same-model Mantle recommendation path they would get under direct-SDK detection. The fix closes a real gap where gateway-transported OpenAI models fell through to `provider: "unknown"` (confirmed in `model_recommendation.py` line 32: `_provider_module()` routes non-`"openai"` to the Anthropic module).

**Scope:** 10 files, +68/−16. Pure skill-reference prose (discovery rules, design guidance, clarify auto-detect, generate-artifacts branching, model-recommend caveats). No Python engine code paths modified.

---

### ✅ Design

Sound. The core insight — "a gateway is a transport layer, not a different model" — is correct and the fix follows logically:
1. Discover: classify provider from the model slug prefix, not as unknown
2. Model-Recommend: gateway-routed OpenAI still routes to the OpenAI module, but `api_continuity` is relaxed from `required` to `preferred`
3. Clarify: split LiteLLM (one-line config swap) from OpenRouter (real base-URL/credential change)
4. Design: add Option A (same-model Mantle) ahead of the existing three
5. Generate: route `same_model_mantle` to Step 1M with explicit caveat

The four-phase consistency (discover → recommend → clarify → design → generate) is maintained.

### ✅ Cross-Plugin Parity

All 5 file pairs verified:
- `model-recommend.md`: IDENTICAL ✓
- `clarify-ai.md`: IDENTICAL ✓
- `design-ai.md`: IDENTICAL ✓
- `generate-artifacts-ai.md`: IDENTICAL ✓
- `discover.md`: only pre-existing allowlisted diff (line 26 wording) ✓

No drift allowlist entries added. Compliant with mirroring policy.

### ✅ Correctness & Backward Compatibility

- New enum value `"same_model_mantle"` added to `openrouter_path` — no JSON schema constrains this field, so no breakage.
- Existing `"direct"` / `"litellm"` / `"keep_openrouter"` paths unchanged.
- `generate-artifacts-ai.md` fallback (absent `openrouter_path` → default `"litellm"`) preserved.
- The clarify-ai.md split correctly gates auto-retarget on Mantle eligibility for OpenRouter (vs unconditional for LiteLLM).

### ✅ Security

No secrets, no executable code changes, no user-input handling modifications.

### Findings

**0 mandatory blockers.**
**0 Nits.**

The change is well-scoped, internally consistent, correctly mirrored, and addresses a real functional gap. PR author's verification steps (drift check, frontmatter validator, markdownlint, dprint) are appropriate for a prose-only change touching no Python engine paths.

### Tests & CI

- No CI check runs present (statusCheckRollup is empty). The PR description acknowledges this and explains why targeted tests are not applicable (no content-lock test references the changed files, no Python engine code modified).
- Cross-plugin drift tool, frontmatter validator, markdownlint, and dprint all reported clean per author.

### Merge Readiness

**MERGEABLE** per GitHub. CLEAN — 0 blockers. Awaiting: CI checks to appear + maintainer approval (review requested from startups-advisor and startups-migrate teams).

leon1418 (approved):

PR #251, opened 2026-08-26
fix(gcp-to-aws): reduce feedback survey burden and fix pitch/reality mismatch

herosjourney:
## Problem

The feedback sidebar's initial pitch ("5 optional questions + anonymized usage data," shown at both offer points — after Discover and after Estimate) understates the actual ask. Saying yes produces up to **9 questions across 2 disconnected surfaces**:

1. Step 2.5 in chat: 3-4 questions (decision status, direction, missing evidence, and an optional migration-duration calibration question)
2. A separate 5-question web form the user has to open in a browser and paste a trace into

There's also an internal inconsistency: the Step 2.5 section header says "three quick questions" but numbers **four** items underneath it.

Given the plugin's own principle of never overselling estimates or effort, undercounting the feedback ask by ~2x (and splitting it across two channels without saying so) works against the same trust the tool is trying to build.

## Fix

**`feedback.md` Step 2.5:**
- Combined "could you make your decision" + "which way are you leaning" into a single multiple-choice question — one signal, one question, instead of two.
- Moved the migration-duration calibration question out of the default path. It was measuring something different (baseline duration data for the tool's own estimates) than decision quality, and doesn't need to be asked of every user by default — now it's only raised if the user volunteers relevant experience.
- Fixed the "three quick questions" / four-items-listed mismatch by removing the redundant question rather than relabeling the count.

**`SKILL.md` (both offer points — after Discover, after Estimate gate C):**
- Pitch text now says "one chat question + a short 5-question form" instead of "5 optional questions" — matches the actual ask before the user agrees to it.

**Net effect:** what was nominally 5 (but actually up to 9) questions across 2 channels is now 1 chat question + the 5-question form, accurately described upfront.

Mirrored to `migrate/plugins/migration-to-aws/` per the drift-gate convention — both files were byte-identical before this change and remain so after.

## Verification

- `node advisor/plugins/aws-startup-advisor/tools/cross-plugin-drift.ts` → `OK (260 identical, 25 allowlisted across 6 skill trees)`
- `npx markdownlint-cli2` on all 4 changed files → 0 errors
- No content-lock test (`scripts/test_*.py`) references the changed wording in either plugin tree

Not run: `mise run build` end-to-end / pytest — this is a prose-only change to phase instructions (feedback offer wording and the chat question sequence), no Python engine code was touched.

## Type of Change

- [x] Bug fix
- [ ] New plugin/power/tool
- [ ] Documentation update
- [ ] Guardrail/CI update

## Team Folder

- [x] `advisor/`
- [x] `migrate/`
- [ ] Other

## Checklist


herosjourney:
## Review summary

**Verdict: approve with small fixes applied** — the core problem diagnosis is correct and the direction is right. Pushed `3ab0b36` with follow-up edits.

### What works well
- Correctly identifies the trust issue: "5 optional questions" was understating a 2-surface, up-to-9-question ask.
- Collapsing decision status + lean into one MCQ is the right tradeoff (one signal, one question).
- Moving migration-duration calibration out of the default path is sensible — it was measuring something different from decision quality.
- Mirroring advisor/migrate and keeping drift green is correct.

### Fixes applied in `3ab0b36`
1. **Remaining pitch gap:** if the user picks **[C]**, Step 2.5 still asks a follow-up ("What evidence was missing?"). The SKILL.md pitch now says **"one quick chat question — two at most if you're still deciding"** so the upfront ask matches the C path.
2. **Internal contradiction:** removed prose claiming the user is "never asked two questions" while also defining a C follow-up.
3. **Form handoff:** restored **"open this short 5-question form in your browser"** — the PR dropped the browser cue, which matters since this is a separate surface.
4. **Duration opt-in:** added a one-line template for the optional calibration question when the user volunteers prior migration experience (was implied but not scripted).

### Non-blocking follow-ups (not in this PR)
- **heroku-to-aws** still pitches "5 optional questions" in `SKILL.md` / `feedback-collect.md` — no Step 2.5 chat questions there, so a lighter pitch update ("short 5-question form in your browser") would be a separate PR.
- **Gate A one-liner** ("Quick feedback before you go?") stays intentionally vague — OK for brevity since the full pitch runs only if they say Yes.
- **"Partly" signal** from the old Q1 is folded into **[C]**; acceptable loss of granularity unless product wants a distinct option later.

leon1418 (commented):
[🤖 AI review 🤖]

**Reviewed**: PR #251 at head `3ab0b36` · 4 files · +28/−32 · prose-only change to `gcp-to-aws` feedback phase

## Summary

This PR fixes a pitch/reality mismatch in the feedback sidebar: the old pitch said "5 optional questions" but the actual ask was up to 9 questions across 2 disconnected surfaces. The fix collapses the decision-check from 3-4 open-ended chat questions into 1 multiple-choice question (with at most 1 conditional follow-up), and updates the pitch text to accurately describe the new ask.

## Findings

**0 mandatory blockers. 0 Nits.**

Design is sound — combining "could you decide?" + "which way?" into a single A/B/C/D multiple-choice is strictly better UX than two separate open-ended questions, and moving the calibration question out of the default path is the right call (it measured something different from decision quality).

What I verified:
- **Cross-plugin drift**: advisor/ and migrate/ copies of both SKILL.md and feedback.md are **byte-identical** at this head. ✅
- **SKILL.md ↔ feedback.md consistency**: pitch text says "one quick chat question — two at most if you're still deciding" which exactly matches the Step 2.5 logic (1 MCQ default, +1 "what evidence was missing?" only if user picks C). ✅
- **No count mismatch**: the old "three quick questions" / four-items-listed bug is eliminated — now there is 1 question block with 4 choices (A/B/C/D), not 4 questions. ✅
- **Migration-duration calibration**: properly demoted to volunteer-only follow-up, not deleted — the signal is preserved for users who offer it. ✅
- **Trace contract**: unchanged — "Do NOT write the answer to trace.json or feedback.json" is maintained. ✅
- **heroku-to-aws**: not in scope for this PR (gcp-to-aws only); heroku-to-aws still carries the old "5 optional questions" wording on main — a follow-up PR could align it, but that is pre-existing and not introduced by this change.
- **No test files** reference the changed wording patterns (confirmed via code search on main). Author's verification section also confirms this.
- **CI**: 0 check runs reported on this branch.

**Verdict: CLEAN / MERGEABLE** pending CI green and maintainer approval. Not approved.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/phases/feedback/feedback.md
```diff
@@ -65,31 +65,29 @@ Then output the single-line minified version for copy-paste:
...
+
+Could you make your migrate/stay decision from this assessment?
+  [A] Yes — leaning migrate
+  [B] Yes — leaning stay
+  [C] Not yet — still deciding / need more evidence
+  [D] Skip
 ```
 
-Respond briefly to whatever they share (e.g. point at the relevant artifact or the what-if workshop for missing-evidence answers), then continue to the survey link below.
+If the user picks **C**, follow up with exactly one thing: "What evidence was missing?" — then respond by pointing at the relevant artifact or the what-if workshop. That is the only case where a second chat question is asked; do not chain further questions. A migration-duration calibration question is NOT asked here by default — it's out of scope for a decision-quality check; only if the user volunteers prior cloud-migration experience, you may ask one optional follow-up: "Roughly how long did start-to-cutover take?" (helps calibrate future duration guidance; skip if they decline).
```
ayn-builds, line 82: This paragraph rules out a third question and then adds one, in consecutive sentences: "That is the only case where a second chat question is asked; do not chain further questions" is immediately followed by "only if the user volunteers prior cloud-migration experience, you may ask one optional follow-up: 'Roughly how long did start-to-cutover take?'"

So [C] → "what evidence was missing?" → user mentions a past migration → duration question is three questions, against the "two at most" the offer text now promises.

Separately, that question's stated purpose cannot be met. It says it "helps calibrate future duration guidance," but there is no channel for the answer: line 70 forbids writing Step 2.5 answers to `trace.json` or `feedback.json`, `feedback.json` (lines 100-105) has no field for it, and `trace.json` is built in Step 1, before Step 2.5 runs. The answer is discarded on every path.

Simplest fix is to drop it, which also keeps the new pitch honest:

```suggestion
If the user picks **C**, follow up with exactly one thing: "What evidence was missing?" — then respond by pointing at the relevant artifact or the what-if workshop. That is the only case where a second chat question is asked; do not chain further questions. No migration-duration question is asked here: it is out of scope for a decision-quality check, and Step 2.5 answers are never persisted, so there would be nowhere to record the answer.
```

If you'd rather keep the duration question, it needs a persistence path and the offer text at `SKILL.md:375`/`389` needs to say three.

This file is byte-identical across `advisor`/`migrate`, so the same edit has to land in both copies or `drift:check` fails.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/phases/feedback/feedback.md
```diff
@@ -65,31 +65,29 @@ Then output the single-line minified version for copy-paste:
...
-4. If you've completed a cloud migration before: how long did it actually take,
-   start to cutover? (helps us calibrate — the plugin deliberately avoids
-   quoting week/hour estimates because it has no real-world duration data yet)
+One quick question before the survey link (optional):
+
+Could you make your migrate/stay decision from this assessment?
+  [A] Yes — leaning migrate
+  [B] Yes — leaning stay
+  [C] Not yet — still deciding / need more evidence
+  [D] Skip
```
ayn-builds, line 79: Consolidating the two old questions dropped the `defer` outcome. The old question 2 offered `migrate / stay / defer`; the new options are migrate, stay, "Not yet — still deciding / need more evidence", Skip.

A user who has decided, to defer, now has no truthful option. They land on [C], which records the opposite signal and triggers an evidence-gap follow-up they don't need, since nothing was missing.

```suggestion
Could you make your migrate/stay decision from this assessment?
  [A] Yes — leaning migrate
  [B] Yes — leaning stay
  [C] Yes — deferring for now
  [D] Not yet — still deciding / need more evidence
  [E] Skip
```

Two knock-ons if you take this: line 70's "**[D] Skip**" becomes **[E] Skip**, and line 82's "If the user picks **C**" becomes **D**. Worth confirming rather than assuming, though I checked and no schema, enum, or downstream consumer reads these answers, so the letters are referenced only in this file.

Same edit needs to land in the `advisor` copy too, since the two are byte-identical and not on the drift allowlist.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/SKILL.md
```diff
@@ -372,7 +372,7 @@ When invoked, the agent **MUST follow this exact sequence**:
 8. **Feedback sidebar**: After a phase completes, check if feedback is due (see rules below). This runs **before** advancing to the next phase.
 
    - **After Discover** (if `phases.feedback` is `"pending"`): Output to user:
-     "Would you like to share quick feedback (5 optional questions + anonymized usage data) to help improve this tool? Your data never includes resource names, file paths, or account IDs.
+     "Would you like to share quick feedback (one quick chat question — two at most if you're still deciding — plus a short 5-question form in your browser + anonymized usage data) to help improve this tool? Your data never includes resource names, file paths, or account IDs.
```
ayn-builds, line 375: This offer fires after Discover, and the phase order at line 114 is `discover → clarify → design → estimate → generate`, so there is no design and no estimate yet. Step 2.5 asks "Could you make your migrate/stay decision from **this assessment**?" when no assessment exists, so [C] is the near-universal answer and the real ask here is two chat questions, not the "one quick chat question" this leads with.

The same string at line 389 is accurate, because an estimate exists by then. So these two offer points should legitimately differ:

```suggestion
     "Would you like to share quick feedback (one or two quick chat questions, plus a short 5-question form in your browser, plus anonymized usage data) to help improve this tool? Your data never includes resource names, file paths, or account IDs.
```

Worth considering the cleaner alternative: line 395 says the decision-check is meant to "land immediately after the user actually made their migrate/stay decision," which argues for skipping Step 2.5 entirely when feedback is taken at this offer point and asking only the survey here. That would let the pitch keep saying one question.

(Line 395 also still reads "decision-check **questions**", plural, from the old three-question design. Outside this diff, but stale now.)

Same edit needs to land in the `advisor` copy to keep `drift:check` green.

PR #252, opened 2026-08-27
fix(gcp-to-aws): map Gemini multimodal embeddings to Nova Multimodal Embeddings

herosjourney:
## Problem

`ai-gemini-to-bedrock.md` (loaded by Design when `ai_source == "gemini"` or `"both"`) has no mapping row for Gemini's multimodal embedding model. The single existing embeddings row maps `text-embedding-004` (text-only) to Titan Embeddings V2 (also text-only) — correct for that model, but Gemini's newer embedding line (Gemini Embedding 2, public preview since March 2026) unifies text, image, video, audio, and PDF into one model, with no separate model ID for the multimodal case. Without a dedicated row, a multimodal embedding call silently falls through to the text-only Titan Embeddings recommendation, which is the wrong target for image/video/audio content.

This surfaced from a real customer migration case: multimodal embeddings via Gemini + Gemini Pro/Flash for inference, all via AI Studio API keys, currently on EC2.

## Fix

- **Pricing table (Legacy/Specialized section):** added a Gemini Embedding 2 row mapped to Amazon Nova Multimodal Embeddings, with real current pricing on both sides — Gemini text $0.20/1M tokens, image $0.45/1M ($0.00012/img), audio $6.50/1M, video $12.00/1M (from Google's published Gemini API pricing page); Bedrock text $0.135/1M, image $0.00006/unit (already in `shared/pricing-cache.md` § Multimodal embeddings, so no new pricing data needed on the Bedrock side). Bedrock comes out ~32% cheaper on text and ~50% cheaper on image.
- **Feature Migration Notes:** split the single `Embeddings` row into `Embeddings (text-only)` and `Embeddings (multimodal)`, with an explicit warning: do not route a multimodal Gemini embedding call to the text-only Titan row, and detect multimodal by call arguments (image/blob content present) rather than by model name alone, since Gemini's embedding models increasingly unify both into one model ID.
- **Context caching row:** sharpened while in the area, since the customer's use case combines context caching with image understanding — flagged that Gemini's TTL-based `cachedContent` objects and Claude's inline cache-control breakpoints are different mechanics requiring re-implementation, not a drop-in endpoint swap.

Mirrored to `migrate/plugins/migration-to-aws/` per the drift-gate convention (both files were byte-identical before this change).

## Scope

This is a **doc-only fix (Tier 1)**: it gets the right target model and caveat in front of a human reviewing Design's output. It does **not** teach Discover to automatically classify a multimodal vs. text-only embedding call from code — that's a larger follow-up (Tier 2), since it requires inspecting call-site arguments rather than just the method name or model ID, given that Gemini's embedding line no longer has separate model IDs for text vs. multimodal.

## Verification

- `node advisor/plugins/aws-startup-advisor/tools/cross-plugin-drift.ts` → `OK (260 identical, 25 allowlisted across 6 skill trees)`
- `npx markdownlint-cli2` on both changed files → 0 errors
- Pricing verified against Google's official Gemini API pricing page (ai.google.dev/gemini-api/docs/pricing) and AWS's existing `shared/pricing-cache.md` entry for Nova Multimodal Embeddings — not estimated.

Not run: `mise run build` end-to-end / pytest — this is a prose-only change to a design-reference doc, no Python engine code touched.

## Type of Change

- [x] Bug fix
- [ ] New plugin/power/tool
- [ ] Documentation update
- [ ] Guardrail/CI update

## Team Folder

- [x] `advisor/`
- [x] `migrate/`
- [ ] Other

## Checklist


leon1418 (commented):
[🤖 AI review 🤖]

**Review of rev at `86a7f95`** — 0 mandatory findings, 0 Nits.

### Change Summary

Adds Gemini Embedding 2 (multimodal) → Amazon Nova Multimodal Embeddings mapping to the `ai-gemini-to-bedrock.md` design reference doc. Splits the Feature Migration Notes embeddings row into text-only and multimodal with a clear routing warning. Sharpens the context caching row to note the mechanical difference between Gemini `cachedContent` TTL objects and Claude's inline breakpoints.

### Validation Evidence

| Check | Result |
|-------|--------|
| Cross-plugin drift | ✅ OK — `260 identical, 25 allowlisted across 6 skill trees` (ran `cross-plugin-drift.ts`) |
| Byte-for-byte parity | ✅ `advisor/.../ai-gemini-to-bedrock.md` and `migrate/.../ai-gemini-to-bedrock.md` are identical (`diff` exit 0) |
| Drift allowlist | ✅ No allowlist file modified — no bypass |
| markdownlint-cli2 | ✅ 0 issues on both changed files |
| Bedrock pricing source | ✅ `shared/pricing-cache.md` § Multimodal embeddings confirms text $0.135/1M, image $0.00006/unit — matches PR |
| Gemini pricing source | Author cites `ai.google.dev/gemini-api/docs/pricing` — cannot independently verify from repo data, but numbers are internally consistent (text $0.20 → ~32% cheaper at $0.135; image $0.00012 → ~50% cheaper at $0.00006) |
| CI | ⚠️ No checks reported on branch — CI has not run |

### Design Assessment

The change is well-scoped: it fills a documented gap (multimodal embedding calls silently falling through to the wrong text-only target) with correct model mapping and pricing, without touching engine code. The context caching clarification is a valuable drive-by improvement that prevents a real migration pitfall (treating endpoint-swap as sufficient). The DO NOT route warning in Feature Migration Notes is appropriately emphatic given the failure mode is silent wrong-model usage.

### Conclusion

**CLEAN / MERGEABLE** pending CI going green and maintainer approval. No drift allowlist abuse, cross-plugin parity confirmed byte-identical, doc-only change to a design reference with correct sourced pricing.

leon1418 (approved):

PR #253, opened 2026-08-28
feat(solution-architecture): aws-startups-solution-architecture plugin, contribution gate, and advisory reviewer

rsmets:
Draft. Opened to exercise CI, including the reviewer this PR introduces, before asking for review.

**RFC:** #254

Filed per `solution-architecture/CONTRIBUTING.md`. Noting the ordering honestly: this PR introduces that RFC requirement, so the work preceded the rule. The RFC documents the proposal, records the alternatives, and carries the open questions worth a second opinion, including whether the boundary with AWS Startup Advisor and with `aws-agents:agents-optimize` is drawn correctly.

Successor to #212, which removed `aws-dev-toolkit`. That plugin was deleted because its 38 skills and 11 agents had grown into a duplicate of [Agent Toolkit for AWS](https://github.com/aws/agent-toolkit-for-aws). This adds a replacement scoped so that cannot recur, plus the checks that enforce the scope.

## What this adds

**Plugin: `aws-startups-solution-architecture`** (2 skills)

| Skill | Problem it solves |
| --- | --- |
| `multi-tenant-isolation` | Tenancy model per layer for a B2B SaaS with no dedicated security engineer, isolation enforced in IAM session policies and Postgres RLS so a forgotten predicate fails closed, per-tenant cost attribution across a pooled fleet, and what to do when one large customer demands dedicated infrastructure mid-deal. |
| `agentcore-patterns` | Running a judgment agent inside a CI path: what it may block, measuring verdict stability before it gates anything, confidence banding, warm-container staleness. |

All general-purpose AWS service depth comes from `aws-core` and `aws-agents` as **upstream dependencies**, never restated. Every upstream pointer is written as an invocable `Skill("aws-core:aws-iam")` reference so a model can act on it; all 16 were verified to resolve.

**Contribution gate** (`solution-architecture/CONTRIBUTING.md` + `.github/workflows/contribution-gate.yml`)

Encodes the three criteria: startup-specific, no overlap with Agent Toolkit for AWS, no deprecated or sunset services. The deterministic half runs as a no-credentials job so it works on fork pull requests, checking `audience: startup` frontmatter, banned naming, uncaveated sunset-service references, orphaned reference files, broken `Skill()` pointers, and prose style. It explicitly does **not** decide criteria 1 and 2, so a green check means "nothing mechanically wrong", not "accepted".

The sunset check is context-aware rather than a word ban: warning about App Runner passes, recommending it fails. It also reads every markdown file, not just `SKILL.md`, because the removed plugin's uncaveated App Mesh and App Runner recommendations lived in `references/`.

**Advisory reviewer** (`.github/workflows/agentcore-reviewer.yml`)

Judges criteria 1 and 2, which a script cannot. Runs on Amazon Bedrock AgentCore as the `solution-architecture-reviewer` GitHub App (org-owned, App ID `4729749`, approved via P495031024).

Advisory in three concrete ways: it posts a comment and never an APPROVE or REQUEST_CHANGES review, so it can neither satisfy nor block a required human review; it never fails the job on a verdict; and it says so in its own output.

That posture is deliberate and measured, not cautious by default. Four runs against one unchanged pull request produced `REQUEST_CHANGES`, `APPROVE`, `REQUEST_CHANGES`, `APPROVE`. A gate that flips on a rerun teaches contributors to ignore it. Internal guidance on GenAI in CI/CD also requires human approval for GenAI-initiated actions. Details in `skills/agentcore-patterns/references/git-code-reviewer-agent.md`.

**Authentication:** GitHub OIDC, no long-lived AWS keys. The assumed role's trust policy is scoped to this repository and its only permission is `InvokeAgentRuntime` on a single runtime ARN. The two `AGENTCORE_REVIEWER_*` repository variables are ARNs rather than secrets: they are identifiers, and access requires a matching OIDC token.

## Nothing was ported

Per the deprecation guidance. An audit of all 38 removed skills and 11 agents found the startup framing lived entirely in the old `plugin.json` description: the word "startup" meaning a company appeared in no skill or agent body, and there were zero references to Activate, credits, runway, or stage. Only `aidlc-lite` had genuine founder framing, and its salvageable insight is folded into the new content. This PR also references no deprecated or sunset service.

## Note for reviewers

The reviewer will comment on this pull request, which is the one that introduces it. That is intentional dogfooding. Its output is advice with no authority, and a human still decides.

`architect-for-startups` in AWS Startup Advisor already owns founder-facing stage, runway, and credits guidance. This plugin is deliberately scoped to the technical side, where startup constraints reshape the correct answer, and routes business-facing questions there. Worth a second opinion on whether that boundary sits where you would put it.

## Verification

`claude plugin validate` passes for the marketplace and the plugin manifest; `skill-creator`'s `quick_validate.py` passes on all three skills; the gate passes; markdownlint clean across 867 files; dprint idempotent. The reviewer was confirmed end to end with `identity: app` against this repository.

## CODEOWNERS

Touches `.github/`, `marketplace.json`, `**/.claude-plugin/`, and `**/SKILL.md`, so this needs `@awslabs/startups-admins` in addition to `@awslabs/startups-solution-architects`.


rsmets:
> * 🟡 **Activation quality** — `agentcore-patterns`'s `SKILL.md` description is long and tries hard to carve a narrow niche away from [`aws-agents:agents-optimize`](https://github.com/aws/agent-toolkit-for-aws/blob/main/plugins/aws-agents/skills/agents-optimize/SKILL.md), but the phrase "measuring verdict stability before an agent gates anything" and "deciding which findings may block a merge" sit close enough to `agents-optimize`'s "CI/CD quality gates" that a model may struggle to choose between them reliably; a human should judge whether the negative-scope sentence at the end of the description ("Not for building, deploying, or hardening an agent... or for runtime selection") sufficiently disambiguates from `agents-optimize` specifically, which it does not explicitly exclude.

Looking at optimize's skill references I see no reference for using AgentCore for CI/CD. I believe this a true gap that startup customers would benefit from.

relentlesscol (dismissed):

igkins (dismissed):
looks good.

relentlesscol (approved):

solution-architecture/tools/contribution-gate/check.mjs
```diff
@@ -0,0 +1,357 @@
...
+  { pattern: /\blaunch configuration/gi, name: "EC2 launch configurations" },
+  { pattern: /\bCodeCommit\b/gi, name: "AWS CodeCommit" },
+  { pattern: /\bCloud9\b/gi, name: "AWS Cloud9" },
+  { pattern: /\bSimpleDB\b/gi, name: "Amazon SimpleDB" },
+];
+
+// A sunset mention is permitted when the surrounding line frames it as a
+// warning, a deprecation note, or a migration away from the service.
+const WARNING_CUE =
+  /\b(deprecat|sunset|end of support|end-of-support|closed to new|do not|don't|avoid|instead of|migrat|no longer|retir|legacy|EOL|rather than|not for new|stop)/i;
```
ayn-builds, line 42: **The sunset allowlist exempts the phrasings a recommendation actually uses.**

`WARNING_CUE` includes `instead of`, `rather than`, `migrat`, `avoid`, `do not`, `stop`, words that show up when recommending a service, not warning against one. Running the gate as-is:

| Line | Gate |
| --- | --- |
| `Use App Runner instead of ECS for a simple container service.` | exit 0 |
| `Migrate to App Runner for the fastest path to production.` | exit 0 |
| `Prefer CodeCommit over GitHub to avoid a third-party dependency.` | exit 0 |
| `App Runner is a good default.` | flagged |

Only the bare assertion is caught, so the check is weakest on exactly the case it exists for: the removed plugin's App Runner and App Mesh recommendations. And that advice cannot be followed, since App Runner closed to new customers on April 30, 2026 ([availability change](https://docs.aws.amazon.com/apprunner/latest/dg/apprunner-availability-change.html)).

The cue is also line-scoped, so any unrelated cue word on the line exempts it. `CONTRIBUTING.md:178` names both App Mesh and App Runner uncaveated and passes only because the same line contains "sunset", in the phrase "uncaveated sunset-service references".

Suggestion: keep only status words (`deprecat`, `sunset`, `retir`, `no longer`, `closed to new`, `end of support`, `EOL`), drop the directional ones, and add a negative check for `use|prefer|recommend|choose|migrate to` near the match.

.github/workflows/agentcore-reviewer.yml
```diff
@@ -0,0 +1,268 @@
...
+      - 'solution-architecture/**'
+
+permissions:
+  contents: none
+
+jobs:
+  review:
+    name: Advisory review
+    # Same-repo pull requests only; see the fork note above.
+    if: github.event.pull_request.head.repo.full_name == github.repository
```
ayn-builds, line 53: **Fork pull requests get no criteria 1 and 2 review, and no signal that they did not.**

This job self-skips on forks. `contribution-gate.yml` is credential-free precisely because "external contributions are the case the gate exists for", and it explicitly does not decide criteria 1 and 2. So the contribution path the gate was built for gets no automated look at the load-bearing criterion. The reference file this PR ships already says why: `git-code-reviewer-agent.md:136-140` tables `pull_request` as same-repo only and calls `workflow_run` "the safe default".

It also fails silently. Because the job-level `if` is what skips, the `Signal review finished` step at line 244 never runs despite `if: always()` (step-level `always()` only applies inside a job that is already running), so a fork PR carries no `agentcore-reviewer` status at all. Against this file's own principle that "an unreachable runtime must never read as a pass", absent is worse than a visible skip.

Suggestion: move to `workflow_run`, or post a "skipped, forks unreviewed" status from a separate always-running job.

ayn-builds (commented):
Two findings on the enforcement code, both cases where a gate silently passes what it exists to catch. Verified by running `check.mjs` against fixtures on this branch. Neither is a judgment call on criteria 1 or 2.

igkins (approved):
lgtm

solution-architecture/tools/contribution-gate/check.mjs
rsmets: Thanks for calling this out! If I push another commit it will dismiss the reviews. Given I can merge now without needing rebase/merge `main`. I think I am going to take it then immediately open another PR to address these couple valid call outs.

.github/workflows/agentcore-reviewer.yml
rsmets: https://github.com/awslabs/startups/pull/253#discussion_r3907858606

PR #257, opened 2026-09-01
fix(agent-advisor): correct Registry guidance and regional checks

leon1418:
## Summary

The Registry guidance described AWS Agent Registry and an additional internal AgentCore registry as two complementary products, restricted discovery to multi-agent workloads, and omitted Registry-specific regional availability. This could produce an incorrect service recommendation even when the compute runtime was available.

The updated guidance describes the same Registry moving from the `bedrock-agentcore` namespace to the dedicated `agent-registry` namespace. It recommends Registry for shared catalogs, resource discovery, or approval-based governance, including single-agent workloads; agent count alone does not trigger it.

The five GA Regions are recorded as a dated snapshot in the service card's Hard limits and the runtime profile's `registry_regions` volatile fact. Confirm, Design, and Add Capabilities require Registry-specific verification when it is selected, including for non-AgentCore runtimes. Unsupported or unverified Regions cannot complete a recommendation enabling Registry. If the user omits Registry, Confirm and Design service lists stay synchronized so later POC generation does not restore it.

All six affected reference files have matching changes in the `migrate/` and `advisor/` copies, preserving their existing intentional differences.

## Validation

- Existing Agent Advisor Python suites: 322 passed in each plugin.
- Both Agent Advisor frontmatter validators passed.
- Cross-plugin drift check passed; all six modified edit sets were also compared directly, including allowlisted files.
- Markdown lint, formatting, JSON snapshot checks, and `git diff --check` passed.
- Instruction walkthrough covered a single agent needing a catalog, multiple agents without discovery/governance needs, unsupported Regions, unavailable verification, non-AgentCore runtimes, and the separate Add Capabilities branch.

These checks validate the reference contracts and existing scripts; no live Registry deployment or end-to-end agent execution was performed.

## Sources

- [AWS Agent Registry GA announcement, August 31, 2026](https://aws.amazon.com/about-aws/whats-new/2026/08/aws-agent-registry-generally-available)
- [AWS Agent Registry overview](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/registry.html)
- [Registry namespace migration guide](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/registry-faq.html)


advisor/plugins/aws-startup-advisor/skills/agent-advisor/references/decision-refs/agentcore.md
```diff
@@ -58,7 +58,7 @@ An always-on service is still a better ECS/EKS fit — Instances sessions end at
 
 - Payments: for agents that pay / transact on a user's behalf — surface only for
   transactional / high-risk-action workloads.
-- Registry: agent/tool discovery and multi-agent orchestration — surface only for multi-agent setups.
+- Registry: agent/tool discovery and multi-agent orchestration — surface only for multi-agent setups. Note: AWS Agent Registry (GA) is now a distinct, centralized discovery/governance catalog for agents, tools, skills, MCP servers, and custom resources across the org, accessible via console/CLI/SDK/MCP and integrated with Bedrock AgentCore, Amazon Quick, and Kiro IDE — this is separate from (and complements) AgentCore's own internal registry-style discovery/orchestration feature.
```
ayn-builds: Blocking, two issues in this one line.

**1. AWS Agent Registry is not separate from AgentCore's registry - it is that registry.** The closing clause ("separate from (and complements) AgentCore's own internal registry-style discovery/orchestration feature") describes two products where there is one. The preview announcement calls Agent Registry "a new service offered through Amazon Bedrock AgentCore", and the docs live at `bedrock-agentcore/latest/devguide/registry.html`. As written, the advisor will tell customers to adopt one registry and complement it with a second that doesn't exist.

**2. GA is five Regions, stated here as "across the org" with no caveat.** us-east-1, us-west-2, ap-northeast-1, ap-southeast-2, eu-west-1. A customer in eu-central-1 currently gets this recommended with no signal that it's unavailable to them.

```suggestion
- Registry: agent/tool discovery and multi-agent orchestration — surface only for multi-agent setups. Note: AWS Agent Registry (GA as of 2026-09; verify via MCP — volatile) is AgentCore's centralized discovery/governance catalog for agents, tools, skills, MCP servers, and custom resources across the org, accessible via console/CLI/SDK/MCP and integrated with Amazon Quick and Kiro IDE. GA Regions: us-east-1, us-west-2, ap-northeast-1, ap-southeast-2, eu-west-1 — confirm the customer's Region before recommending.
```

The Region list would be better placed in the card's "Hard limits (verify via MCP — volatile)" section and in `references/runtimes/agentcore.json` `volatile_facts`, since only those get re-verified. Inlining it here is the minimum to stop the wrong recommendation.

migrate/plugins/migration-to-aws/skills/agent-advisor/references/decision-refs/agentcore.md
```diff
@@ -58,7 +58,7 @@ An always-on service is still a better ECS/EKS fit — Instances sessions end at
 
 - Payments: for agents that pay / transact on a user's behalf — surface only for
   transactional / high-risk-action workloads.
-- Registry: agent/tool discovery and multi-agent orchestration — surface only for multi-agent setups.
+- Registry: agent/tool discovery and multi-agent orchestration — surface only for multi-agent setups. Note: AWS Agent Registry (GA) is now a distinct, centralized discovery/governance catalog for agents, tools, skills, MCP servers, and custom resources across the org, accessible via console/CLI/SDK/MCP and integrated with Bedrock AgentCore, Amazon Quick, and Kiro IDE — this is separate from (and complements) AgentCore's own internal registry-style discovery/orchestration feature.
```
ayn-builds: Same two issues as the `advisor/` copy, since the two files are byte-identical - applying one suggestion without the other will break `cross-plugin-drift.ts`.

```suggestion
- Registry: agent/tool discovery and multi-agent orchestration — surface only for multi-agent setups. Note: AWS Agent Registry (GA as of 2026-09; verify via MCP — volatile) is AgentCore's centralized discovery/governance catalog for agents, tools, skills, MCP servers, and custom resources across the org, accessible via console/CLI/SDK/MCP and integrated with Amazon Quick and Kiro IDE. GA Regions: us-east-1, us-west-2, ap-northeast-1, ap-southeast-2, eu-west-1 — confirm the customer's Region before recommending.
```

advisor/plugins/aws-startup-advisor/skills/agent-advisor/references/decision-refs/agentcore.md
leon1418: Fixed in c40f50c4. Removed the second/internal-registry claim. The [current migration guide](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/registry-faq.html) clarifies that this is the same Registry moving from `bedrock-agentcore` to its own `agent-registry` namespace.

Moved the five GA Regions into the service card's Hard limits and the profile's `registry_regions` volatile fact, with the 2026-08-31 source date. Design and Add Capabilities now verify Registry availability independently of the compute runtime; unsupported or unverified Regions cannot complete a recommendation enabling it. A user-confirmed omission also updates Confirm so POC generation cannot restore the earlier selection.

Also corrected the inherited orchestration/multi-agent-only wording: Registry provides catalog/discovery/governance, and Confirm now follows those needs rather than agent count. Updated both plugin copies and the PR description.

Validation: 322 existing Python tests passed per plugin; frontmatter, Markdown lint, formatting, and cross-plugin drift checks passed. Regional behavior was reviewed through the instruction paths; no live Registry deployment was performed. Ready for re-review.

migrate/plugins/migration-to-aws/skills/agent-advisor/references/decision-refs/agentcore.md
leon1418: Fixed in c40f50c4 in both copies. The service cards and Registry Region snapshots match, and the same fixes were applied to Confirm, Design, freshness verification, and Add Capabilities.

Verified all six mirrored edit sets directly, preserving existing intentional differences in the allowlisted files. `cross-plugin-drift.ts` passed, both Agent Advisor suites passed (322 tests each), and frontmatter, Markdown lint, and formatting checks passed.

ayn-builds (approved):

PR #259, opened 2026-09-01
fix(solution-architecture): catch recommendations of sunset services, and test the gate

rsmets:
Addresses both findings from @ayn-builds on #253. Both are fixed. A third problem turned up while testing the first, and is fixed here too.

No RFC: this is a fix to enforcement tooling, not new content. See the new "What does not need an RFC" section below, which writes that rule down.

## Fixed: the sunset check exempted the phrasings a recommendation actually uses

`WARNING_CUE` mixed words describing a service's **status** with words describing a **direction**, and matched line-wide. So `instead of`, `migrat`, and `avoid` counted as warnings wherever they appeared on the line, including when they appeared because the sentence was recommending something. Measured against the gate before this change:

| Line | Before | After |
| --- | --- | --- |
| `Use App Runner instead of ECS for a simple container service.` | exit 0 | flagged |
| `Migrate to App Runner for the fastest path to production.` | exit 0 | flagged |
| `Prefer CodeCommit over GitHub to avoid a third-party dependency.` | exit 0 | flagged |
| `App Runner is a good default.` | flagged | flagged |

Only the bare assertion was caught, which left the check weakest on the exact case it exists for. The removed `aws-dev-toolkit` recommended App Mesh and App Runner, and it recommended them in prose that reads like a recommendation.

**The fix** splits the cues into two lists with different scopes, because the two kinds of word behave differently:

- **Status words** (`deprecat`, `sunset`, `retir`, `no longer`, `closed to new`, `EOL`, `legacy`, `unsupported`) say the service itself is on the way out. Unambiguous wherever they sit, so still matched line-wide.
- **Directional phrases** (`do not`, `avoid`, `instead of`, `rather than`, `migrate away from`, `replace`) only mean "warning" when they come *before* the service name, so they are matched against the preceding text only.

That position rule is what separates `use A instead of B` from `use B instead of A`: same words, opposite meanings, decided by where they sit.

Five cases that must keep passing were verified alongside the four above, and do:

```
App Runner is deprecated, use ECS instead.
Do not use App Runner for new services.
Migrate away from CodeCommit before it is retired.
Avoid CodeCommit; it is closed to new customers.
Replace App Mesh with ECS Service Connect.
```

## Fixed: the gate had no tests, which is why this survived review

A false negative in an enforcement script is invisible by construction. Nothing fails, so nobody looks. `check.test.mjs` adds the cases a reader would assume were already covered, and CI now runs them *before* trusting the gate's own verdict.

Fixtures are written to a temp directory with the gate's `cwd` set there, because fixtures committed inside `solution-architecture/` would be picked up by the gate's own full-tree scan and fail CI.

## Fixed: what does not need an RFC

`CONTRIBUTING.md` opened with "Start with an RFC" and never said where that stops, so a one-line tooling fix looked like it needed a proposal issue. The requirement is about content, because content is what the three criteria judge, and an RFC exists so a scope argument happens before someone writes a skill. A regex fix has no scope argument to have.

The new section says which changes are exempt and gives the test: not how large the change is, but whether a maintainer could reasonably want to redirect it before the work happens.

The advisory reviewer reads `CONTRIBUTING.md` at the pull request head on every run, so it picks this rule up with no change to the reviewer.

## Fixed: fork pull requests now get reviewed

The finding was correct and every specific verified, including the line numbers. The `review` job carried `if: head.repo.full_name == github.repository`, and a skipped job runs none of its steps, `if: always()` included, because step-level `always()` only survives earlier *step* failures inside a job that is already running. So `Signal review finished` could not fire and a fork pull request carried no `agentcore-reviewer` status at all: not a failure, not a skip, **absent**. A reader could not tell "not reviewed" from "nothing to say", and external contributions are the case this folder's gate exists for.

**The remedy suggested first does not work, and that is worth recording.** I implemented the always-running job posting a "skipped, forks unreviewed" status, then reverted it: under `pull_request` a fork's `GITHUB_TOKEN` is read-only regardless of the `permissions:` block, so that job would run and then fail with 403 posting the status. Worse than the silence it was meant to fix.

So the trigger moved to `workflow_run` on Build, which runs in the base repository's context and therefore has credentials for fork pull requests. This is what the shipped reference already called the safe default. It is safe **only** because the runtime fetches the diff itself and nothing is ever checked out; `pull_request_target` grants the same credentials, and checking out a fork head under them is the classic way to leak them.

Three `workflow_run` properties are handled rather than assumed, each commented where it matters:

- No `paths:` support, so the `solution-architecture/**` scope moved into the job.
- `workflow_run.pull_requests` is empty **precisely when the head repo is a fork**, the case this exists to serve, so the number is resolved from the head sha via `repos/{repo}/commits/{sha}/pulls`.
- Only the copy on the default branch ever runs, so this change cannot be exercised by the pull request that makes it. **That includes this one.** Expect no advisory review on this PR; it starts working once merged.

## Fixed: CodeCommit is not sunset any more

Found while writing the tests. CodeCommit closed to new customers on 2024-07-25 and **reopened on 2025-11-25** ([document history](https://docs.aws.amazon.com/codecommit/latest/userguide/history.html)), so flagging it was a false positive, and the cue fix above made it worse by flagging more phrasings.

Worth naming why it rotted: **a closure is announced loudly and a reopening is not**, so an entry in this list decays silently in the direction that blocks correct advice. A check meant to stop bad recommendations had begun stopping a good one. The list now carries an audit date and says to re-verify rather than trust it, and a test asserts CodeCommit is not flagged so the entry cannot quietly come back.

Fixtures that used CodeCommit now use Cloud9, verified still closed to new customers rather than assumed.

## Verification

```
node --test solution-architecture/tools/contribution-gate/check.test.mjs   6 pass, 0 fail
node solution-architecture/tools/contribution-gate/check.mjs              exit 0 on the real tree
markdownlint-cli2 'solution-architecture/**/*.md'                          0 errors
dprint fmt                                                                 clean
both workflow YAML files parse                                             ok
```


solution-architecture/tools/contribution-gate/check.mjs
```diff
@@ -31,15 +31,37 @@ const SUNSET = [
...
+  // makes it stop good ones. Last audited 2026-09-01.
 ];
 
-// A sunset mention is permitted when the surrounding line frames it as a
-// warning, a deprecation note, or a migration away from the service.
-const WARNING_CUE =
-  /\b(deprecat|sunset|end of support|end-of-support|closed to new|do not|don't|avoid|instead of|migrat|no longer|retir|legacy|EOL|rather than|not for new|stop)/i;
+// Words saying the service itself is on the way out. Unambiguous wherever they sit
+// on the line, so they are matched line-wide.
+const SUNSET_STATUS_CUE =
```
ayn-builds: The line-wide half still exempts recommendations, and the words added here widen it rather than narrow it. Ran the gate on fixtures:

| Line | Caught? | Exempted by |
|---|---|---|
| `Use App Runner to deploy the container without managing infrastructure.` | ✅ | — |
| `We removed the old build script, so use App Runner for the container.` | ❌ | `removed` (new) |
| `App Runner is a good fit for legacy workloads you do not want to rewrite.` | ❌ | `legacy` |
| `The retired v1 pipeline is gone; deploy with App Runner instead.` | ❌ | `retir` |
| `Cloud9 gives every developer an identical unsupported-free env.` | ❌ | `unsupported` (new) |
| `Deprecated tooling aside, App Runner is the fastest path to production.` | ❌ | `deprecat` |

The comment says these are "unambiguous wherever they sit on the line." They aren't - `legacy`, `removed`, `retired`, `unsupported`, `deprecat` can describe anything else on the line, and one occurrence exempts every sunset service on it. `unsupported`, `removed` and `retired` are all new here. `CONTRIBUTING.md:199` is a live example: it still self-exempts, now via both `sunset` and `removed`.

I tried the obvious narrowing first - requiring the status word within ±30 chars of the match instead of line-wide. It keeps all six of your warning asserts passing but only fixes 1 of the 7 cases above, because the status words fall inside the window anyway. So this needs a structural change, not a tighter regex.

What did work in a scratch test, 18/20 on the union of your fixtures and mine: detect the recommendation itself rather than its absence - `use|prefer|choose|pick|deploy with|adopt|migrate to` within ~45 chars before the name, or `is/gives/offers … good|best|fastest|default|fit` after it - and keep the existing cue lists purely as an override. The two it still misses are the directional problem I've raised separately on `MOVING_AWAY_CUE` below.

.github/workflows/agentcore-reviewer.yml
```diff
@@ -64,14 +116,63 @@ jobs:
...
+            echo "relevant=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          echo "Resolved pull request #${NUMBER} for ${HEAD_SHA}"
+
+          # Stands in for `paths:`, which workflow_run does not support. Reviewing a pull
+          # request that touches nothing in scope would spend a model call to say nothing
+          # and post a status on changes this folder does not govern.
+          if gh api "repos/${GITHUB_REPOSITORY}/pulls/${NUMBER}/files?per_page=100" \
+            --jq '.[].filename' | grep -q '^solution-architecture/'; then
```
ayn-builds: This runs under `set -euo pipefail`. `grep -q` exits on first match and SIGPIPEs `gh api`; `pipefail` promotes that 141 to the pipeline status, so the `if` takes the `else` branch *despite* a match:

```
$ bash -c 'set -euo pipefail; if seq 1 5 | grep -q "^1$"; then echo MATCHED; else echo MISSED; fi'
MATCHED
$ bash -c 'set -euo pipefail; if seq 1 5000000 | grep -q "^1$"; then echo MATCHED; else echo MISSED; fi'
MISSED
```

It needs output past the pipe buffer, so a large PR or long paths rather than a small one. The consequence is the bad direction: `relevant=false` reaches line 386 and posts a green status whose description positively asserts `No solution-architecture/ changes, nothing to review` on a pull request that did change them. Given the status is `success` in every outcome by design, that description is the only signal a reader has, so it stating something false matters more here than it would elsewhere.

```suggestion
          FILES="$(gh api "repos/${GITHUB_REPOSITORY}/pulls/${NUMBER}/files?per_page=100" \
            --jq '.[].filename')"
          if grep -q '^solution-architecture/' <<<"$FILES"; then
```

ayn-builds (commented):
The positional split is the right idea, and the tests are real: 6/6 pass, and putting fixtures in a `mkdtempSync` dir rather than committing them is correct given the gate's own full-tree scan. The `workflow_run`-without-checkout argument is sound too - privileged context, nothing from the fork is ever checked out, the runtime fetches the diff itself.

Two things inline. The sunset finding isn't closed yet, and there's one shell issue that fails in the green direction.

Also checked and fine as written: the job grants only `id-token: write` + `statuses: write`, and both new `gh api` reads return `200` anonymously on this public repo, so no `pull-requests: read` is needed — worth a comment noting that depends on the repo staying public. And `state=success` in every outcome is deliberate per the comment at line 357, so I'm not raising it.

solution-architecture/tools/contribution-gate/check.mjs
```diff
@@ -31,15 +31,37 @@ const SUNSET = [
...
+//
+// The previous single list mixed both kinds and matched line-wide, which exempted the
+// phrasings a recommendation actually uses. Verified against the gate at the time:
+// "Use App Runner instead of ECS", "Migrate to App Runner", and "Prefer CodeCommit
+// over GitHub to avoid a third-party dependency" all passed, because `instead of`,
+// `migrat`, and `avoid` were treated as warnings wherever they appeared. Only the bare
+// "App Runner is a good default." was caught, leaving the check weakest on exactly the
+// case it exists for: the removed plugin's App Runner and App Mesh recommendations were
+// written as recommendations.
+const MOVING_AWAY_CUE =
```
ayn-builds: The positional check confirms the cue comes *before* the name, but not that the name is the cue's **object**. Where the sunset service is the destination rather than the thing being left, the exemption fires backwards:

```
Instead of Jenkins, use App Runner.          -> not flagged (MOVING: "Instead of")
Migrate off Jenkins and onto App Runner.     -> not flagged (MOVING: "Migrate off")
Replace the Jenkins box with App Runner.     -> not flagged (MOVING: "Replace")
```

`instead of X, use Y` is the most natural phrasing for recommending a replacement, so this is fairly central. The test at line 75 covers the mirror case - a directional phrase *after* the name - but nothing covers a directional phrase before the name pointing at a different service, so these three would make good additions either way.

Fixing it properly means knowing which service the phrase attaches to, which a line-level regex can't do. The cheap 90% is requiring that no other service-shaped token sits between the cue and the match.

solution-architecture/tools/contribution-gate/check.mjs
```diff
@@ -31,15 +31,37 @@ const SUNSET = [
   { pattern: /\bElastic Beanstalk\b/gi, name: "Elastic Beanstalk" },
```
ayn-builds: Not urgent, and it fails in the safe direction - but while the list is being audited, two entries look like the same class of problem the CodeCommit removal just fixed.

Removing CodeCommit is right, and I verified it: the user guide's document history has *"AWS CodeCommit is now available to new customers | November 25, 2025"*, plus the [returns-to-GA post](https://aws.amazon.com/blogs/devops/aws-codecommit-returns-to-general-availability/). The "re-verify, because a reopening isn't announced loudly" note is exactly the right instinct. Applying it to the neighbours:

- **Elastic Beanstalk** isn't sunset. Only individual *platform branches* retire; the [platform support policy](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/platforms-support-policy.html) describes the AL2023 branches as current and fully supported. So this entry blocks correct advice, which is the failure mode your own comment warns about.
- **Kinesis Data Analytics** (line above) is subtler. Only the *for SQL Applications* variant is discontinued - [applications deleted from January 27, 2026](https://docs.aws.amazon.com/kinesisanalytics/latest/dev/discontinuation.html) - while the service itself was **renamed** to Amazon Managed Service for Apache Flink in August 2023. Flagging a bare mention is defensible since it's a stale name at minimum, but `name: "Kinesis Data Analytics"` makes the finding read as "this service is gone," which isn't true. Either narrow the pattern to `for SQL`, or keep it broad with a message that says to use the current name.

rsmets: Part of me wants to remove this deterministic check logic and off load to the agent reviewer... what do you think?

dgeske (dismissed):

.github/workflows/agentcore-reviewer.yml
```diff
@@ -38,19 +38,91 @@ name: AgentCore Reviewer
...
+  # or run pull-request code. Two invariants keep it safe, and both must hold for every
+  # future edit to this file:
+  #
+  #   - No checkout, ever. There is no `actions/checkout` here and there must never be
+  #     one. The runtime reads the diff through the API instead.
+  #   - No fork-controlled text in a shell. The only event values used below are
+  #     `head.sha` and `number`, a hex string and an integer. Never interpolate a title,
+  #     body, branch name, or login into a `run:` block: that is script injection with
+  #     credentials attached.
+  pull_request_target:
```
ayn-builds, line 99: **Blocking.** Switching the fork arm to `pull_request_target` means any GitHub user gets a run, with no approval gate, that holds base-repository credentials: `statuses: write`, the OIDC role assumption, and the AgentCore invocation. The pull request they control supplies the content that run acts on.

Worth flagging that the OIDC `sub` claim is identical for `pull_request` and `pull_request_target`, so a trust policy written with same-repo PRs in mind will accept fork-triggered runs with no change on the AWS side. The comments just above are careful about checkout and about interpolating fork text, which covers the injection path, but not the fact that the run itself is now unapproved and credentialed.

The compensating control is the contribution-gate assert, and it does not hold: see my two comments on `contribution-gate.yml`.

Can we go back to `workflow_run` on Build, which is what the PR description says this change does, or else gate the job on an explicit maintainer-applied label?

ayn-builds (commented):
Reviewed at `fff5bee`. Posting the four items I would want resolved before this merges. The first three are one chain rather than three independent issues: the `pull_request_target` switch opens the exposure, and the assert meant to compensate for it does not hold in two separate ways.

.github/workflows/contribution-gate.yml
```diff
@@ -59,6 +59,41 @@ jobs:
...
+      # The gate is an enforcement script, so a false negative in it is silent: nothing
+      # fails and nobody looks. It shipped a sunset check that passed three of the four
+      # phrasings it existed to catch, and no test would have let that through. Run the
+      # tests before trusting the gate's own verdict below.
+      # The reviewer uses `pull_request_target` so fork pull requests get reviewed without
+      # a maintainer approving a workflow run. That grants base-repository credentials to a
+      # job whose inputs a contributor controls, and it is safe on exactly two invariants.
+      # Both are one careless edit from being lost, and losing either is a credential leak
+      # rather than a broken build, so they are asserted rather than trusted to a comment.
+      - name: Assert the reviewer stays safe under pull_request_target
```
ayn-builds: **Blocking.** This assert is the stated compensating control for the `pull_request_target` switch, but it lives in a workflow gated on `paths: solution-architecture/**`. A pull request that touches only `.github/workflows/agentcore-reviewer.yml`, for example one adding an `actions/checkout` of the PR head, never triggers this workflow at all, so the assert never runs on the change it exists to catch. The new CONTRIBUTING section also makes exactly that kind of PR RFC-exempt, so neither guard covers it.

Second problem: on a fork pull request this greps the fork's copy of `agentcore-reviewer.yml`, not the base copy that actually executes. Those can differ in either direction, so a passing assert here does not tell you anything about the workflow that will run.

Suggest moving the assert into a job with no `paths` filter, and having it read the base ref's copy explicitly rather than the checked-out one.

.github/workflows/contribution-gate.yml
```diff
@@ -59,6 +59,41 @@ jobs:
...
+        run: |
+          set -euo pipefail
+          F=.github/workflows/agentcore-reviewer.yml
+          fail=0
+
+          # 1. Never check out contributor code into a credentialed runner.
+          if grep -qE '^\s*(-\s*)?uses:\s*actions/checkout' "$F"; then
```
ayn-builds: **Blocking.** Both greps in this step are used as `if` conditions, so grep's exit 2 (file missing or unreadable) is indistinguishable from exit 1 (no match). Either way the branch is skipped and `fail` stays 0.

I reproduced this locally: rename or delete `agentcore-reviewer.yml` and this step prints its "invariants hold" message and exits 0 without ever having read the file. So the assert becomes silently vacuous on exactly the edit most likely to break it, a rename or a move.

A precondition closes it:

```bash
[ -f "$F" ] || { echo "::error::$F is missing; the reviewer safety assert cannot run" >&2; exit 1; }
```

solution-architecture/tools/contribution-gate/check.mjs
```diff
@@ -153,13 +180,12 @@ function checkAnyMarkdown(file) {
     let m;
     while ((m = pattern.exec(text)) !== null) {
       const lineNo = lineOf(text, m.index);
-      const line = lines[lineNo - 1] ?? "";
-      if (WARNING_CUE.test(line)) continue; // warned about, not recommended
-      add(
+      note(
```
ayn-builds, line 243: **Blocking**, though not a security issue like the other three.

With `WARNING_CUE` removed, sunset hits are now `note()` calls on the stdout of a step that passes. There is no `::warning` annotation and no step summary entry, so in practice nobody sees them: a green check with notes buried in collapsed log output reads as "clean" to a contributor. The stated fallback is the advisory reviewer, but that never blocks, and it no-ops entirely when it is unconfigured.

Net effect is that criterion 3 is both unenforced and invisible, while `contribution-gate.yml:118` and `solution-architecture/README.md:24` still tell contributors the gate checked it.

If non-blocking really is the intent here, emitting `::warning file=...,line=...` instead would at least surface these in the Files changed view so the claim in the summary stays honest.

.github/workflows/agentcore-reviewer.yml
rsmets: The OIDC point is correct and I verified it: the trust policy in the CDK conditions only on `aud` and `sub`, with `StringLike` on `repo:awslabs/startups:*`. Nothing pins the workflow, so a fork-triggered run is accepted with no AWS-side change. Thanks, that is a real gap.

I am not taking either suggested remedy, because both reintroduce the problem this change exists to fix. `workflow_run` on Build was the previous commit on this branch, and the PR description is stale rather than accurate; I will correct it. It was replaced deliberately: a fork `pull_request` run requires maintainer approval for a first-time contributor, and a `workflow_run` chained off it inherits that wait. A maintainer-applied label is a maintainer action by definition. The requirement is that an outside contribution gets criteria 1 and 2 review with nobody clicking anything, which is exactly what awslabs/startups#278 did not get: `Advisory review` skipped, no `agentcore-reviewer` status at all, on a PR from an outside contributor adding a new skill.

On the exposure itself, I want to be precise about what an attacker gains, because "credentialed" covers a wide range. With no checkout and no fork-controlled interpolation, a fork PR can cause two things: a Bedrock runtime invocation, and a commit status on its own head sha. That is cost and noise. It is not a path to the role's credentials or to writing repository contents. The role grants `bedrock-agentcore:InvokeAgentRuntime` on one runtime ARN and nothing else, and `statuses: write` is the only repository permission.

What I am doing instead, which keeps the no-approval property and closes the gap you found:

- Pin `workflow_ref` in the trust policy to `awslabs/startups/.github/workflows/agentcore-reviewer.yml@refs/heads/main`. Because `pull_request_target` always runs the default branch's copy, a fork cannot alter what that claim contains. This is strictly tighter than the current `sub`-only condition, and it also constrains same-repo runs.
- Add a `concurrency` group keyed on the PR so a push storm cannot fan out invocations.

If you still consider unapproved-and-credentialed unacceptable in principle even with the role scoped to a single `InvokeAgentRuntime` action and the workflow pinned, say so and I will take it to a maintainer decision rather than settle it in the diff. But reverting to `workflow_run` means external contributions keep getting no criteria 1 or 2 review, which is the failure mode that removed the previous plugin from this folder.

.github/workflows/contribution-gate.yml
rsmets: Both halves are correct and I confirmed both. Fixing.

The paths gating is real: `contribution-gate.yml` is `on: pull_request: paths: ['solution-architecture/**']`, so a PR touching only `.github/workflows/agentcore-reviewer.yml` never runs this assert. Worse than you flagged, actually, because the CONTRIBUTING section added in this same PR lists "Fixing the reviewer workflow in `.github/workflows/`" as RFC-exempt. So I built a guard, exempted the exact change class it guards against from the RFC requirement, and left it unable to fire on that class. That is a self-defeating combination and it is on me.

The fork-copy problem is also right, and it is the more interesting one: on a fork PR this workflow checks out the PR head, so the assert reads the fork's `agentcore-reviewer.yml` while the base copy is what `pull_request_target` actually executes. A green assert therefore says nothing about the file that runs. It is not just unhelpful, it is actively misleading, which is worse than absent.

Taking your suggestion as written: moving the assert into its own job with no `paths` filter, and reading the base ref's copy explicitly through the API rather than trusting the checkout.

.github/workflows/contribution-gate.yml
rsmets: Confirmed by reproducing it. With `F` pointing at a nonexistent file:

```
$ grep -qE 'actions/checkout' absent.yml; echo $?
2
```

The `if` is not taken, `fail` stays 0, and the step prints its "invariants hold" message and exits 0 without having read anything. So the assert is silently vacuous on precisely the edit most likely to break it, a rename or a move, which is also the edit least likely to look suspicious in review.

Taking your precondition verbatim. Also switching both checks off `if grep` and onto an explicit exit-code test so a future exit 2 cannot be swallowed the same way, and adding a negative control to the change: I verified locally that both assertions fire when an `actions/checkout` and a `${{ github.event.pull_request.title }}` are injected into a copy of the file, so the guard is known to work in both directions rather than only known to pass.

solution-architecture/tools/contribution-gate/check.mjs
rsmets: Valid, and the honesty half is the part that matters. Verified both claims:

- `contribution-gate.yml:118` still prints "uncaveated sunset-service references and prose style" in its summary.
- `solution-architecture/README.md:24` still tells contributors the three criteria include no references to sunset services.

So the output claims a check that no longer happens. That is manifest drift in the gate's own reporting, which is a category this repo's reviewer is supposed to catch, and it would have read as clean to a contributor.

Non-blocking is the intent, and I will defend that part: three cue-word implementations got recommend-versus-warn wrong in both directions, including exemptions I added myself, so the script reports and does not judge. But you are right that reporting into the stdout of a passing step is indistinguishable from not reporting at all.

Taking your suggestion: emitting `::warning file=...,line=...` so each mention lands in the Files changed view, plus a step-summary section so the count is visible without expanding logs. And correcting both stale claims so the summary describes what the gate now does, which is report the mention and leave the framing to review.

On the fallback being weak, that is fair and I do not want to oversell it. `staleness` is one of only three categories the reviewer can block on, so it is not purely advisory, but it does no-op when unconfigured. The visibility fix is what makes criterion 3 land regardless of whether the reviewer ran.

rsmets:
## Requesting another round of review

Posting as Ray's agent, on his behalf.

@ayn-builds thank you for the four blocking comments, and for the reproductions. Three of the four were defects in code I had written on this branch, and one was a valid diagnosis whose suggested remedy would have undone the change. All four are addressed, and the reasoning for the one I declined is in that thread rather than only in a commit message.

Because the fixes turned into a rewrite rather than a patch, I would rather have more eyes on it than fewer. **All 15 checks are green.**

### What changed since your review

**The safety assert was not merely weak, it was inverted.** Adversarial review wrote sixteen hostile reviewer workflows against my two greps. Fifteen passed. The one it rejected was the safe `env:`-indirection pattern its own error message recommended. `uses: "actions/checkout@v4"` beat the anchored pattern, a folded scalar beat it again, and `${{ format('{0}', github.event.pull_request.title) }}` beat the expression pattern because `[^}]*` cannot cross the `}` inside `{0}`.

It is now an allowlist over the parsed YAML in `tools/reviewer-guard/reviewer_guard.py`, where anything unrecognised fails closed.

**Then the allowlist was broken too**, by a step that reads as an ordinary caching optimisation:

```yaml
run: |
  gh api "repos/${GITHUB_REPOSITORY}/tarball/${HEAD_SHA}" > contribution.tgz
  tar xzf contribution.tgz -C contribution --strip-components=1
  make -C contribution/solution-architecture prepare
```

No `uses:`, no matched pattern, every expression allowlisted. I had left one denylist inside an allowlist guard, and it missed ten of twelve fetch spellings; `git -c … fetch` and `git -C . checkout` defeat `\bgit\s+fetch\b` purely because the words are not adjacent. It also could not see `$GITHUB_HEAD_REF` or `$GITHUB_EVENT_PATH`, which carry fork-controlled text to a shell with no `${{ }}` at all.

**Shell bodies are now pinned by sha256**, so any edit to a script fails until the hash is updated and that hash change is the review artifact. Both rounds' proofs of concept are kept as 32 fixtures and re-proved on every run, because a guard that only ever passes is indistinguishable from a broken one.

Also fixed from your comments specifically: the assert now reads the **head** copy so it can fail the pull request that removes an invariant, not just the base copy; `base.ref` was wrong anyway, since `pull_request_target` loads the **default** branch; the `grep` exit-2 vacuity you reproduced is gone; the gate's sunset notes are `::warning` annotations rather than buried stdout; and the two documents claiming the gate still enforces criterion 3 are corrected.

Separately, the OIDC trust policy conditioned only on `sub`, which is byte-identical for `pull_request` and `pull_request_target`. It now pins `workflow_ref`.

### Where I would most value scrutiny

1. **`pull_request_target` at all.** Ray's requirement is that an outside contribution is reviewed with nobody clicking anything, which `workflow_run` cannot do because it inherits the first-time-contributor approval gate. #278 is the case in the wild: `Advisory review` skipped, no status at all, on a new skill from an outside contributor. If you still consider unapproved-and-credentialed unacceptable with the role scoped to one `InvokeAgentRuntime` action and the workflow pinned, that is a maintainer decision rather than something to settle in the diff.
2. **The hash-pinning tradeoff.** It fails closed on every shell edit. That is the point, and it is also friction. If you think it will get commented out rather than maintained, say so now.
3. **`tools/reviewer-guard/fixtures/`.** These are deliberately hostile workflows. They were under `.github/workflows/` at first and every scanner in CI correctly failed on them, which is why they moved and carry a `.yml.fixture` extension. Worth confirming nothing else picks them up.
4. **What the guard still cannot do.** It runs on `pull_request`, so a pull request editing the guard is checked by the edited copy. `.github/CODEOWNERS` routing `/.github/` to @awslabs/startups-admins is what actually closes that, plus marking the check required. A self-check cannot, because the attacker edits the checker. I have documented that rather than claimed otherwise.

@awslabs/startups-admins flagging you directly, since `/.github/` is yours under CODEOWNERS and this changes what a credentialed job may execute. The RFC-exemption bullet in `solution-architecture/CONTRIBUTING.md` covering workflow fixes is worth a second look too: I narrowed it, but whether a change that alters a credentialed job's execution should ever be RFC-exempt is your call, not mine.

ayn-builds (approved):

hyun0131 (approved):
Checked at `76a8263`: `check.test.mjs` 17/17 pass, the gate exits 0 on the real tree, and `reviewer_guard.py` accepts the real workflow and rejects all 31 fixtures. Sunset mentions are now report-only, the list carries an audit date, and README/CONTRIBUTING match what the gate actually does. The RFC-exemption section is fine with me: CODEOWNERS still forces admin review on everything it exempts.

PR #260, opened 2026-09-01
fix(advisor,migrate): require gcp-to-aws for llm-to-bedrock and agent-advisor plan

herosjourney:
## Problem

`llm-to-bedrock`'s Assess phase and `agent-advisor`'s migration-plan phase both delegate entirely to the `gcp-to-aws` skill via cross-skill invocation. Neither has a standalone implementation of what it delegates:

- `llm-to-bedrock/SKILL.md` (Phase A): "You MUST use the Skill tool to invoke `aws-startup-advisor:gcp-to-aws` (a sibling skill in this same plugin)" — with no check that it's actually installed.
- `agent-advisor`'s migration-plan phase reads `gcp-to-aws`'s phase files directly off disk via `$GCP_BASE = ${CLAUDE_PLUGIN_ROOT}/skills/gcp-to-aws`, with no fallback when `${CLAUDE_PLUGIN_ROOT}` doesn't resolve (it already has a presence check for the standalone-bundle case, but the path resolution itself was incomplete).

This matters because `advisor`'s primary non-Claude-Code install path (`npx skills add ... --skill <name>`) supports installing any single skill on its own — nothing enforces that `gcp-to-aws` comes along with `llm-to-bedrock` or `agent-advisor`. A user who installs just `llm-to-bedrock` gets no warning until they hit Phase A and the Skill tool invocation fails or does something undocumented.

## Fix

**`llm-to-bedrock/SKILL.md`** (advisor + migrate copies): adds Step 0b — a presence check for the `gcp-to-aws` sibling directory before Phase A commits to delegating Assess to it. Missing sibling → stop, tell the user exactly how to install it, do not attempt Assess as a workaround (there is no standalone Assess logic in this skill to fall back to). Frontmatter description and the opening section now state the hard dependency explicitly instead of reading as end-to-end and self-contained.

Also documents a related, separate risk: the Skill-tool cross-skill invocation mechanism itself is confirmed for Claude Code but unverified on every agent `npx skills` supports. If that invocation fails even with `gcp-to-aws` present, the doc now describes a manual two-step handoff (run `gcp-to-aws` yourself, then come back once its Generate phase is `completed`) rather than leaving the failure mode undefined.

**`agent-advisor/.../migration-plan.md`** (advisor + migrate copies): adds a relative-path fallback for $GCP_BASE when ${CLAUDE_PLUGIN_ROOT} doesn't resolve, mirroring the fallback `agent-advisor`'s own SKILL.md already uses for $PLUGIN. Also fixes an ordering bug — "Path definitions" now comes before the "Step -1" presence check that depends on it (previously the doc told the reader to resolve a variable defined further down). `agent-advisor` is lower-severity than `llm-to-bedrock` here — the migration-plan step already degrades gracefully to `not_applicable` when the engine is absent; runtime recommendations and design-backed POCs are unaffected.

**`advisor/README.md`, `advisor/AGENTS.md`, `setup.md`**: state the `gcp-to-aws` dependency in each skill's own blurb (not only in an install-appendix footnote) — hard requirement for `llm-to-bedrock`, soft/graceful-degradation note for `agent-advisor`.

## Scope

Documentation + presence-check logic only. No change to the DSL, `_exec` dispatch, MCP config, or any skill's core migration logic. Both plugin copies (`advisor/` and `migrate/`) were updated in lockstep — these two files were previously on the cross-plugin drift tool's allowlist (an intentional-divergence list), which meant a fix landing on only one side would go undetected by CI. Verified both copies stay in sync per the tool's existing normalization rules.

## Verification

- `markdownlint-cli2` (repo's own `.markdownlint-cli2.yaml`) clean on all seven changed files.
- `dprint check` clean on all seven changed files.
- `node advisor/plugins/aws-startup-advisor/tools/cross-plugin-drift.ts` → `OK (270 identical, 25 allowlisted across 6 skill trees + agents/)` — confirms the advisor and migrate copies of the two touched SKILL.md/migration-plan.md files remain equivalent after this change.


leon1418 (commented):
[🤖 AI review 🤖]

**Reviewed PR #260** at head `767b4b73` — 7 files, +188/−31. Documentation + presence-check logic for `llm-to-bedrock` ↔ `gcp-to-aws` hard dependency and `agent-advisor` ↔ `gcp-to-aws` soft dependency.

## Change summary

1. **`llm-to-bedrock/SKILL.md`** (advisor + migrate): Adds Step 0b — a sibling-directory presence check for `gcp-to-aws` that fails fast with install instructions if missing. Frontmatter description now states the hard dependency explicitly. Separately documents the Skill-tool cross-agent invocation risk with a manual two-step workaround.

2. **`migration-plan.md`** (advisor + migrate): Moves "Path definitions" section **above** "Step -1" (fixing the ordering bug where Step -1 referenced `$GCP_BASE` before it was defined). Adds a relative-path fallback when `${CLAUDE_PLUGIN_ROOT}` does not resolve. Widens the "absent" scenario from ATX-only to include partial `npx skills add` / local-symlink installs.

3. **`advisor/README.md`, `advisor/AGENTS.md`, `advisor/setup.md`**: Surface the cross-skill dependencies in each skill's own blurb with install commands, not just in footnotes.

## Findings

**0 mandatory blockers. 0 Nits.**

### Design

The design is sound. The two dependency modes are correctly differentiated:
- `llm-to-bedrock` has a **hard** dependency — no standalone Assess exists, so the skill must stop and instruct the user to install `gcp-to-aws`. This is the right call; a silent no-op or a half-baked inline Assess would be worse.
- `agent-advisor` has a **soft** dependency — migration-plan degrades gracefully to `not_applicable` while runtime recommendations and POCs continue working. Correct — forcing the user to install `gcp-to-aws` when they only want a runtime recommendation would be wrong.

The ordering fix (Path definitions before Step -1) eliminates a real bug where an agent following the doc linearly would hit an undefined `$GCP_BASE` reference. The relative-path fallback for `${CLAUDE_PLUGIN_ROOT}` is a pragmatic fix for real-world install layouts.

### Cross-plugin parity

Both cross-plugin skill pairs (`llm-to-bedrock/SKILL.md`, `migration-plan.md`) have **intentional, context-appropriate differences** — install instructions reference the correct plugin name and install mechanism for each copy (`npx skills add` for advisor, `/plugin install migration-to-aws@startups` for migrate). The semantic content (presence check logic, fallback behavior, path resolution formula, error messaging structure) is identical across both copies. The PR body confirms the drift tool passed: `OK (270 identical, 25 allowlisted across 6 skill trees + agents/)`.

The three advisor-only files (`AGENTS.md`, `README.md`, `setup.md`) have no counterparts in `migrate/` — `migrate/AGENTS.md` and `migrate/.../setup.md` do not exist, and `migrate/README.md` does not mention `llm-to-bedrock`. No cross-plugin sync violation.

### Functionality

The `<SKILL_BASE>/../gcp-to-aws/SKILL.md` existence check is correct — it tests for the SKILL.md file (not just the directory), which confirms the skill is actually installed rather than an empty or partial directory. The sibling-relative path resolution avoids any dependency on `${CLAUDE_PLUGIN_ROOT}`.

### Documentation

README install instructions are clear and the "Cross-skill dependencies" section is well-placed after the single-skill examples. The "this only matters if you're installing a subset" clarification at the end correctly scopes the section.

## Validation evidence

- Diff: all 7 files reviewed line-by-line
- Cross-plugin parity: both skill pairs fetched at PR head and diffed; differences are install-path-specific language only
- Ordering fix: confirmed `## Path definitions` now precedes `## Step -1` in both copies (was reversed on main)
- advisor-only files: verified `migrate/AGENTS.md` (404), `migrate/.../setup.md` (404), `migrate/README.md` (no llm-to-bedrock mention) — no sync needed
- CI: **0 check runs reported** on this branch
- Reviews: 0 prior reviews
- Merge status: MERGEABLE
- Approvals: REVIEW_REQUIRED (0 approvals)

**Verdict: CLEAN. Merge-ready pending CI green and maintainer approval.**

leon1418 (approved):

PR #262, opened 2026-09-02
fix(pricing): verify Anthropic batch rows; retract unverified Codex GA claim

herosjourney:
## Problem

A routine audit of the Anthropic batch & prompt-cache table (`pricing-cache.md`) and the OpenAI-on-Bedrock reference (`openai-on-bedrock.md`) surfaced two pricing facts that were recorded without a citation to a model-support table, so nothing would catch them drifting from reality:

1. **Unverified batch rows presented as fact.** The batch table lists rates for Fable 5, Sonnet 5, and Opus 4.8 with no distinction from the rest of the table. None of those three appear on the official [batch-inference-supported-models table](https://docs.aws.amazon.com/bedrock/latest/userguide/batch-inference-supported.html) — their rows were a 50%-of-on-demand projection, not a confirmed SKU. An estimate that builds a batch-discount TCO on one of these three models would be citing a rate that may not exist.

2. **A fabricated GA claim for Codex on Bedrock.** `openai-on-bedrock.md` stated "Codex on Bedrock is GA with pay-per-token pricing." Codex does not appear on the [OpenAI model card index](https://docs.aws.amazon.com/bedrock/latest/userguide/model-cards-openai.html), and no Codex rate exists anywhere in this plugin — not on the pricing page, not in `pricing-cache.md`, not in `bedrock_pricing.py`'s static table. Any coding-agent workload routed through this claim would hit a pricing dead end with no guidance on what to do instead.

## Solution

**1. Verify the batch table row-by-row and mark what isn't confirmed (`pricing-cache.md`).**

- Checked all nine rows against the batch-inference-supported-models table. Opus 4.6, Sonnet 4.6, Opus 4.5, Sonnet 4.5, and Haiku 4.5 are listed there (via CRIS profiles), and their cached batch rates match the pricing page's 50%-of-on-demand discount exactly — confirmed correct.
- Fable 5, Sonnet 5, and Opus 4.8 are not on that table. Marked their batch cells `†` with a footnote: unverified, do not build a batch-discount TCO on them, price on-demand instead.
- Added a note against relying on the AWS Pricing Calculator to cross-check this table — its Bedrock batch/cache coverage lags the docs and pricing page for several of these models, so it isn't a reliable second source here. The batch-inference-supported-models table and pricing page remain the sources of truth.

**2. Retract the Codex claim (`openai-on-bedrock.md`).**

Rewrote the paragraph: re-check the model card index on refresh; if Codex is still absent, route coding-agent workloads to a priced cross-family target (GPT-5.6 Sol / Terra fit the coding-agent tier) or report `pricing_source: "unverified"` — never attach a fabricated rate to an estimate.

Both files mirrored to `advisor/plugins/aws-startup-advisor/` per the drift-gate convention (both copies were byte-identical before this change).

## Verification

- `mise run drift:check` — OK (advisor vendored copies synced, 270 identical / 25 allowlisted)
- `mise run shared:check`, `mise run lint:model-ids` — OK
- `mise run lint:md`, `dprint check` — clean on changed files
- `mise run pricing:staleness` — pre-existing warnings only (global CRIS GPT-5.6 rows, untouched here)
- Batch support verified against the AWS docs batch-inference-supported-models table (2026-09-02), not estimated

## Not in scope

- The pre-existing `pricing:staleness` drift warnings on the `global.openai.gpt-5.6-*` `STATIC_FALLBACK` rows vs. the cache rate card
- `Claude Opus 5` appears on the docs batch table but isn't in the pricing cache at all yet — separate refresh

## Type of Change

- [x] Bug fix
- [ ] New plugin/power/tool
- [ ] Documentation update
- [ ] Guardrail/CI update


advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/references/shared/pricing-cache.md
```diff
@@ -458,18 +458,32 @@ Image editing services (inpaint, erase, upscale, etc.) are priced at $0.03–$0.
...
+> **Do not cross-check these rows against the AWS Pricing Calculator** — as of Sep 2026 the calculator carries no
+> batch metered IDs for Opus 4.6 / Sonnet 4.6 and collapses other batch/cache SKUs (known calculator defect, not a
+> pricing-page signal). The pricing page and the docs table above are the sources of truth.
+
 | Model                    | Batch in | Batch out | 5m cache write | 1h cache write | Cache read |
 | ------------------------ | -------- | --------- | -------------- | -------------- | ---------- |
-| Claude Fable 5           | 5.00     | 25.00     | 12.50          | 20.00          | 1.00       |
-| Claude Sonnet 5          | 1.00     | 5.00      | 2.50           | 4.00           | 0.20       |
-| Claude Opus 4.8          | 2.50     | 12.50     | 6.25           | 10.00          | 0.50       |
+| Claude Fable 5           | 5.00 †   | 25.00 †   | 12.50          | 20.00          | 1.00       |
```
leon1418: Some new models not in this table, e.g. opus-5, fable-5-1, etc. Do we need to add them, or do we still want to maintain this cache table locally?

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/shared/openai-on-bedrock.md
```diff
@@ -240,8 +240,14 @@ days for automated abuse detection; retained inputs/outputs are stored and proce
 OpenAI unless the customer opts in. Prompts and completions are not used to train models. Calls run under the
 customer's IAM policies, inside their VPC, logged to CloudTrail, and in-region inference keeps data in-region.
 
-**Codex on Bedrock is GA** with pay-per-token pricing, inference through Bedrock, and usage counting toward AWS
-commitments — relevant when the source workload is a coding agent.
+**Codex on Bedrock: unverified — do not price it.** An earlier revision of this file stated Codex is GA on Bedrock
```
ayn-builds, line 244: Good catch retracting this, and I like that the replacement says what to do instead of just deleting the claim. One gap: the same claim is still live one file over.

`references/design-refs/ai-openai-to-bedrock.md:257`:

```
| Codex | Codex on Bedrock (GA) | Pay-per-token, inference through Bedrock, counts toward AWS commitments |
```

That's the "GA" plus "pay-per-token" pair this paragraph now walks back, and `design-refs/index.md:59` routes OpenAI workloads straight into that file at Design, so an agent can read the old claim and never reach the retraction here. Lines 13-15 of this file make it explicit that shouldn't happen:

> **This file is the single source of truth for OpenAI-on-Bedrock facts in this plugin.** `ai-openai-to-bedrock.md` (mapping policy), `design-ai.md` (selection), `estimate-ai.md` (costing), and `ai-migration-guardrails.md` (quota risk) all defer to it. Do not restate model IDs, regions, or endpoint paths elsewhere — link here.

Line 257 is that restatement. I swept both plugin copies and it's the only other substantive Codex claim in the repo, so it's a one-row fix:

```
| Codex | Not verified on Bedrock - see `shared/openai-on-bedrock.md` | Do not price; re-check the model card index first |
```

Also worth adding Codex to the "Not on Bedrock" list at `:41-43`. Right now it's in neither the catalog table nor that list, so "treat it as a model not in the catalog above" takes a bit of inference to land.

advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/references/shared/openai-on-bedrock.md
```diff
@@ -240,8 +240,14 @@ days for automated abuse detection; retained inputs/outputs are stored and proce
 OpenAI unless the customer opts in. Prompts and completions are not used to train models. Calls run under the
 customer's IAM policies, inside their VPC, logged to CloudTrail, and in-region inference keeps data in-region.
 
-**Codex on Bedrock is GA** with pay-per-token pricing, inference through Bedrock, and usage counting toward AWS
-commitments — relevant when the source workload is a coding agent.
+**Codex on Bedrock: unverified — do not price it.** An earlier revision of this file stated Codex is GA on Bedrock
```
ayn-builds, line 244: Same note as on the `migrate/` copy: the "GA, pay-per-token" claim survives at `design-refs/ai-openai-to-bedrock.md:257`, which Design reaches via `design-refs/index.md:59` and which `:13-15` here says defers to this file. Codex is also still missing from the "Not on Bedrock" list at `:41`. Byte-identical copies, so both sides need the same fix.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/shared/pricing-cache.md
```diff
@@ -458,18 +458,32 @@ Image editing services (inpaint, erase, upscale, etc.) are priced at $0.03–$0.
...
+| Claude Sonnet 5          | 1.00 †   | 5.00 †    | 2.50           | 4.00           | 0.20       |
+| Claude Opus 4.8          | 2.50 †   | 12.50 †   | 6.25           | 10.00          | 0.50       |
 | Claude Sonnet 4.6 (+ LC) | 1.50     | 7.50      | 3.75           | 6.00           | 0.30       |
 | Claude Opus 4.6 (+ LC)   | 2.50     | 12.50     | 6.25           | 10.00          | 0.50       |
 | Claude Opus 4.5          | 2.50     | 12.50     | 6.25           | 10.00          | 0.50       |
 | Claude Haiku 4.5         | 0.50     | 2.50      | 1.25           | 2.00           | 0.10       |
 | Claude Sonnet 4.5        | 1.50     | 7.50      | 3.75           | 6.00           | 0.30       |
 | Claude Sonnet 4.5 — LC   | 3.00     | 11.25     | 7.50           | 12.00          | 0.60       |
 
+† Not listed on the batch-supported models table as of 2026-09-02 — rate is the standard 50%-of-on-demand
```
ayn-builds: The footnote asks for the right thing but nothing downstream can act on it.

"Treat as `unverified` in estimate output" has no consumer. Both places that apply the discount are unchanged:

- `phases/design/design-ai.md:189` - `` | `ai_latency = "flexible"` | Any model; flag Batch API for 50% savings | ``
- `phases/estimate/estimate-ai.md:140` - `| Batch API | 50% | Non-real-time workloads |`

"Any model" and a flat 50% with no carve-out, so a Fable 5 or Opus 4.8 estimate still produces a clean 50% batch saving with no unverified marker anywhere the customer sees. I grepped both files for any other batch qualifier and there isn't one.

There is a gate that does roughly this, but it won't pick these rows up as written. `estimate-ai.md:126` says to "treat rows marked `unverified` there as blocking for any quoted figure — resolve them from the Bedrock pricing page first" - it keys on the literal word, and it's the marker this file already uses at `:845-846`:

```
| GPT-5.6 Sol   | _unverified_ | _unverified_ | 1M      | frontier  |
```

But that paragraph is under "Pricing source caveat for OpenAI models," so it's scoped to the GPT rows and wouldn't cover Anthropic batch cells even if they said `_unverified_`. So I think this needs two edits rather than one:

- Use `_unverified_` in the batch cells (or `5.00 (_unverified_)` if you want to keep the projection visible), matching the vocabulary already in the file
- Add the model qualifier to the two batch-discount rows, e.g. `| Batch API | 50% | Non-real-time workloads - confirmed-batch models only; see pricing-cache.md |`, and widen the `unverified` gate at `estimate-ai.md:126` beyond OpenAI so it covers any unverified row

nit: `†` is already taken nearby. `design-refs/ai-anthropic-to-bedrock.md:18` uses it for "Claude Sonnet 5 intro pricing through Aug 31, 2026; then $3 / $15." Same glyph, same model, different meaning, two files in the same skill - and per the note on the Sonnet 5 row, someone reconciling the two is going to have a confusing minute.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/shared/pricing-cache.md
```diff
@@ -458,18 +458,32 @@ Image editing services (inpaint, erase, upscale, etc.) are priced at $0.03–$0.
...
+> batch metered IDs for Opus 4.6 / Sonnet 4.6 and collapses other batch/cache SKUs (known calculator defect, not a
+> pricing-page signal). The pricing page and the docs table above are the sources of truth.
+
 | Model                    | Batch in | Batch out | 5m cache write | 1h cache write | Cache read |
 | ------------------------ | -------- | --------- | -------------- | -------------- | ---------- |
-| Claude Fable 5           | 5.00     | 25.00     | 12.50          | 20.00          | 1.00       |
-| Claude Sonnet 5          | 1.00     | 5.00      | 2.50           | 4.00           | 0.20       |
-| Claude Opus 4.8          | 2.50     | 12.50     | 6.25           | 10.00          | 0.50       |
+| Claude Fable 5           | 5.00 †   | 25.00 †   | 12.50          | 20.00          | 1.00       |
+| Claude Sonnet 5          | 1.00 †   | 5.00 †    | 2.50           | 4.00           | 0.20       |
```
ayn-builds: Not something this PR broke, but it's now inside the scope of what the PR vouches for, so raising it here.

These cells derive from the intro rate, which expired two days ago. `:395` has Sonnet 5 at `2.00 / 10.00` with status `active (intro $2/$10 thru Aug 31, 2026; then $3/$15)`, and `:386` says the same in prose. `1.00 / 5.00` batch and `2.50 / 4.00 / 0.20` cache are exactly 50% / 1.25x / 2x / 0.1x of $2/$10. At the current $3/$15 the row should be:

```
| Claude Sonnet 5          | 1.50 †   | 7.50 †    | 3.75           | 6.00           | 0.30       |
```

Same numbers as the Sonnet 4.6 row at `:477`, which is what you'd expect now that the two share on-demand pricing. The on-demand row at `:395` needs the update too.

Fair pushback: `Last updated: 2026-08-24` predates the rollover, so the row was correct when the cache was refreshed. Two reasons I'd still fold it into this PR rather than leave it:

- The staleness mechanism can't catch this one. The 30-day rule at `:9` doesn't fire until 2026-09-23, three weeks after the rate changed, and this isn't unknowable drift - the file states the exact rollover date itself.
- The new footnote and banner are dated 2026-09-02 and describe these specific cells as "the standard 50%-of-on-demand projection" with "Prompt-cache columns are unaffected." True for Fable 5 and Opus 4.8. For Sonnet 5 the cache columns *are* affected, by the rollover rather than by batch, so the PR ends up certifying numbers that aren't right.

"Not in scope" carves out the gpt-5.6 warnings and Opus 5 but doesn't mention the Sonnet 5 rollover, so I assume this was missed rather than deferred.

ayn-builds (commented):
Both changed files are byte-identical across `advisor/` and `migrate/` at `ee781f7`, so no drift.

The substance here is right and I'd like it to land. Retracting the Codex claim with instructions for what to do instead beats deleting it, and the batch banner is the right shape - `:386` already carried an ad hoc version of this exact warning ("confirm **batch** availability per model (Opus 4.7 batch was **not** listed on the global cross-region table when this row was added)"), so promoting it to a marker plus footnote is a real improvement. The calculator caveat is a genuinely useful thing to have written down.

Inline notes. The first two are both "the fix doesn't reach where it needs to":

1. The Codex retraction doesn't reach `design-refs/ai-openai-to-bedrock.md:257`, which still says "Codex on Bedrock (GA)" with pay-per-token. Design routes there directly, and `openai-on-bedrock.md:13-15` explicitly says that file defers to this one.
2. The `†` marker has no consumer. `design-ai.md:189` and `estimate-ai.md:140` still apply 50% to "any model," and the `unverified` gate at `estimate-ai.md:126` is scoped to OpenAI rows, so it won't cover these even if the cells are relabelled.
3. Adjacent rather than yours: Sonnet 5's batch and cache cells derive from the $2/$10 intro rate that ended Aug 31. The cache's `Last updated: 2026-08-24` predates that, so the row was fine when written - but the new banner is dated 2026-09-02 and now certifies those cells as correct, and the 30-day staleness rule won't fire until Sep 23.

Plus two nits (the ninth table row isn't named in the banner, and `†` already means something else at `ai-anthropic-to-bedrock.md:18`) and one question about whether the CRIS premium affects the confirmed batch rates.

All line edits in files this PR already touches, except the design-ref row and the two batch-discount tables.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/shared/pricing-cache.md
```diff
@@ -458,18 +458,32 @@ Image editing services (inpaint, erase, upscale, etc.) are priced at $0.03–$0.
...
+| Claude Opus 4.8          | 2.50 †   | 12.50 †   | 6.25           | 10.00          | 0.50       |
 | Claude Sonnet 4.6 (+ LC) | 1.50     | 7.50      | 3.75           | 6.00           | 0.30       |
 | Claude Opus 4.6 (+ LC)   | 2.50     | 12.50     | 6.25           | 10.00          | 0.50       |
 | Claude Opus 4.5          | 2.50     | 12.50     | 6.25           | 10.00          | 0.50       |
 | Claude Haiku 4.5         | 0.50     | 2.50      | 1.25           | 2.00           | 0.10       |
 | Claude Sonnet 4.5        | 1.50     | 7.50      | 3.75           | 6.00           | 0.30       |
 | Claude Sonnet 4.5 — LC   | 3.00     | 11.25     | 7.50           | 12.00          | 0.60       |
```
ayn-builds, line 486: nit: the banner names 8 models but the table has 9 rows, and this is the one left out. Sonnet 4.6 and Opus 4.6 fold their long-context variants in as `(+ LC)` so those are covered by name, but Sonnet 4.5 LC is a separate row at a separate rate (`3.00 / 11.25`, 50% of the `6.00 / 22.50` on-demand at `:403`) and isn't listed as confirmed or unverified. The description says all nine rows were checked, so I'd guess it's confirmed and just missing from the list.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/shared/pricing-cache.md
```diff
@@ -458,18 +458,32 @@ Image editing services (inpaint, erase, upscale, etc.) are priced at $0.03–$0.
 
 Per 1M tokens unless noted. See [Bedrock pricing](https://aws.amazon.com/bedrock/pricing/) for full regional and tier tables.
 
+> **Batch support verified 2026-09-02** against the
+> [supported Regions and models for batch inference](https://docs.aws.amazon.com/bedrock/latest/userguide/batch-inference-supported.html)
+> table: **Opus 4.6, Sonnet 4.6, Opus 4.5, Sonnet 4.5, and Haiku 4.5 are listed** (via cross-region inference
+> profiles; batch is 50% of on-demand per the pricing page, matching the rows below). **Fable 5, Sonnet 5, and
```
ayn-builds: Question rather than a finding, since I can't check the pricing page from here: the banner says these five are listed "via cross-region inference profiles," and `:386` notes CRIS runs above N. Virginia ("US East (Ohio) cross-region inference for Claude Sonnet 4.6 is listed at $3.30 / $16.50 per 1M input/output (≈10% above N. Virginia)"). If batch is only reachable through a CRIS profile, is the batch rate 50% of the N. Virginia price, which is what these rows show, or 50% of the CRIS price? If the latter, the confirmed rows are ~10% low. Either way a clause in the banner would save the next person re-deriving it.

advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/references/shared/pricing-cache.md
```diff
@@ -458,18 +458,32 @@ Image editing services (inpaint, erase, upscale, etc.) are priced at $0.03–$0.
...
+| Claude Sonnet 5          | 1.00 †   | 5.00 †    | 2.50           | 4.00           | 0.20       |
```
ayn-builds: Same notes as on the `migrate/` copy: the `†` marker doesn't reach `design-ai.md:189` or `estimate-ai.md:140`, and the `unverified` gate at `estimate-ai.md:126` is OpenAI-scoped so it won't pick up Anthropic batch rows; and Sonnet 5's cells derive from the $2/$10 intro rate that expired Aug 31, so they should be `1.50 / 7.50` and `3.75 / 6.00 / 0.30`. Byte-identical copies, same fix both sides.

herosjourney:
Thanks @ayn-builds — all addressed in `08b521b`.

**1. Codex retraction reaches the design-ref.** `design-refs/ai-openai-to-bedrock.md` row is now `Not verified on Bedrock — see shared/openai-on-bedrock.md / Do not price; re-check the OpenAI model card index first`, and Codex is added to the **Not on Bedrock** list in `openai-on-bedrock.md`. Design no longer has a path to the stale GA claim.

**2. The batch marker now has a consumer.**
- Switched the glyph to `‡` (the `†` collision with `ai-anthropic-to-bedrock.md:18` is gone).
- Footnote now uses the `_unverified_` vocabulary the gate keys on.
- Added the model qualifier to both batch-discount rows (`design-ai.md`, `estimate-ai.md`): confirmed-batch models only.
- Widened the `unverified` gate in `estimate-ai.md` beyond OpenAI so it covers any `_unverified_` row, including the Anthropic batch cells.

**3. Sonnet 5 rollover.** The intro $2/$10 ended Aug 31, so the on-demand row, the prose, and the batch/cache cells now use the $3/$15-derived values (`1.50 / 7.50` batch, `3.75 / 6.00 / 0.30` cache — matching Sonnet 4.6). You were right it was missed, not deferred.

**Nits.** Added Sonnet 4.5 — LC (the 9th row) to the confirmed list in the banner.

**CRIS question.** Good catch — rather than re-derive, I added a clause to the banner: the confirmed rows are 50% of the **N. Virginia** on-demand price; where batch is only reachable via a CRIS profile listing above N. Virginia (e.g. Sonnet 4.6 in Ohio, ≈10% higher), the effective batch rate is ≈10% above these cells, and you should re-derive from the profile's on-demand rate for a region-specific quote.

Verification: `mise run build` lint/frontmatter/drift all OK, `drift:check` 270 identical / 25 allowlisted (no new suppressions), markdownlint 0 errors, dprint clean. All 5 files remain byte-identical across `advisor/` and `migrate/`.

herosjourney:
**Conflicts resolved at `cf5589f`** (merge of `origin/main` into this branch).

The conflict in both copies of `pricing-cache.md` was factual, not textual. This branch assumed Claude Sonnet 5's scheduled Sep 1 increase to $3/$15 took effect and derived its batch/cache row from that. `main` (ee92c83, Sep 3) records that Anthropic made the $2/$10 launch rate the standard price and cancelled the increase. Checked against the primary source before choosing: Anthropic's pricing page says the $2/$10 rate "is now the standard price" and the $3/$15 increase "will not occur".

Resolution, identical in both plugin copies:

- Took `main`'s Sonnet 5 on-demand narrative and $2/$10 quick-reference row.
- Kept this branch's batch-verification work unchanged (2026-09-02 docs-table check, `‡` markers, CRIS and Pricing Calculator caveats, `_unverified_` gate footnote).
- Re-derived the Sonnet 5 batch/cache row from $2/$10 using the same multipliers as every other row (50% / 1.25× / 2× / 0.1×): batch **1.00 ‡ / 5.00 ‡**, cache write **2.50 / 4.00**, cache read **0.20**. The pre-merge 1.50 / 7.50 / 3.75 / 6.00 / 0.30 would have overstated Sonnet 5 batch and cache by 50%.

Verified after the merge: twins byte-identical; `mise run lint` exit 0; `drift:check` 275 identical / 25 allowlisted (unchanged); `shared:check` OK; node suites pass; `test_bedrock_pricing.py` 21/21 (`uv run --group dev pytest`); sweep for any remaining "$3/$15 Sonnet 5" claim across both trees returned nothing.

Note for reviewers: the PR description still calls the unverified marker `†`; the files use `‡`.

ayn-builds (approved):

PR #267, opened 2026-09-03
fix(gcp-to-aws): record Fable 5.1 pricing and enforce restricted-model selection

leon1418:
The Fable 5.1 launch introduced a new Bedrock model with access and regional restrictions that must be reflected in both pricing data and model selection. This PR records verified rates, excludes restricted frontier models from default recommendations, and keeps those restrictions intact when the cache is refreshed.

## Changes

- Add Claude Fable 5.1 (`us.anthropic.claude-fable-5-1` for US Geo inference) with 1M context, $10/$50 per million input/output tokens, cache writes of $12.50/$20 for 5m/1h, and $0.25 cache reads. Both Fable 5 and 5.1 have `N/A` Bedrock batch rates.
- Mark Fable 5 and 5.1 as `restricted`: both require the `aws_review` data-retention opt-in. Exclude restricted models from `recommended_model`, `backup_model`, and mapping defaults; describe them only on explicit request with their access requirements. Keep Mythos out of the recommendation table because it is a gated preview, not unlocked by an account-level opt-in alone.
- Apply these selection rules across source-provider guides, widen the `_unverified_` pricing gate to all providers, and preserve `restricted` during lifecycle-only cache refreshes.
- Distinguish endpoint-specific regional support. Both Fable models use `us.`/`global.` inference profiles on `bedrock-runtime`; Fable 5 also has in-region access through `bedrock-mantle` in us-east-1. Neither has an `eu.` Geo profile, so the existing EU filter remains. Generic disclosures now name any applicable endpoint-specific regional or CRIS-only restrictions.
- Preserve the batch-availability safeguards from main: Sonnet 5 and Opus 4.8 batch rates remain marked unverified (‡), while Fable 5 / 5.1 retain `N/A`. Keep the estimate guide's unverified examples consistent with those rows. The Sonnet 5 pricing and comparison corrections from merged #269 are included through main.

All six changed file pairs are synchronized between the advisor and migrate plugins. #268 is superseded by this PR.

## Sources

- [Fable 5.1 AWS launch, September 1, 2026](https://aws.amazon.com/about-aws/whats-new/2026/09/claude-fable-5-1-aws/)
- [Fable 5.1 GovCloud launch](https://aws.amazon.com/about-aws/whats-new/2026/09/claude-fable-5-1-aws-govcloud/)
- [Fable 5 model card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-fable-5.html) and [Fable 5.1 model card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-fable-5-1.html): endpoint and regional support checked September 9, 2026.
- [Amazon Bedrock pricing](https://aws.amazon.com/bedrock/pricing/): Fable rates and batch availability checked during the September 3–4 revisions.
- [Anthropic pricing](https://platform.claude.com/docs/en/about-claude/pricing): Sonnet 5's permanent $2/$10 rate checked September 9, 2026.

## Validation

- Cross-plugin drift check: passed (275 identical files, 25 allowlisted differences).
- Repository-wide `dprint check --incremental=false`: passed.
- Node test suites: 149 passed, 0 failed, using Node 24.14.1 and Terraform 1.13.5.
- Vendored shared-reference checks: passed for both plugins.
- `git diff --check`: passed.

This branch includes main at `57708ee0`, including merged #269. The September 16 merge resolves the two pricing-cache conflicts, preserving the Fable/restricted-model additions and upstream batch-price warnings. Both estimate-guide copies also distinguish the remaining unverified rates from Fable's `N/A` rates.


vagmur-aws (dismissed):

shrugninreddy-hash (dismissed):

advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/references/shared/pricing-cache.md
```diff
@@ -392,6 +392,8 @@ See `shared/ai-model-lifecycle.md` for lifecycle details. **Do not recommend Leg
 | Model                            | Model ID                                 | Provider  | Input $/1M | Output $/1M | Context | Tier      | Status                                                       |
 | -------------------------------- | ---------------------------------------- | --------- | ---------- | ----------- | ------- | --------- | ------------------------------------------------------------ |
 | Claude Fable 5                   | anthropic.claude-fable-5                 | Anthropic | 10.00      | 50.00       | 1M      | frontier  | active                                                       |
+| Claude Fable 5.1                 | anthropic.claude-fable-5-1               | Anthropic | TBD        | TBD         | TBD     | frontier  | active (GA on AWS)                                           |
```
ayn-builds: `TBD` isn't a marker anything downstream recognizes. This file's convention for an unknown price is `_unverified_` (lines 833-834, and `openai-on-bedrock.md:256` says "resolve any row still marked _unverified_").

Worth flagging that swapping in `_unverified_` isn't sufficient on its own here. The rule that makes `unverified` blocking is `phases/estimate/estimate-ai.md:126`, and it's scoped by its own heading to OpenAI models - so an Anthropic row carrying `_unverified_` still isn't gated. `pricing-fallback.md:8-11` only fires on whole-cache staleness AND an MCP failure, neither of which is true here.

Net effect today: a frontier row with no price passes every guard and can reach the ROI table. Two things needed - use `_unverified_` for consistency, and widen the :126 caveat so it's not provider-specific.

Applies identically to the `migrate/plugins/migration-to-aws/...` mirror at the same line.

ayn-builds: The "don't default to this" guardrails all name **Claude Fable 5** literally, and none were extended to 5.1:

- `pricing-cache.md:386` - "Do not default to Claude Fable 5 (frontier)."
- `design-refs/ai-anthropic-to-bedrock.md:22` - "Do **not** default to Claude Fable 5 (frontier / Mythos-class pricing — opt-in only)."
- `design-refs/ai-openai-to-bedrock.md:14` - "Do not default to Claude Fable 5."
- `design-refs/ai-gemini-to-bedrock.md:11` - same sentence.

Separately, `agent-advisor/references/models/anthropic-bedrock-2026-07-21.json:7` excludes "Claude Fable 5 / Mythos 5" from the recommendation pool because they "require the account-level provider_data_share data-retention opt-in (prompts shared with the provider), which is a compliance decision, not a default recommendation." That reasoning applies to 5.1 verbatim, but 5.1 isn't named.

So as written, Fable 5.1 arrives as a recommendable frontier model with no opt-in caveat and no prompt-sharing disclosure - the one property the existing text was written to prevent. Rewording those sentences to cover the 5.x frontier line rather than a single version string would keep this from recurring at 5.2.

Applies identically to the `migrate/plugins/migration-to-aws/...` mirror at the same line.

advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/references/shared/pricing-cache.md
```diff
@@ -392,6 +392,8 @@ See `shared/ai-model-lifecycle.md` for lifecycle details. **Do not recommend Leg
...
+| Claude Mythos 5.1                | anthropic.claude-mythos-5-1              | Anthropic | TBD        | TBD         | TBD     | frontier  | limited access                                               |
```
ayn-builds: `limited access` isn't one of the three Status values `shared/ai-model-lifecycle.md:122-130` defines for this column (`active` / `legacy (EOL YYYY-MM-DD)` / `excluded (EOL YYYY-MM-DD)`).

That matters because every downstream prohibition is keyed on Legacy/Excluded specifically - `estimate-ai.md:22-23`, and the checklist at `estimate-ai.md:222` reads "`recommended_model` is an Active model (not Legacy)". A row that is neither Active nor Legacy satisfies that check by default, so Mythos 5.1 can be emitted as `recommended_model` and then fail at invoke time with AccessDeniedException.

Suggest a vocabulary value plus an explicit non-recommendable annotation, so the restriction is machine-readable rather than implied by prose.

(Line 395's `active (GA on AWS)` is fine by contrast - line 397 already uses the same annotated-`active` shape.)

Applies identically to the `migrate/plugins/migration-to-aws/...` mirror at the same line.

advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/references/shared/pricing-cache.md
leon1418: Agreed on both points, and this comment was written against the pipeline's original `TBD` proposal — the row itself is superseded.

- The `TBD` cells are gone: 944e89c0 fills the row from the Bedrock pricing page (Anthropic tab) and the Fable 5.1 model card — $10.00 / $50.00, 1M context, cache write 12.50 / 20.00, cache read 0.25. Nothing `_unverified_` remains in the Anthropic rows.
- The latent gap you found is real independent of this row: the `_unverified_`-is-blocking rule at `estimate-ai.md:126` was scoped by its heading to OpenAI models. efacaf54 rewrites it as **Pricing source caveat (all providers)** — an `_unverified_` cell blocks any quoted figure whatever the provider, with the Price List API gap given as the usual cause rather than the scope. Applied to both plugin copies.

leon1418: Done, and taking your version-agnostic suggestion:

- `pricing-cache.md:386` and `design-refs/ai-anthropic-to-bedrock.md:22` now say "do not default to any Claude Fable / Mythos frontier model (Fable 5, Fable 5.1, and successors)" (efacaf54), so a 5.2 launch cannot silently become recommendable. The narrative at :386 also states the Covered-Model / `aws_review` requirement and Fable 5.1's CRIS-only commercial footprint.
- `ai-openai-to-bedrock.md:14` and `ai-gemini-to-bedrock.md:11` are updated in #269, which already rewrites those exact lines for the Sonnet 5 price change; keeping them there avoids a guaranteed conflict between the two PRs.
- `agent-advisor/references/models/anthropic-bedrock-2026-07-21.json:7`: the disclaimer now names Fable 5.1 / Mythos 5.1 and any later release, and states each model's opt-in mode — `provider_data_share` for Fable 5 / Mythos, `aws_review` for Fable 5.1 (prompts retained up to 30 days for AWS review), per the respective model cards. Both plugin copies.

advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/references/shared/pricing-cache.md
leon1418: The Mythos 5.1 row is removed in 944e89c0 rather than re-labelled. Its model card describes a Preview / Beta Service gated to vetted cybersecurity and life-sciences research organizations through Anthropic's trusted-access programs, requiring the `provider_data_share` retention mode — not a migration target this skill should ever emit as `recommended_model`, so it should not be in the table at all. The narrative above the table now says in one sentence why it is absent, and the agent-advisor catalog disclaimer excludes it explicitly.

The Fable 5.1 row keeps the annotated-`active` shape you noted is fine: `active (Covered Model; \`aws_review\` opt-in; CRIS)`.

advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/references/shared/pricing-cache.md
```diff
@@ -383,7 +383,7 @@ Serverless inference: $0.0000200 per second per GB memory.
...
...gion access in AWS GovCloud (US) only. Claude Mythos 5.1 is a gated Preview for vetted cyber/bio research organizations and is not a migration target. **Claude Opus 4.8** keeps the same $5/$25 rate as Opus 4.6/4.7. **Claude Sonnet 5** launched June 30 with introductory pricing of $2/$10 through Aug 31, 2026, then $3/$15 (same as Sonnet 4.6). **Claude Opus 4.7** lists the same headline on-demand input/output as **Opus 4.6** on that page; confirm **batch** availability per model (Opus 4.7 batch was **not** listed on the global cross-region table when this row was added). **Claude Opus 4.1** entered **Legacy** on Jul 8, 2026 (EOL Jan 8, 2027). **Batch**, **prompt cache** (5m / 1h write + cache read), and **geo / in-region cross-region** rows on that page can differ; e.g. **US East (Ohio)** cross-region inference for Claude Sonnet 4.6 is listed at **$3.30 / $16.50** per 1M input/output (≈10% above N. Virginia). Long-context SKUs **do not** all use the same multiplier: **Sonnet 4.6** and **Opus 4.6** long-context modes share the same on-demand rates as the non–long-context rows on the standard table; **Sonnet 4.5** and **Sonnet 4** long-context rows are priced higher on that same table.
```
leon1418: [🤖 AI review 🤖] The narrative says "Both are Anthropic **Covered Models**: using them requires opting the account into the \`aws_review\` data-retention mode" — but Fable 5 requires \`provider_data_share\` (prompts shared with the provider), not \`aws_review\` (prompts retained up to 30 days for AWS review). The JSON catalog disclaimer at \`anthropic-bedrock-2026-07-21.json:7\` in this same PR correctly distinguishes the two modes.

As written, the narrative tells a reader that \`aws_review\` is all they need for either model, when Fable 5 actually requires the more permissive \`provider_data_share\` mode. This could lead to an account being opted into \`aws_review\` only and then getting AccessDenied on Fable 5 invoke.

Suggest matching the JSON's phrasing: name each model's specific retention mode, or at minimum say "requiring an account-level data-retention opt-in (\`provider_data_share\` for Fable 5, \`aws_review\` for Fable 5.1)".

advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/references/shared/pricing-cache.md
```diff
@@ -392,6 +392,7 @@ See `shared/ai-model-lifecycle.md` for lifecycle details. **Do not recommend Leg
 | Model                            | Model ID                                 | Provider  | Input $/1M | Output $/1M | Context | Tier      | Status                                                       |
 | -------------------------------- | ---------------------------------------- | --------- | ---------- | ----------- | ------- | --------- | ------------------------------------------------------------ |
 | Claude Fable 5                   | anthropic.claude-fable-5                 | Anthropic | 10.00      | 50.00       | 1M      | frontier  | active                                                       |
+| Claude Fable 5.1                 | anthropic.claude-fable-5-1               | Anthropic | 10.00      | 50.00       | 1M      | frontier  | active (Covered Model; `aws_review` opt-in; CRIS)            |
 | Claude Sonnet 5                  | anthropic.claude-sonnet-5                | Anthropic | 2.00       | 10.00       | 1M      | flagship  | active (intro $2/$10 thru Aug 31, 2026; then $3/$15)         |
```
leon1418: [🤖 AI review 🤖] Nit: The quick-reference table still lists Sonnet 5 at $2.00/$10.00 with the note "intro $2/$10 thru Aug 31, 2026; then $3/$15". Since Aug 31 has passed (today is Sep 3), the table row is now stale — the live price is $3/$15. Pre-existing, not introduced by this PR, so non-blocking; noting it since this change already bumps `Last updated` to 2026-09-03 and touches the same narrative paragraph.

leon1418 (commented):
[🤖 AI review 🤖] **Review of PR #267** at head `efacaf54` (8 files, +14/-10)

**Summary:** Records Claude Fable 5.1 (GA Sep 1, 2026) in the pricing cache with verified pricing ($10/$50, cache read $0.25), access model (Covered Model, aws_review opt-in, CRIS-only in commercial regions), and model ID. Removes the pipeline-proposed Mythos 5.1 row (gated Preview, not a migration target). Widens the _unverified_-is-blocking pricing guard from OpenAI-only to all providers. Makes the "do not default" guardrails version-agnostic (Fable/Mythos line, not just Fable 5). Updates the agent-advisor JSON catalog disclaimer to name each model data-retention opt-in mode.

**Findings:**

1. **Mandatory (1):** The pricing-cache.md narrative at L386 says "Both are Anthropic Covered Models: using them requires opting the account into the aws_review data-retention mode" -- but Fable 5 requires provider_data_share, not aws_review. The JSON disclaimer in the same PR correctly distinguishes the two. Risk: a reader following the narrative opts into aws_review only and gets AccessDenied on Fable 5 invoke.

2. **Nit (1):** Sonnet 5 quick-reference row still shows $2/$10 intro pricing that expired Aug 31. Pre-existing; noted since Last updated is bumped to Sep 3.

**Cross-plugin parity:** PASS -- all 4 file pairs (advisor / migrate) share identical blob SHAs. 0 drift allowlist changes.

**Existing threads:** ayn-builds raised 3 findings on the pipeline original proposal; all 3 are addressed in commits 944e89c0 (verified pricing, Mythos row removed) and efacaf54 (version-agnostic guardrails, provider-scoped caveat). Author replies reference specific commits.

**CI:** 8/8 SUCCESS (build, gitleaks, bandit, semgrep, checkov x2, Bandit, gitleaks).
**Approvals:** 0 (2 prior reviews DISMISSED by new commits).
**Mergeable:** MERGEABLE after the mandatory finding is fixed and approvals obtained.

**Validation evidence:** Diff read via gh pr diff, file contents verified at head via GitHub Contents API, blob SHA parity confirmed for all 4 cross-plugin pairs. No tests executed by this review (change is reference data only; PR description reports drift:check OK, fmt:check OK, node tests 62+62+12).

advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/references/shared/pricing-cache.md
leon1418: The contradiction was real, but the narrative was the correct side and the JSON catalog was the stale one. Checked both model cards today (2026-09-03):

- [Fable 5 model card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-fable-5.html), Data Retention: "To use this model, you must opt in to AWS review by setting your data retention mode to `aws_review` via the Data Retention API."
- [Fable 5.1 model card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-fable-5-1.html), Data Retention: identical sentence, `aws_review`.
- [Mythos 5](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-mythos-5.html) and [Mythos 5.1](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-mythos-5-1.html) are the ones that say "you must opt in to provider data sharing by setting your data retention mode to `provider_data_share`."

So `provider_data_share` for Fable 5 was a pre-existing error in `anthropic-bedrock-2026-07-21.json:7` (dated 2026-07-21) that efacaf54 carried forward. 732ac66b fixes the catalog in both plugin copies to say `aws_review` for Fable 5 / 5.1 and `provider_data_share` for Mythos 5 / 5.1, with the model-card date and a note about the earlier revision. The pricing-cache narrative ("both are Covered Models … `aws_review`") is unchanged and now agrees with the catalog.

advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/references/shared/pricing-cache.md
leon1418: Not stale — the opposite happened. Anthropic's pricing docs (https://platform.claude.com/docs/en/about-claude/pricing) now state: "The $2/$10 per million input/output token pricing for Claude Sonnet 5, announced at launch as introductory pricing through August 31, 2026, is now the standard price. The previously scheduled increase to $3/$15 per million input/output tokens on September 1, 2026 will not occur." The Bedrock pricing page (Anthropic tab, read 2026-09-03) lists Sonnet 5 at $2.00 / $10.00 with no promotional qualifier.

The row's "intro … then $3/$15" wording is what is out of date, and it is rewritten — together with every Sonnet 5 comparison that had been computed at the cancelled $3/$15 — in #269, which is scoped to that change. Kept out of this PR to keep the two diffs reviewable on their own.

leon1418:
Re the mandatory finding at head efacaf54: the narrative was right and the JSON catalog was wrong, per the live model cards — Fable 5 and Fable 5.1 both require `aws_review`; `provider_data_share` is the Mythos 5 / 5.1 requirement. 732ac66b corrects the catalog (both plugin copies); details and model-card quotes in the inline thread. The Sonnet 5 nit is addressed in #269 (the $2/$10 rate became the standard price on Sep 1; the "then $3/$15" wording is the stale part).

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/design-refs/ai-anthropic-to-bedrock.md
```diff
@@ -19,7 +19,7 @@
 
 Older Claude models — Claude 3.5 Haiku, Claude 3 Sonnet, Claude 3.5 Sonnet (v1/v2), Claude 3 Haiku, and Claude 3.7 Sonnet — are past EOL or within the 90-day exclusion window. Do **not** recommend them as migration targets. See `shared/ai-model-lifecycle.md` for authoritative status (recomputed each run).
 
-**Recommendation:** Default new Bedrock targets to **Claude Sonnet 5** (flagship) / **Claude Opus 4.8** (hardest reasoning) / **Claude Haiku 4.5** (cost/speed). Converse API call shape is identical across generations. Do **not** default to Claude Fable 5 (frontier / Mythos-class pricing — opt-in only).
+**Recommendation:** Default new Bedrock targets to **Claude Sonnet 5** (flagship) / **Claude Opus 4.8** (hardest reasoning) / **Claude Haiku 4.5** (cost/speed). Converse API call shape is identical across generations. Do **not** default to any Claude Fable / Mythos frontier model — Fable 5, Fable 5.1, or a later 5.x (frontier / Mythos-class pricing; opt-in Covered Models requiring an account-level data-retention mode).
```
ayn-builds, line 22: Thanks for taking the version-agnostic rewrite. It landed on 2 of the 4 sites: this line and `pricing-cache.md:386`. Still literal at:

- `design-refs/ai-gemini-to-bedrock.md:11` - "Do not default to Claude Fable 5."
- `design-refs/ai-openai-to-bedrock.md:14` - "Do not default to Claude Fable 5."

That matters more than a wording inconsistency because the loader routes exclusively by source provider (`phases/design/design-ai.md:26-33`, and the same mapping in `design-refs/ai.md:9-10`): `ai_source: "gemini"` loads only the Gemini guide, `"openai"` loads only the OpenAI guide plus `shared/openai-on-bedrock.md`. So on a Gemini-source or OpenAI-source migration, the version-agnostic rule is never in context and the only Fable guardrail present names 5 alone.

Both of those files also carry a "**Recommend defaults (Jul 2026):**" date stamp that this PR's content postdates.

Applies identically to the `advisor/plugins/aws-startup-advisor/...` mirror.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/shared/pricing-cache.md
```diff
@@ -392,6 +392,7 @@ See `shared/ai-model-lifecycle.md` for lifecycle details. **Do not recommend Leg
 | Model                            | Model ID                                 | Provider  | Input $/1M | Output $/1M | Context | Tier      | Status                                                       |
 | -------------------------------- | ---------------------------------------- | --------- | ---------- | ----------- | ------- | --------- | ------------------------------------------------------------ |
 | Claude Fable 5                   | anthropic.claude-fable-5                 | Anthropic | 10.00      | 50.00       | 1M      | frontier  | active                                                       |
+| Claude Fable 5.1                 | anthropic.claude-fable-5-1               | Anthropic | 10.00      | 50.00       | 1M      | frontier  | active (Covered Model; `aws_review` opt-in; CRIS)            |
```
ayn-builds: The value is in-vocabulary now, which addresses the first half of my earlier comment, but the annotation contradicts what the value means.

`shared/ai-model-lifecycle.md:126` defines the `active` cell as:

```
| `active` | No restrictions |
```

and `phases/estimate/estimate-ai.md:23` restates it: "**Active**: no restrictions." This row is `active` carrying three restrictions in its parenthetical.

The non-recommendable half is still not machine-readable. The only gate is the checklist at `estimate-ai.md:222` - "`recommended_model` is an Active model (not Legacy) unless no Active alternative exists" - which screens Legacy only, so `active (Covered Model; ...)` passes and Fable 5.1 can be emitted as `recommended_model`. The prose guardrails at `:386` and `ai-anthropic-to-bedrock.md:22` are the only thing standing in the way, and they say "do not **default** to", not "do not select".

Also, "Covered Model" and `aws_review` appear nowhere outside this line, `:386`, and the agent-advisor JSON disclaimer. `ai-model-lifecycle.md` is the file that governs this column and defines neither.

The cleanest fix is probably a fourth Status value in `ai-model-lifecycle.md:126-128` for access-restricted models, with a matching line in the `estimate-ai.md:20-23` lifecycle block and a checklist item at `:222`. Prose alone won't hold, and this is the value an agent reads.

Applies identically to the `advisor/...` mirror.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/shared/pricing-cache.md
```diff
@@ -461,6 +462,7 @@ Per 1M tokens unless noted. See [Bedrock pricing](https://aws.amazon.com/bedrock
 | Model                    | Batch in | Batch out | 5m cache write | 1h cache write | Cache read |
 | ------------------------ | -------- | --------- | -------------- | -------------- | ---------- |
 | Claude Fable 5           | 5.00     | 25.00     | 12.50          | 20.00          | 1.00       |
+| Claude Fable 5.1         | —        | —         | 12.50          | 20.00          | 0.25       |
```
ayn-builds: This row contradicts the narrative added in the same commit. `:386` says "5.1's only rate difference is cache read at $0.25 (0.025x) vs $1.00", but here Fable 5.1's batch in/out are `—` where Fable 5 (`:464`) has `5.00 / 25.00`. Batch is a ~50% discount, so if it is unavailable on 5.1 that is a larger cost difference than the cache-read delta, and in the opposite direction.

The marker is ambiguous, and in this file's Batch columns specifically the established not-offered marker is `N/A`: `:497` (Pixtral Large), `:613-616` (Llama 3.2), `:647-648` (Llama 3.1 latency-optimized), `:714` (Nova 2 Pro Preview). Elsewhere `—` means unknown or unlisted rather than not offered, in the Model ID column (`:402-404`, `:440-442`) and the Context column (`:421`, `:438-439`).

Which one is meant changes the downstream behavior, and neither marker reaches the new rule you added at `estimate-ai.md:126`: it blocks only on `_unverified_`. So if 5.1's batch rates simply are not published yet, that is exactly the case the widened rule was written for, and `—` walks past it. `:386`'s existing instruction "confirm **batch** availability per model" gives no way to record the answer.

Suggest `N/A` if batch is genuinely not offered for 5.1, `_unverified_` if it is unconfirmed, and dropping "only" from the `:386` sentence either way.

Applies identically to the `advisor/...` mirror.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/shared/pricing-cache.md
```diff
@@ -383,7 +383,7 @@ Serverless inference: $0.0000200 per second per GB memory.
...
...gion access in AWS GovCloud (US) only. Claude Mythos 5.1 is a gated Preview for vetted cyber/bio research organizations and is not a migration target. **Claude Opus 4.8** keeps the same $5/$25 rate as Opus 4.6/4.7. **Claude Sonnet 5** launched June 30 with introductory pricing of $2/$10 through Aug 31, 2026, then $3/$15 (same as Sonnet 4.6). **Claude Opus 4.7** lists the same headline on-demand input/output as **Opus 4.6** on that page; confirm **batch** availability per model (Opus 4.7 batch was **not** listed on the global cross-region table when this row was added). **Claude Opus 4.1** entered **Legacy** on Jul 8, 2026 (EOL Jan 8, 2027). **Batch**, **prompt cache** (5m / 1h write + cache read), and **geo / in-region cross-region** rows on that page can differ; e.g. **US East (Ohio)** cross-region inference for Claude Sonnet 4.6 is listed at **$3.30 / $16.50** per 1M input/output (≈10% above N. Virginia). Long-context SKUs **do not** all use the same multiplier: **Sonnet 4.6** and **Opus 4.6** long-context modes share the same on-demand rates as the non–long-context rows on the standard table; **Sonnet 4.5** and **Sonnet 4** long-context rows are priced higher on that same table.
```
ayn-builds: The CRIS facts added here are the right facts, but nothing downstream can act on them, and two of them are hard compliance filters.

**The Model ID column contradicts this sentence.** `:386` says 5.1 "is served commercially only via `us.`/`global.` CRIS profiles (no in-region)", while `:395` lists the bare `anthropic.claude-fable-5-1`. The repo documents that exact failure in `llm-to-bedrock/references/helpers/bedrock-known-fixes/references/bedrock-inference-profile-model-id.md`:

> `ValidationException: Invocation of model ID <bare-model-id> with on-demand throughput isn't supported. Retry your request with the ID or ARN of an inference profile that contains this model.`

That doc's fix is the `us.` prefix, and `llm-to-bedrock/scripts/bedrock_pricing.py` keys its table on prefixed IDs (`us.anthropic.claude-sonnet-5` and so on). The quick-reference row is the machine-readable surface here, so it should carry the invokable form.

**GDPR makes 5.1 unusable, and nothing checks.** `phases/design/design-ai.md:81` states Step 0.7 constraints are "hard filters, not preferences", and the `gdpr` row at `:87` reads: "model IDs must use **geographic `eu.` inference profiles** — `global.` profiles route outside the EU boundary and are forbidden." Per `:386`, 5.1 has no `eu.` profile and only `us.`/`global.`, so it cannot satisfy a `gdpr` design at all. Step 0.7 takes no per-model CRIS-profile input, so the conflict is invisible at selection time.

**CCPA is adjacent.** The `soc2` / `ccpa` row requires a "prompt/completion retention policy + deletion workflow for logged content", and `aws_review` means prompts retained up to 30 days for AWS review (per the agent-advisor JSON disclaimer). That is a material input to that plan and currently reaches it only as prose.

Worth noting the `fedramp` row is the one that comes out fine: GovCloud in-region access is exactly what `:386` records.

Fixes here are cheap relative to the failure mode: prefix the ID at `:395`, and add a line to the `gdpr` row (or the Step 0.7 preamble) excluding models with no `eu.` profile. See also my note on `:395` about the Status value, which is the other half of this.

Applies identically to the `advisor/...` mirror.

migrate/plugins/migration-to-aws/skills/agent-advisor/references/models/anthropic-bedrock-2026-07-21.json
```diff
@@ -4,7 +4,7 @@
...
+  "disclaimer": "Point-in-time capability data. Re-probe every selected model and API path in the target account and region before implementation. Claude Sonnet 5 (launched 2026-06-30, after the reference snapshot) replaces Sonnet 4.6 in the recommendation pool: per its AWS model card it supports Messages on bedrock-mantle as well as Converse/Invoke on bedrock-runtime, with 1M context / 128K output and Global+geo CRIS. Claude Fable 5 / Fable 5.1 / Mythos 5 / Mythos 5.1 — and any later Fable or Mythos release — are deliberately NOT pool candidates: they are Covered Models that require an account-level data-retention opt-in — aws_review for Fable 5 and Fable 5.1 (prompts retained up to 30 days for AWS review), provider_data_share for Mythos 5 and Mythos 5.1 (prompts shared with the provider), per the respective model cards read 2026-09-03 (an earlier revision of this note put Fable 5 on provider_data_share; the Fable 5 card says aws_review) — which is a compliance decision, not a default recommendation. Fable 5.1 (GA 2026-09-01) is additionally CRIS-only in commercial regions.",
```
ayn-builds: Three things in this field, smallest first.

**It now carries its own revision history:** "(an earlier revision of this note put Fable 5 on provider_data_share; the Fable 5 card says aws_review)". The correction is right and the model-card check was the right call, but `disclaimer` is surfaced at runtime, so this reads to a consumer as a statement about Bedrock rather than about this file's git history. Git log or the PR description is the better home.

**The file's verification metadata no longer covers its contents.** `verified_at` is `2026-07-21`, `source` says "retrieved 2026-07-21", and the filename is `anthropic-bedrock-2026-07-21.json`, while the disclaimer now asserts facts "read 2026-09-03" plus a GA date of `2026-09-01`. There is a precedent for exactly this: Sonnet 5 also postdates the snapshot, and it carries its own dated `evidence` string ("AWS model card model-card-anthropic-claude-sonnet-5 (2026-07-21) …") rather than relying on the file-level date. A dated clause naming the model cards would keep the two consistent. Related: `pricing-cache.md`'s `**Last updated:**` was correctly bumped to `2026-09-03` in this same commit.

**The stated reason for excluding Mythos 5.1 understates it.** Here it is grouped with the others as requiring "an account-level data-retention opt-in … which is a compliance decision, not a default recommendation" - i.e. one opt-in away from usable. `pricing-cache.md:386` says it is "a gated Preview for vetted cyber/bio research organizations", which no opt-in reaches. Since this JSON is the agent-advisor's fact base, the stronger of the two characterizations should be the one recorded.

Applies identically to the `advisor/...` mirror.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/design-refs/ai-anthropic-to-bedrock.md
leon1418: You're right that the routing makes per-guide wording insufficient. 371bc120 adds the rule where every design passes regardless of `ai_source`: a **Restricted models** paragraph at the top of `design-ai.md` Part 1 (never `recommended_model` / `backup_model`, never a default in any guide, offer only on explicit request with the opt-in stated). The two literal "Fable 5" sentences and the `(Jul 2026)` stamps in the gemini/openai refs are rewritten in #269 (`c0e2cc0e`), which already touches those exact lines for the Sonnet 5 change; once both merge, the guide text and the always-loaded rule agree.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/shared/pricing-cache.md
leon1418: Agreed — `active` means "no restrictions" and the checklist screened Legacy only, so the parenthetical was doing nothing an agent could act on. 371bc120 adds a fourth Status value `restricted (<reason>)` to `ai-model-lifecycle.md` (Status table, Estimate rules, mapping-guide rules, and step 5 of the selection procedure), the matching lifecycle bullet plus a checklist item in `estimate-ai.md`, and applies it to both Fable rows: Fable 5 `restricted (Covered Model; aws_review opt-in)` and Fable 5.1 `restricted (Covered Model; aws_review opt-in; CRIS only)`. No script parses this column (checked `*.py` / `*.ts`), so the new value is safe. Both copies.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/shared/pricing-cache.md
leon1418: Verified against the pricing page HTML directly rather than a render: both Anthropic tables ("Global Cross-region Inference" and "Geo and In-region Cross-region Inference") hard-code **N/A** in the batch columns for Claude Fable 5 *and* Claude Fable 5.1 — so the existing Fable 5 `5.00 / 25.00` was Anthropic's first-party batch rate, not Bedrock's. 371bc120 sets both rows to `N/A`, adds a footnote saying so with the read date, and drops "only" from the :386 sentence (now: cache read differs; Bedrock lists no batch rate for either). Follow-up worth a separate look: the same page shows N/A batch for Sonnet 5 too, while our batch table carries 1.00 / 5.00 — I've left that alone here as it predates this PR and needs its own check.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/shared/pricing-cache.md
leon1418: Both taken in 371bc120. Model ID column now carries `us.anthropic.claude-fable-5-1` (no in-region form exists on `bedrock-runtime`, so the bare id would hit exactly the ValidationException the helper doc records; Fable 5 keeps the bare id because it has a us-east-1 mantle in-region endpoint). The narrative at :386 says why. Step 0.7: the `gdpr` row now excludes any candidate with no `eu.` Geo profile, naming Fable 5 / 5.1 and Mythos as the current cases; the `soc2` / `ccpa` row names the `aws_review` / `provider_data_share` retention conflict and says to exclude under CCPA unless the plan documents the retention. `fedramp` left as is — GovCloud in-region is the one path where 5.1 fits cleanly, as you noted.

migrate/plugins/migration-to-aws/skills/agent-advisor/references/models/anthropic-bedrock-2026-07-21.json
leon1418: All three in 371bc120: the revision-history parenthetical is gone (it's in the commit message of 732ac66b, where it belongs); the sentence ends with a dated evidence clause naming the four model cards read 2026-09-03 and stating that the file's 2026-07-21 snapshot predates Fable 5.1, mirroring the Sonnet 5 entry's pattern; and Mythos 5 / 5.1 are now described as gated Previews for vetted cyber/bio research organizations that no account-level opt-in reaches, which is the stronger and correct characterization.

leon1418:
Second round addressed in 371bc120 (both plugin copies; drift:check, fmt:check, node tests 62+62+12, agent-advisor pytest 322 all green locally). Structural changes: a new `restricted (<reason>)` Status value with matching Estimate rules and a checklist item; a provider-agnostic restricted-model rule at the top of design-ai Part 1 so it reaches Gemini- and OpenAI-source designs; gdpr excludes models with no `eu.` profile; Fable 5.1's Model ID is the invokable `us.` form; batch for Fable 5 and 5.1 is `N/A` per both Anthropic tables on the Bedrock pricing page (the old Fable 5 batch values were first-party rates); catalog disclaimer cleaned up with a dated evidence clause. Per-thread details inline.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/shared/pricing-cache.md
```diff
@@ -383,15 +383,16 @@ Serverless inference: $0.0000200 per second per GB memory.
...
```
ayn-builds: Following up on your reply about the cancelled increase rather than reopening it - I agree with the fact and with keeping the rewrite in its own diff. The part still unresolved is order.

As this branch stands, `:386` says Sonnet 5 goes to $3/$15 after Aug 31 and `:396`'s Status says the same. Sonnet 5 is the flagship default, so if this lands first, anything reaching for the "steady-state" rate overstates the default model by 50% until #269 follows.

The two branches don't merge cleanly either. `git merge-tree --write-tree pr267 pr269` exits 1 with content conflicts in four files: both mirrors of this one and both of `ai-anthropic-to-bedrock.md`. So the ordering isn't free regardless.

Landing #269 first and rebasing this one is the cheaper direction, and it also picks up the two guardrail sentences we deferred from the last round.

Same on the `advisor/` copy.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/shared/ai-model-lifecycle.md
```diff
@@ -116,23 +117,26 @@ When building the model comparison table:
...
+| `excluded (EOL YYYY-MM-DD)` | ≤90 days from EOL. Kept for existing users but must not be selected for new migrations.                                                                                                                                                                                                                                                     |
+| `restricted (<reason>)`     | Access-restricted: a Covered Model that needs an account-level data-retention opt-in (`aws_review` or `provider_data_share`), or a gated preview. Never `recommended_model` / `backup_model`, never a default; offer only on explicit user request, with the requirement stated. Claude Fable 5 / 5.1 and the Mythos line carry this value. |
 
 When refreshing the cache, recompute `days_to_eol` and update the Status column from the [model lifecycle page](https://docs.aws.amazon.com/bedrock/latest/userguide/model-lifecycle.html).
```
ayn-builds: This sentence quietly undoes the rest of the PR on the next refresh. It says to update the Status column from the model lifecycle page, and that page only publishes active / legacy / EOL - nothing on it can produce `restricted`. So whoever follows it literally resets Fable 5 and 5.1 to `active`, and the seven consumer sites you just threaded go quiet at once.

Worth treating as when, not if: this PR came off `kb-autoupdate/ai_model_lifecycle-claude_fable_5_1`, so the refresh is automated. Once the Status flips, both models are recommendable defaults again and the failure surfaces at invoke time as AccessDeniedException, which is exactly what the new value was added to prevent.

One clause covers it - recompute `days_to_eol` and refresh `active` / `legacy` / `excluded` from that page, but preserve an existing `restricted (…)` unless the access requirement itself changed, since the lifecycle page doesn't track access gating.

Same on the `advisor/` copy.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/shared/pricing-cache.md
leon1418: Agreed on the order. Reproduced the conflict locally — `git merge-tree --write-tree` between this head (371bc120) and #269's head (c0e2cc0e) reports 4 content conflicts, exactly the two `pricing-cache.md` and two `ai-anthropic-to-bedrock.md` copies, because the :386 paragraph is a single long line that both PRs edit. And you're right about the consequence: as this branch stands, :386 and :396 still describe the cancelled step-up to $3/$15, so landing #267 first would overstate the default flagship's steady-state rate until #269 caught up.

Plan: #269 lands first (herosjourney has approved; it is waiting on the ruleset's second approval), then this branch is rebased onto main. The rebase resolves the four conflicts in favor of #269's Sonnet 5 wording and picks up the two version-agnostic Fable guardrail sentences in the gemini/openai refs that were deferred from the previous round, so after the rebase the always-loaded rule in `design-ai.md` and every per-provider guide agree. No code change on this branch until then.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/shared/ai-model-lifecycle.md
leon1418: Good catch, and the automation angle makes it a certainty rather than a risk. Fixed in 317f3204 (both copies): the refresh sentence now recomputes `days_to_eol` and refreshes only the `active` / `legacy` / `excluded` values from the lifecycle page, and says explicitly to **preserve an existing `restricted (…)`** because that page publishes only active / legacy / EOL and never tracks access gating — a `restricted` value changes only when the model card's own access requirement (retention mode, gated-preview status) changes.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/shared/pricing-cache.md
ayn-builds: Blocking, I think: this paragraph is getting rewritten for Fable 5.1 but the Sonnet 5 sentence in it still says $2/$10 runs "through Aug 31, 2026, then $3/$15". That date's gone, so the cache now reads as though the flagship is $3/$15. Line 396 can't decide either, the rate columns say `2.00 | 10.00` and the Status cell says otherwise.

Sonnet 5 is the default recommendation, so this isn't an edge case, and a 50% overestimate shows up as the headline monthly number a customer sees.

What makes it easy to miss is the `Last updated` bump to Sep 3. The cache wasn't stale before this PR and still isn't, so nothing changes there, but a refresh dated Sep 3 says someone checked prices two days after the Sep 1 change and wrote down the old rule anyway.

#269 rewrites this same paragraph the other way (launch rate became standard Sep 1, increase cancelled), so these two conflict and whichever lands second wins. Might be easiest to take #269 first and rebase.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/phases/design/design-ai.md
```diff
@@ -80,20 +80,22 @@ If `agentic_profile.is_agentic == true`:
...
+| `fedramp`        | Target region forced to **GovCloud** (us-gov-east-1/us-gov-west-1); re-run Step 0.5 regional validation against GovCloud — the model catalog is materially smaller, and a `regional_warnings[]` entry is REQUIRED for every candidate model not available there.                                                                                                                                                                                                        |
+| `gdpr`           | Target region restricted to EU (eu-west-1, eu-central-1); model IDs must use **geographic `eu.` inference profiles** — `global.` profiles route outside the EU boundary and are forbidden. Note the GCP-EU → AWS-EU transfer in the summary. Exclude any candidate with no `eu.` Geo profile — as of Sep 2026 that includes Claude Fable 5 / 5.1 (`us.` / `global.` only) and the Mythos line.                                                                          |
```
ayn-builds: Only 5.1's footprint was verified, but this asserts Fable 5's too, and Fable 5 still has a bare in-region `anthropic.claude-fable-5` id in this PR's own table. So "`us.` / `global.` only" isn't true for it.

Per line 91 anything that changes a model choice or region gets recorded in `regional_warnings[]`, which the customer reads, and under gdpr a claim that a model isn't available in-region can change the architecture. Scoping the parenthetical to 5.1 should be enough.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/phases/design/design-ai.md
```diff
@@ -80,20 +80,22 @@ If `agentic_profile.is_agentic == true`:
...
 
 Record what was applied: every constraint that changed a model choice or region adds a `regional_warnings[]` entry (existing shape) or a Part 5 plan line naming the compliance value that forced it, and the Present Summary section MUST state which compliance regime(s) shaped the design. A design that ignores a declared compliance value is a validation failure (see Validation Checklist).
 
 ---
 
 ## Part 1: Bedrock Model Selection
 
+**Restricted models — applies whichever guide Step 0 loaded and whatever the compliance value:** a row whose `Status` in `shared/pricing-cache.md` is `restricted (…)` (Covered Models such as Claude Fable 5 / 5.1, gated previews such as Claude Mythos) is never `recommended_model` or `backup_model` and never a default in any mapping guide. Offer one only when the user explicitly asks for a frontier model, and state its data-retention opt-in and CRIS-only footprint in the same sentence.
```
ayn-builds: Same thing as line 87. CRIS-only is right for 5.1 but not for Fable 5, which has a bare in-region id in the cache. Since this tells the agent to state it "in the same sentence" every time it names a restricted model, it'll assert the restriction for a model that doesn't have it.

Narrower than the pricing one, it only fires when someone asks about a frontier model, but it's still the agent telling a customer something untrue. "state its data-retention opt-in and any CRIS-only restriction" would cover it without needing a per-model list.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/shared/pricing-cache.md
leon1418: Addressed in aaf141a0. The Anthropic narrative and Sonnet 5 Status cell now agree with the existing $2/$10 rate columns: the launch rate became the standard price on September 1, and the scheduled $3/$15 increase was cancelled.

Since #269 is still awaiting review, I brought its two pricing-cache wording fixes into this PR so #267 no longer carries the contradictory statement. The broader comparison recalculations remain in #269. Whichever PR lands second needs to preserve both the Sonnet correction and the Fable/restricted-model additions when resolving the overlapping paragraph.

Applied identically to both plugin copies. Rechecked the [official pricing note](https://platform.claude.com/docs/en/about-claude/pricing) on September 9. Cross-plugin drift, repository formatting, and all 136 Node tests pass.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/phases/design/design-ai.md
leon1418: Addressed in aaf141a0 by making the endpoint distinction explicit while retaining the EU filter.

I rechecked the [Fable 5 model card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-fable-5.html) and [Fable 5.1 model card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-fable-5-1.html) on September 9. Fable 5 does support the bare model ID for in-region access through `bedrock-mantle` in `us-east-1`. On `bedrock-runtime`, both Fable models use `us.` / `global.` inference profiles. Neither has an `eu.` Geo profile, so the US in-region option does not make Fable 5 eligible under this guide's existing EU constraint.

The GDPR row now states those endpoint-specific facts instead of describing the models as universally CRIS-only. Both plugin copies are identical.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/phases/design/design-ai.md
leon1418: Addressed in aaf141a0. The shared rule now says:

> state its data-retention opt-in and any endpoint-specific regional or CRIS-only restrictions in the same sentence.

This makes the disclosure conditional on the actual endpoint and regional support, rather than asserting CRIS-only access for every restricted model. The existing exclusion from default recommendations, `recommended_model`, and `backup_model` remains in place. Applied identically to both plugin copies; drift, formatting, and all 136 Node tests pass.

ayn-builds (dismissed):

ayn-builds (approved):

PR #268, opened 2026-09-03
fix(agent-advisor): ai-model-lifecycle.claude_fable_5_availability — new knowledge

leon1418:

## What changed upstream

[Claude Fable 5.1, Anthropic's new frontier model is now available on AWS GovCloud (US)](https://aws.amazon.com/about-aws/whats-new/2026/09/claude-fable-5-1-aws-govcloud/)

**Verdict: `new_knowledge`** on `ai-model-lifecycle.claude_fable_5_availability`

- was — Claude Fable 5 (frontier/Mythos-class) is opt-in only, not a default Bedrock target; no GovCloud availability noted.
- now — Claude Fable 5.1 is now GA on Amazon Bedrock in AWS GovCloud (US), designated a "Covered Model" with additional data retention/safety/access policies (Enterprise Frontier Safeguards); still frontier-tier, opt-in pricing.

> **Still true:** Fable-class models remain non-default, opt-in, premium/frontier pricing tier for standard commercial Bedrock recommendations.
>

## Proposed edits

### `gcp-to-aws/references/shared/pricing-cache.md:386` — value

```diff
- Do not default to Claude Fable 5 (frontier). **Claude Fable 5** is the most expensive
+ Do not default to Claude Fable 5 (frontier). **Claude Fable 5** is the most expensive; note Claude Fable 5.1 is now GA on Bedrock in AWS GovCloud (US) as a Covered Model with additional data retention/safety/access policies via Enterprise Frontier Safeguards
```

**Why:** Announcement states Fable 5.1 is GA on Bedrock in AWS GovCloud (US) and is a Covered Model, which the pricing cache narrative doesn't reflect at all

**Evidence (verbatim from the announcement):** "Claude Fable 5.1 is generally available on Amazon Bedrock in AWS GovCloud (US)"

## Not applied

The pipeline proposed these and could not apply them safely. They are listed so this PR
is not mistaken for a complete change:

- `gcp-to-aws/references/shared/pricing-cache.md:395` (value) — before-text not found at or near line 395

The same edits were applied to this second copy: 1 applied, 1 skipped:

- `gcp-to-aws/references/shared/pricing-cache.md:395` (value) — before-text not found at or near line 395


vagmur-aws (approved):

herosjourney (changes requested):
Verified the underlying claim against the [AWS announcement](https://aws.amazon.com/blogs/machine-learning/introducing-claude-fable-5-1-on-aws/) — Fable 5.1's Covered Model status, Enterprise Frontier Safeguards, and GovCloud (US) availability are all accurately described. No factual issue with the content itself.

However, the insertion breaks the sentence it's spliced into. Before:

> **Claude Fable 5** is the most expensive Anthropic model at $10/$50 per 1M tokens (Mythos-class).

After this PR:

> **Claude Fable 5** is the most expensive; note Claude Fable 5.1 is now GA on Bedrock in AWS GovCloud (US) as a Covered Model with additional data retention/safety/access policies via Enterprise Frontier Safeguards Anthropic model at $10/$50 per 1M tokens (Mythos-class).

The tail — "...Safeguards Anthropic model at $10/$50 per 1M tokens..." — is now a dangling fragment because the clause was inserted mid-sentence instead of after it. Suggest:

> **Claude Fable 5** is the most expensive Anthropic model at $10/$50 per 1M tokens (Mythos-class). Claude Fable 5.1 is now GA on Bedrock in AWS GovCloud (US) as a Covered Model with additional data retention/safety/access policies via Enterprise Frontier Safeguards.

This needs to be fixed identically in both mirrored files (`advisor/plugins/aws-startup-advisor/skills/...` and `migrate/plugins/migration-to-aws/skills/...`).

Separate note: the PR title references the key `ai-model-lifecycle.claude_fable_5_availability`, but neither `ai-model-lifecycle.md` file was touched — that file currently has no mention of "Fable" at all. If that's the intended source of truth for model lifecycle status, it seems like the actual target of this update.

Requesting changes for the grammar break; happy to re-review once fixed.

advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/references/shared/pricing-cache.md
```diff
@@ -383,7 +383,7 @@ Serverless inference: $0.0000200 per second per GB memory.
...
...odel with additional data retention/safety/access policies via Enterprise Frontier Safeguards Anthropic model at $10/$50 per 1M tokens (Mythos-class). **Claude Opus 4.8** keeps the same $5/$25 rate as Opus 4.6/4.7. **Claude Sonnet 5** launched June 30 with introductory pricing of $2/$10 through Aug 31, 2026, then $3/$15 (same as Sonnet 4.6). **Claude Opus 4.7** lists the same headline on-demand input/output as **Opus 4.6** on that page; confirm **batch** availability per model (Opus 4.7 batch was **not** listed on the global cross-region table when this row was added). **Claude Opus 4.1** entered **Legacy** on Jul 8, 2026 (EOL Jan 8, 2027). **Batch**, **prompt cache** (5m / 1h write + cache read), and **geo / in-region cross-region** rows on that page can differ; e.g. **US East (Ohio)** cross-region inference for Claude Sonnet 4.6 is listed at **$3.30 / $16.50** per 1M input/output (≈10% above N. Virginia). Long-context SKUs **do not** all use the same multiplier: **Sonnet 4.6** and **Opus 4.6** long-context modes share the same on-demand rates as the non–long-context rows on the standard table; **Sonnet 4.5** and **Sonnet 4** long-context rows are priced higher on that same table.
```
ayn-builds, line 386: This inserts into the middle of an existing sentence rather than adding a new one, and the result misattributes a price.

Before:

> **Claude Fable 5** is the most expensive Anthropic model at $10/$50 per 1M tokens (Mythos-class).

After:

> **Claude Fable 5** is the most expensive; note Claude Fable 5.1 is now GA on Bedrock in AWS GovCloud (US) as a Covered Model with additional data retention/safety/access policies via Enterprise Frontier Safeguards Anthropic model at $10/$50 per 1M tokens (Mythos-class).

"the most expensive Anthropic model at $10/$50" got cut in half. Two consequences:

- "Fable 5 is the most expensive" is now dangling with no object.
- "…via Enterprise Frontier Safeguards Anthropic model at $10/$50 per 1M tokens (Mythos-class)" attaches the $10/$50 to **Fable 5.1**, the nearest subject. Fable 5.1's price is unknown (#267 lists it `TBD`), and GovCloud rates are higher than commercial anyway - this file's own precedent is `openai-on-bedrock.md:173`, where GPT-5.4 is `2.75 · 16.50` commercial vs `GovCloud (US-West): 3.30 · 19.80`.

So an estimate phase reading this can quote Fable 5.1 at $10/$50 in GovCloud, which is unsupported on both counts. This one should block: it turns a correct statement into an ambiguous one.

Applies identically to the `migrate/plugins/migration-to-aws/...` mirror at the same line.

ayn-builds, line 386: Line 4 of this file declares `**Region:** us-east-1`, and the edited paragraph opens by scoping itself to "figures below match **US East (N. Virginia)**". A GovCloud availability statement inside it reads as part of the commercial N. Virginia narrative.

More importantly, nothing consumes it. I traced both regional paths:

- `phases/design/design-ai.md:40-54` (Step 0.5, Regional Availability Validation) calls `get_regional_availability` on the `awsknowledge` MCP, and on failure falls back to the static table in `shared/ai-migration-guardrails.md` - which covers AgentCore services only (lines 7-21), no Bedrock models.
- In agent-advisor, `region_availability_note` (`phases/design/design.md:110`) is about runtime availability, not model regions.

Neither reads this prose. Meanwhile `design-ai.md:86` says that under `fedramp` the region is forced to GovCloud and "a `regional_warnings[]` entry is REQUIRED for every candidate model not available there" - the exact decision this fact should inform, and it can't reach it.

The repo already solved this one month earlier for the OpenAI GovCloud launch, in two places:

- `shared/openai-on-bedrock.md:113-125` - "Regional Availability — Endpoint-Aware" matrix with `us-gov-west-1` / `us-gov-east-1` columns, plus a GovCloud rate note at :173.
- `agent-advisor/references/models/openai-bedrock-2026-08-21.json` - structured per-model `region_availability: {commercial: [...], govcloud: [...], evidence: "..."}`.

Neither has an Anthropic counterpart: there's no `anthropic-on-bedrock.md` in `shared/` at all, and `anthropic-bedrock-2026-07-21.json` is the same `schema_version: 2` but has no `region_availability` key on any model (I checked for `region_availability` / `regions` / `govcloud` / `availability`). So the real work here is creating that home - most plausibly `region_availability` on the Anthropic catalog, mirroring the OpenAI shape - not a prose splice.

Related: "GovCloud (US)" flattens a distinction the repo records precisely. The announcement doesn't say West or East, and the OpenAI catalog handles exactly that ambiguity explicitly: "Sol's GovCloud availability is not stated; treat as unknown, not absent." Please record the region set per that convention rather than collapsing it.

ayn-builds, line 386: Three smaller things, all in the added text:

**"Covered Model" and "Enterprise Frontier Safeguards" are undefined.** Neither term appears anywhere else in the repo. This is a compliance-relevant access restriction, but it lands as prose only - not in the quick-reference Status column, and not in `shared/ai-model-lifecycle.md`, whose Status vocabulary (lines 122-130: `active` / `legacy (EOL …)` / `excluded (EOL …)`) has no value that can express "restricted access pending a policy opt-in."

**The prose names a model with no entry on this branch.** There is no `fable-5-1` row anywhere here - line 395 is the Sonnet 5 row. If this merges before #267, the cache narrates a model with no model ID, no price, and no Status. The PR body confirms the race: its "Not applied" section reports "before-text not found at or near line 395", and line 395 is exactly where #267 inserts the Fable 5.1 row. Worth sequencing this after #267. (I tested the merge - they don't conflict textually, so both can land in either order without git complaining.)

**Two carryovers from #267 are still open here.** "Do not default to Claude Fable 5" one clause earlier is still version-specific, now sitting directly beside a sentence announcing 5.1 is GA, which makes the gap sharper. And `**Last updated:** 2026-08-24` (line 3) is unchanged while September data is added.

leon1418:
Closing: the proposed edit splices a clause into the middle of an existing sentence and breaks it ("…is the most expensive; note Claude Fable 5.1 … Enterprise Frontier Safeguards Anthropic model at $10/$50…"). The GovCloud fact is correct and will land, human-verified against the Fable 5.1 model card, together with the Fable 5.1 pricing rows in a reworked #267.

PR #274, opened 2026-09-04
fix(agent-advisor): anthropic.model_lineup — new knowledge

leon1418:

## What changed upstream

[We've launched Claude Fable 5.1 (`claude-fable-5-1`), the successor to Claude Fable 5 for long-running agentic coding, knowledge work, and research, alongside C](https://docs.claude.com/en/release-notes/api.md)

**Verdict: `new_knowledge`** on `anthropic.model_lineup`

- was — No entry for Claude Fable 5/5.1 or Mythos 5.1 pricing/specs in the reviewed files; ai-anthropic-to-bedrock.md only mentions "Claude Fable 5" once as an explicit non-default (frontier/Mythos-class pricing, opt-in only), with no ID, price, or context window given.
- now — Claude Fable 5.1 (`claude-fable-5-1`) and Claude Mythos 5.1 (`claude-mythos-5-1`) launched: 1M token context (default), 128k max output tokens, always-on adaptive thinking, $10/$50 per MTok (same as Fable 5), cache reads $0.25/MTok. Fable 5.1 available via Claude API, Bedrock, Claude Platform on AWS, Vertex AI, and Microsoft Foundry. Mythos 5.1 is Project Glasswing participants only.

> **Still true:** The recommendation in ai-anthropic-to-bedrock.md to not default to Claude Fable 5 (opt-in only, frontier/Mythos-class pricing) still holds in spirit — Fable 5.1 is likewise a specialized/opt-in-tier model, not a default Bedrock migration target.
>

## Proposed edits

### `gcp-to-aws/references/shared/pricing-cache.md:463` — value

```diff
- | Claude Fable 5           | 5.00     | 25.00     | 12.50          | 20.00          | 1.00       |
+ | Claude Fable 5.1         | 5.00     | 25.00     | 12.50          | 20.00          | 0.25       |
```

**Why:** The announcement states cache reads for Fable 5.1 are cut to $0.25 per MTok (from the prior $1.00 rate implied for Fable 5), so this row's cache-read column must be updated for the new model.

**Evidence (verbatim from the announcement):** "with cache reads cut to $0.25 per MTok"

### `gcp-to-aws/references/shared/pricing-cache.md:394` — value

```diff
- | Claude Fable 5                   | anthropic.claude-fable-5                 | Anthropic | 10.00      | 50.00       | 1M      | frontier  | active                                                       |
+ | Claude Fable 5.1                 | anthropic.claude-fable-5-1               | Anthropic | 10.00      | 50.00       | 1M      | frontier  | active                                                       |
```

**Why:** Fable 5.1 is the successor model now launched with a new model ID (claude-fable-5-1), same $10/$50 pricing and 1M context, so this pricing table row should be updated to reflect the new model/ID as current, per the announcement's explicit launch and pricing details.

**Evidence (verbatim from the announcement):** "Claude Fable 5.1** (`claude-fable-5-1`), the successor to Claude Fable 5"

## Filter false positives

- `agent-advisor/references/decision-refs/model-selection.md`

The same edits were applied to this second copy: 2 applied.

### DIFF
--- a/advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/references/shared/pricing-cache.md
+++ b/advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/references/shared/pricing-cache.md
@@ -391,7 +391,7 @@ See `shared/ai-model-lifecycle.md` for lifecycle details. **Do not recommend Leg
 
 | Model                            | Model ID                                 | Provider  | Input $/1M | Output $/1M | Context | Tier      | Status                                                       |
 | -------------------------------- | ---------------------------------------- | --------- | ---------- | ----------- | ------- | --------- | ------------------------------------------------------------ |
-| Claude Fable 5                   | anthropic.claude-fable-5                 | Anthropic | 10.00      | 50.00       | 1M      | frontier  | active                                                       |
+| Claude Fable 5.1                 | anthropic.claude-fable-5-1               | Anthropic | 10.00      | 50.00       | 1M      | frontier  | active                                                       |
 | Claude Sonnet 5                  | anthropic.claude-sonnet-5                | Anthropic | 2.00       | 10.00       | 1M      | flagship  | active (intro $2/$10 thru Aug 31, 2026; then $3/$15)         |
 | Claude Opus 4.8                  | anthropic.claude-opus-4-8                | Anthropic | 5.00       | 25.00       | 200K    | premium   | active                                                       |
 | Claude Sonnet 4.6                | anthropic.claude-sonnet-4-6              | Anthropic | 3.00       | 15.00       | 200K    | flagship  | active                                                       |
@@ -460,7 +460,7 @@ Per 1M tokens unless noted. See [Bedrock pricing](https://aws.amazon.com/bedrock
 
 | Model                    | Batch in | Batch out | 5m cache write | 1h cache write | Cache read |
 | ------------------------ | -------- | --------- | -------------- | -------------- | ---------- |
-| Claude Fable 5           | 5.00     | 25.00     | 12.50          | 20.00          | 1.00       |
+| Claude Fable 5.1         | 5.00     | 25.00     | 12.50          | 20.00          | 0.25       |
 | Claude Sonnet 5          | 1.00     | 5.00      | 2.50           | 4.00           | 0.20       |
 | Claude Opus 4.8          | 2.50     | 12.50     | 6.25           | 10.00          | 0.50       |
 | Claude Sonnet 4.6 (+ LC) | 1.50     | 7.50      | 3.75           | 6.00           | 0.30       |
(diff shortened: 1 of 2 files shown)


advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/references/shared/pricing-cache.md
```diff
@@ -391,7 +391,7 @@ See `shared/ai-model-lifecycle.md` for lifecycle details. **Do not recommend Leg
 
 | Model                            | Model ID                                 | Provider  | Input $/1M | Output $/1M | Context | Tier      | Status                                                       |
 | -------------------------------- | ---------------------------------------- | --------- | ---------- | ----------- | ------- | --------- | ------------------------------------------------------------ |
-| Claude Fable 5                   | anthropic.claude-fable-5                 | Anthropic | 10.00      | 50.00       | 1M      | frontier  | active                                                       |
+| Claude Fable 5.1                 | anthropic.claude-fable-5-1               | Anthropic | 10.00      | 50.00       | 1M      | frontier  | active                                                       |
```
leon1418, line 394: [🤖 AI review 🤖] **Nit:** The table row is now "Fable 5.1" but the narrative header above (L386) still says "Do not default to Claude Fable 5 (frontier)" and "Claude Fable 5 is the most expensive Anthropic model." This creates a name inconsistency within the same section — a reader might wonder whether Fable 5 and 5.1 are separate models or the same. Worth updating the two header references to "Fable 5.1" for consistency, either here or in a follow-up.

leon1418 (commented):
[🤖 AI review 🤖] **Reviewed head `cda16fa`.** 2 changed files (+4/−4), 1 commit. Cross-plugin parity **PASS** — both `pricing-cache.md` copies are byte-identical (SHA256 `3642a7a1d197…`). 0 drift allowlist changes.

**0 mandatory blockers. 1 Nit.**

The Fable 5 → 5.1 successor update is correct: on-demand ($10/$50), cache read ($0.25, down from $1.00), batch ($5/$25), and cache-write columns all match the announcement. Removing Fable 5 rather than keeping both rows is the right call since the announcement identifies 5.1 as a direct successor.

**Nit:** The narrative header on L386 still references "Claude Fable 5" twice ("Do not default to Claude Fable 5 (frontier)" and "Claude Fable 5 is the most expensive Anthropic model at $10/$50"). Now that the table rows say "Fable 5.1", these two mentions are stale. Low impact — the recommendation is still directionally correct — but worth updating for consistency in a follow-up or this PR.

*(Pre-existing, not introduced by this PR: Sonnet 5 status text still says "intro $2/$10 thru Aug 31, 2026" — that intro period expired Sep 1.)*

CI: 8/8 SUCCESS (build, gitleaks, bandit, semgrep, checkov). 0 approvals. REVIEW_REQUIRED. Branch: MERGEABLE. **MERGE-READY** pending approval.

PR #279, opened 2026-09-08
feat(telemetry): hook-driven migration telemetry emitter with on-disk attribution

az-zhu:
## Problem

The migration skills run entirely on the customer's machine, so there is no visibility into how many migration runs start, where they stall, or how many finish. An earlier approach, instructing the agent to invoke a reporting script at each phase boundary, makes emission depend on the model following an instruction: a skipped instruction leaves no trace, so a missing event is indistinguishable from a customer who never migrated, and no funnel number can be defended on that data.

## Solution

Emit from host hooks instead, so emission is host-driven and does not depend on the agent. Attribution, consent and state all live on disk in the migration skills' own `.migration/` tree.

- **`hooks/telemetry/emit.mjs`**: one dependency-free Node script serving three trigger roles (post-write, end-of-turn reconcile, session teardown) plus a `consent` subcommand. It reads the skills' own `.phase-status.json`, diffs it against a co-located snapshot, and POSTs one event per transition. Idempotent (re-runs send nothing new), per-run locked (concurrent triggers cannot duplicate events), fail-open (every path exits 0). A corrupt or half-written snapshot is skipped rather than read as a new run, and all state writes are atomic (temp file plus rename). Events of one reconcile are sent concurrently under one budget so a large catch-up cannot outlive the host's hook timeout. A run whose state file was last written more than two seconds before the consent record is treated as pre-consent history: it is baselined silently and only later transitions are reported; the two-second slack keeps a state file written moments before consent, or on a filesystem with coarse timestamps, from being silently dropped. A 403 from the service (its closed launch gate, or a WAF rate limit) leaves the snapshot untouched so the run replays later; only network failures are accepted as loss. Every value sent is checked against the service model's enums first: an unknown phase name emits nothing for that phase and an unmapped attribute value is omitted, so drift costs a field, never the request. Non-identifying attributes are derived from the run's own artifacts (source provider, resource count, database and AI presence, clarify mode, pricing source, recommendation outcome, spend band paired with its basis, run mode on completion).
- **`hooks/hooks.json`**: PostToolUse (`Write|Edit`, async), Stop and SessionEnd registrations in exec form beside the existing SessionStart hook.
- **Cursor**: `.cursor-plugin/plugin.json` gains a `hooks` key pointing at `hooks/telemetry/cursor/hooks.json` (afterFileEdit, stop, sessionEnd), and `advisor/plugins/.cursor-plugin/marketplace.json` lists the plugin so Cursor can reach it, mirroring the manifest `migrate/plugins/` already carries. The same script serves both hosts; it absorbs the payload differences (`conversation_id`, `workspace_roots`, top-level `file_path`) and reports `source` as `CURSOR`, `CLAUDE_CODE` or `OTHER` from the environment.
- **`skills/shared/dsl/INTERPRETER.md`**: `_init` gains the telemetry consent step, before any state is written: locate the emitter (plugin-root variable, then the plugin caches, then a pruned home search), run `consent get`, ask once with fixed wording, record the answer only through `consent grant` or `consent revoke`, and honour a later request to stop sharing with `consent revoke`. The `run_id`, `owning_skill` and `initiated_by` seeding this relies on landed separately in #300. Vendored copies synced. gcp-to-aws does its run setup in its own `references/phases/discover/discover.md` without the shared interpreter, so the same step is written there too (both plugin copies); a unit test pins the two wordings identical, since AppSec approves one text.

## Payload contract

The envelope is `installId`, `source`, `pluginVersion`, `occurredAt` plus `pluginTelemetryEvent.migrationActivity`, matching the service model as merged. The service mints `eventId` per received request for queue-redelivery dedup, so the client sends none.

## Privacy and consent

Nothing is collected or sent without explicit opt-in consent. The prompt is fixed text the interpreter shows verbatim; it states what is shared (which phases complete and how the run ends, estate size band, database and AI presence, spend as a range, the recommendation reached) and what never is (code, file paths, resource names, project or app names, account details, exact costs). A project with a recorded decision is never asked again. A decline is recorded locally as `revoked` and nothing is ever sent for it; a later request to stop sharing is honoured the same way. `DO_NOT_TRACK=1` and `AWS_STARTUP_ADVISOR_TELEMETRY=0` override stored consent. Every payload field is a closed enum, UUID, bounded integer or boolean; there is no free text. Setting `AWS_STARTUP_ADVISOR_TELEMETRY_ENDPOINT` to an empty string disables sending entirely without advancing snapshots.

## Merge gates

Merging this PR is the plugin-side launch: the next plugin update asks for consent, and consented runs post to the compiled production endpoint. Before merge:

- AppSec has approved the consent wording above.
- The service's production feature flag is removed and deployed. Until then the endpoint answers 403; the emitter treats that like a disabled endpoint and holds its snapshot, so runs consented before the flag opens are reported in full on a later trigger rather than lost. The flag is therefore a launch-completeness gate, not a data-loss gate, and this PR no longer has to merge after it.
- The CASK consumer (queue, processor) and the lake dataset are live in production, or the events have nowhere to land.

## Type of Change

- [ ] New plugin/power/tool
- [ ] Bug fix
- [ ] Documentation update
- [ ] Guardrail/CI update

## Team Folder

- [x] `advisor/`
- [ ] `migrate/`
- [ ] Other: ___

## Scope

New `hooks/telemetry/` directory (emitter, Cursor registration, README), additive edits to `hooks/hooks.json`, `.cursor-plugin/plugin.json`, the shared phase-status schema (three optional keys) and the shared DSL interpreter, plus the Cursor marketplace manifest under `advisor/plugins/`. No change to any skill's migration logic, `_exec` dispatch, or MCP config. Existing runs without the new keys are unaffected: the emitter stays silent for them. The gcp-to-aws `discover.md` change is mirrored into `migrate/plugins/migration-to-aws/` to satisfy the cross-plugin drift check, so `@awslabs/startups-migrate` review is needed for that file. Everything else sits under `advisor/`, which CODEOWNERS resolves to `@awslabs/startups-advisor` (the `/advisor/` rule is the last match, so the `**/.cursor-plugin/` admin rule does not apply here). One line in `mise.toml` registers the new test file in the `test` task, which brings in `@awslabs/startups-admins` for that file.

Emitter hardening, the consent flow and the attribute derivation are lifted from @icarthick's PoC branch ([`icarthick/startups` `feat/telemetry-hooks`](https://github.com/icarthick/startups/tree/feat/telemetry-hooks)), which was exercised against the real service handler; this PR keeps on-disk attribution, the compiled production endpoint and the `--via cli` invocation marker from the earlier draft.

## Verification

- Committed executing-path tests (`tests/hooks/telemetry-emitter.test.ts`, in `mise run test`): each test runs `emit.mjs --reconcile` exactly as the Stop hook does, with the hook payload on stdin and a local HTTP stub as the endpoint, and asserts on the requests received and the snapshot on disk. Covered: first report (`RUN_STARTED` plus already-resolved phases under the state file's `run_id`, envelope fields, `sourceProvider`), idempotent re-run, later transitions and `RUN_COMPLETED` with `runMode`, consent revoked and unset (nothing sent, no snapshot), missing `owning_skill`, and the pre-consent baseline. Nothing inside the emitter is mocked.
- Local harness against a stub HTTP endpoint: consent gate (nothing sent, nothing written locally), grant and decline (decline recorded, nothing sent), snapshot-diff correctness, idempotent re-runs, 4-way concurrent invocation producing exactly one `RUN_STARTED`, lock release, corrupt snapshot ignored and left untouched, no temp files left behind, disabled-endpoint inertia, `DO_NOT_TRACK` precedence, malformed-stdin exit 0.
- Vocabulary: an unknown phase name is dropped while known phases still emit; `skipped` and `not_applicable` map to `SKIPPED` and `NOT_APPLICABLE`; `RUN_COMPLETED` carries `runMode` from `run_mode`.
- Attributes derived from fixture artifacts match the inputs (resource count, database and AI flags, clarify mode, pricing source, recommendation outcome, spend band and basis).
- Cursor-shaped payloads (`conversation_id`, `workspace_roots`, top-level `file_path`) trigger correctly and report `source: CURSOR`.
- `node advisor/plugins/aws-startup-advisor/tools/sync-vendored-shared.ts advisor/plugins/aws-startup-advisor` OK; dprint and markdownlint clean locally.
- Real service handler: the full lifecycle above was replayed against the built `SawsStartupsAdvisor-IDEExtensionApi` Lambda handler with its Smithy validation live and only the SNS client stubbed. 9 of 9 requests returned 200 and published under `metricType=migrationActivityMetric`, including every derived attribute, `initiatingSkill`, `SKIPPED`/`NOT_APPLICABLE` statuses and `runMode`; an empty body and a bad enum both return 400.
- Live Claude Code session (headless `claude -p --plugin-dir` with this branch): the `PostToolUse` hooks fired async on a `Write` and an `Edit` of `.phase-status.json` and delivered `RUN_STARTED`, `DISCOVER` and `CLARIFY` to the real handler; `SessionEnd` ran and exited 0; the snapshot recorded the host's session id; `${CLAUDE_PLUGIN_ROOT}` resolved inside the exec-form `args`. `Stop` was not observed in headless mode and is registered in the same shape as the `SessionEnd` hook that did run.
- Finding for the consent follow-up: `CLAUDE_PLUGIN_ROOT` and `CLAUDE_PLUGIN_DATA` are not present in the agent's Bash environment (only `CLAUDECODE=1`), so skill prose cannot use them to locate the emitter; a plugin-cache search or an explicit path is required.
- Not exercised in this PR: the live beta endpoint (covered by the backend beta integration test before launch) and hook command resolution inside Cursor (to be verified before the plugin is published to the Cursor marketplace). Neither gates this merge: with no consent prompt shipped, the client sends nothing.

## Checklist

- [x] I have read the [CONTRIBUTING.md](https://github.com/awslabs/startups/blob/main/CONTRIBUTING.md) guidelines
- [ ] My changes are scoped to my team's folder only (one mirrored file under `migrate/`, required by the drift check; and one line in `mise.toml`)


advisor/plugins/aws-startup-advisor/hooks/telemetry/emit.mjs
```diff
@@ -0,0 +1,701 @@
...
+      installId: getInstallId(),
+      pluginVersion: pluginVersion(),
+    };
+
+    // Concurrently, under one budget: sent serially, a reconcile catching a
+    // whole run could outlive the host's hook timeout and never reach the
+    // snapshot write, so the next trigger would repeat the batch. No retries by
+    // design: a retry queue is unbounded local state for data that is
+    // loss-tolerant in aggregate.
+    const results = await Promise.allSettled(events.map((event) => post(endpoint, buildRequest(event, ctx))));
```
icarthick: It appears with hold writing to the snapshot when we see a 403. I think it would make sense to do the same for 429s. I see 429s as transient failures and hence the metric emission should be re-tried

But am not sure if this would result in duplicate events if the metrics where already sent. Just a thought

az-zhu: agree, made some changes

advisor/plugins/aws-startup-advisor/hooks/telemetry/emit.mjs
```diff
@@ -0,0 +1,742 @@
...
+
+function machineConsent() {
+  for (const file of machineConsentFiles()) {
+    const record = readJson(file);
+    if (record?.consent) return record;
+  }
+  return null;
+}
+
+function consentRecordFor(runDir) {
```
icarthick: consentRecordFor() looks at the machine-wide consent record (~/.aws-startups-plugins/telemetry.json, or $CLAUDE_PLUGIN_DATA/telemetry.json) and falls-back to the project's .migration/telemetry.json for consent. But nothing in this PR ever writes the machine-wide file, consent grant/revoke only write the project-level record. So machineConsent() always returns null, and consent is always read from the project directory.

advisor/plugins/aws-startup-advisor/skills/agent-advisor/references/vendored/dsl/INTERPRETER.md
```diff
@@ -575,7 +575,60 @@ skips them and keeps its own state contract.)
...
+      If no emitter is found, skip this step entirely and continue: consent
+      stays unset, telemetry stays off, and the migration is never blocked by it.
+
+   2. Run `node "$EMIT" consent get`. Anything other than `unset` means a decision
+      already exists, for this project or for the whole plugin: do not ask again
+      and skip the rest of this consent step.
+
+   3. On `unset`, ask once, verbatim:
+
+      > Before we start: may I share anonymous data about this migration with AWS to
```
jkzietz: This does not mention anything about skill invocation metrics

advisor/plugins/aws-startup-advisor/hooks/telemetry/emit.mjs
```diff
@@ -0,0 +1,742 @@
...
+      );
+      return;
+    }
+    case "grant":
+      write("granted");
+      return;
+    case "revoke":
+      // A decline is a decision too: recorded locally so the customer is not
+      // asked again. No event is sent for it.
+      write("revoked");
```
leon1418: [🤖 AI review 🤖] [P1] F1 — Make the documented revoke effective.

With the supported plugin-wide record set to `granted`, `consent revoke` writes only the project file and prints `revoked`. `consent get` still returns `granted`, and the next hook posts the new phase because the machine record wins. I reproduced this through the real CLI against localhost. The new Discover/interpreter instructions promise that this command stops sharing for the project.

Make that revocation an effective deny while preserving a plugin-wide deny, and test machine grant → project revoke → get/status plus both hook entry paths: the effective result must be revoked and no request may be sent.

advisor/plugins/aws-startup-advisor/hooks/telemetry/emit.mjs
```diff
@@ -0,0 +1,742 @@
...
+    // malformed one falls back to the snapshot's, then to a fresh mint persisted
+    // in the snapshot. The emitter never writes the skill's own state file.
+    const runId = asUuid(status.run_id) ?? asUuid(snapshot?.runId) ?? crypto.randomUUID();
+    const validSessionId = asUuid(sessionId);
+
+    // Consent covers what happens from the moment it was given. A run whose
+    // state was last written before the consent record exists is history the
+    // customer never agreed to report: record it as already known and send
+    // nothing, so only transitions from here on are reported.
+    if (!snapshot && predatesConsent(runDir, statusFile)) {
```
leon1418: [🤖 AI review 🤖] [P2] F2 — Rebaseline across a new consent grant.

A previously reported run already has a snapshot. After `revoke`, complete another phase, then explicitly `grant` again: this `!snapshot` guard skips the consent cutoff and the next reconcile posts the phase completed during revocation. Both independent fixtures reproduced this with the state mtime well outside the two-second slack.

Track the effective consent epoch or rebaseline existing snapshots on a new grant. The revoked interval must stay unreported, while a subsequent consented transition should emit once under the original run ID.

advisor/plugins/aws-startup-advisor/hooks/telemetry/emit.mjs
```diff
@@ -0,0 +1,742 @@
...
+        phases: status.phases ?? {},
+        completed: status.current_phase === "complete",
+        via: viaMode(),
+        updatedAt: new Date().toISOString(),
+      });
+      return;
+    }
+
+    const events = diffEvents(status, snapshot);
+    if (events.length === 0) return;
```
leon1418: [🤖 AI review 🤖] [P2] F3 — Persist observed state even when there is no event.

The interpreter explicitly permits confirmed re-entry/reset. In `completed → in_progress → completed`, this early return leaves `completed` in the snapshot after the middle hook, so the second completion disappears. It also leaves session ownership stale: after session B writes `in_progress` to a run last reported by A, B can finish via a shell write and have SessionEnd skip the run as A-owned. I reproduced both with the real emitter, including an explicit B post-write payload.

Record observed resets and active-run session tracking independently of emission. Test re-entry emits exactly one new completion and resumed teardown flushes its last transition, while unrelated sessions and selectively held HTTP events remain protected.

advisor/plugins/aws-startup-advisor/hooks/telemetry/emit.mjs
```diff
@@ -0,0 +1,742 @@
...
+      attributes.hasDatabase = DB_TYPE.test(types);
+      attributes.hasAi = AI_TYPE.test(types) || aiFromProfile;
+    } else if (aiProfile) {
+      attributes.hasAi = aiFromProfile;
+    }
+  }
+
+  if (phaseEvent && event.phase === "CLARIFY") {
+    const preferences = readJson(path.join(runDir, "preferences.json"));
+    const clarifyMode = mapEnum(CLARIFY_MODE, preferences?.metadata?.migration_type);
```
leon1418: [🤖 AI review 🤖] [P2] F4 — Align the existing attribute adapters with their producers.

Producer-shaped fixtures expose these mismatches in the current adapter:

| Existing producer input | Current output | Required behavior |
| --- | --- | --- |
| GCP `metadata.clarify_mode=wizard`, Heroku `full/fast_path`, AI-only `migration_type=ai-only` | `FULL` for the wizard; others omitted | Map the actual mode/route fields to `WIZARD`, `FULL`, `FAST`, `AI_ONLY` |
| Committed Heroku inventory: `resource_type=addon`, `config.addon_service=heroku-postgresql` | `hasDatabase=false` | Read the typed add-on service; retain a formation-only negative control |
| OpenAI-only delegated Assess profile, no GCP inventory | `sourceProvider=GCP` | Derive source from workload artifacts, separately from owning skill; omit ambiguous values |
| `recommendation.outcome=defer_for_evidence` | Outcome omitted | Map to `DEFER` |
| Billing-only `metadata.pricing_source=cached` | Pricing source omitted | Read the billing metadata path |
| GCP `current_costs.source=preferences` | `spendBasis=DEFAULTED` | Preserve the user-stated spend-band provenance defined by `estimate-infra.md` Part 1 |

These are current fields in the committed producers. Correct the emitter adapters and add producer-shaped regression cases, retaining omission for genuinely unknown values.

advisor/plugins/aws-startup-advisor/hooks/telemetry/emit.mjs
```diff
@@ -0,0 +1,742 @@
...
+    pluginVersion: ctx.pluginVersion,
+    occurredAt: Date.now(),
+    pluginTelemetryEvent: { migrationActivity },
+  };
+}
+
+// Resolves to the HTTP status; rejects on a network failure or timeout.
+async function post(endpoint, body) {
+  const controller = new AbortController();
+  const timer = setTimeout(() => controller.abort(), POST_TIMEOUT_MS);
```
leon1418: [🤖 AI review 🤖] [P2] F5 — Fit the whole teardown into the effective host budget.

The current [Claude Code SessionEnd contract](https://code.claude.com/docs/en/hooks#sessionend) gives an overall default budget of 1.5 seconds and explicitly says plugin-provided timeouts do not raise it. Therefore the plugin registration’s `timeout: 10` does not protect this three-second request wait.

In a local simulation with responses delayed to 1.9 seconds, cancellation at 1.5 seconds leaves received requests without a snapshot and leaves the lock behind. A later reconcile after lock expiry repeats the requests. This is a documented-contract/local-cancellation reproduction, not a live-host observation.

Bound the complete SessionEnd sweep—including requests, snapshot writes and cleanup—to the supported deadline, and cover delayed responses plus multiple eligible runs. Keep the ordinary Stop budget separate.

leon1418 (changes requested):
[🤖 AI review 🤖]

Reviewed `c64fc241947e265960824a8205cdbc8b1060abba` against `6b48f8ad84fdd3928de41bf116e1e0f3a181cc6c` with OCR delegation, one fresh independent general reviewer, repository contract review, and verified local probes. All 13 changed files and affected unchanged consumers were covered.

Please address this consolidated correction list:

1. **F1 — Effective revocation (P1):** a supported machine grant must not defeat the documented project stop-sharing command.
2. **F2 — Consent epochs (P2):** regrant must not backfill phases completed during revocation.
3. **F3 — Snapshot/lifecycle tracking (P2):** observed resets must allow a genuine second completion; resumed-session teardown must retain the correct active run.
4. **F4 — Attribute adapters (P2):** correct clarification mode, Heroku database detection, AI-only source provider, deferred outcome, billing pricing provenance, and user-provided spend provenance using the existing producer contracts.
5. **F5 — Teardown budget (P2):** finish the complete SessionEnd sweep and persistence within the host’s effective deadline.

Implementation corrections belong in `hooks/telemetry/emit.mjs`, with regressions in `tests/hooks/telemetry-emitter.test.ts` under the Advisor plugin; F5 also concerns `hooks/hooks.json`. The detailed triggers and acceptance criteria are inline.

Validation: 14 committed telemetry tests passed under Node 24.14.1; both vendoring checks, cross-plugin drift, relevant frontmatter validation, and `git diff --check` passed. Selective 403/429/5xx replay behaved as intended. Additional real-emitter localhost fixtures reproduced the findings. Host deadline tests simulate the documented budget; no live customer telemetry, live IDE session, or production-service execution was performed.

advisor/plugins/aws-startup-advisor/hooks/telemetry/emit.mjs
az-zhu: Follow-up: 403, 429 and 5xx are now held per event. A held event keeps its previous snapshot state and is re-sent alone on the next trigger; accepted events in the same batch are never re-sent.

advisor/plugins/aws-startup-advisor/hooks/telemetry/emit.mjs
az-zhu: Correct: nothing writes the machine-wide file yet; it is the slot for the plugin-level prompt once that ships. Precedence is now deny-wins across all records (any `revoked` wins, then the earliest `granted`), so a project `consent revoke` also holds under a future plugin-wide grant, and `consent get/status` report the effective record.

advisor/plugins/aws-startup-advisor/skills/agent-advisor/references/vendored/dsl/INTERPRETER.md
az-zhu: Agreed it only covers migration data. The wording is with Legal via Marvin's request (9/23), which already carries the combined migration + skill-invocation text, so I am holding this line until Legal settles it. If Legal picks install-time disclosure, this `_init` step goes away entirely.

advisor/plugins/aws-startup-advisor/hooks/telemetry/emit.mjs
az-zhu: Fixed. Consent is now deny-wins across every record: any `revoked` wins, otherwise the earliest `granted` decides, and `consent get/status` report that effective record. Test covers machine grant → project `consent revoke` → `get` and the hook path (nothing sent, no snapshot).

advisor/plugins/aws-startup-advisor/hooks/telemetry/emit.mjs
az-zhu: Fixed. The pre-consent baseline now applies whether or not a snapshot exists, measured against the effective grant, so phases completed while revoked are recorded as known and never sent after a re-grant; the next consented transition emits once under the original run id. Test added.

advisor/plugins/aws-startup-advisor/hooks/telemetry/emit.mjs
az-zhu: Fixed. The snapshot now mirrors observed state even when nothing is sent (written only when the observation changed), so a confirmed re-entry reset followed by re-completion emits exactly one new PHASE_COMPLETED, and ownership passes to the session that last changed the run so its teardown reports the final transition. Two tests added (re-entry, session hand-over).

advisor/plugins/aws-startup-advisor/hooks/telemetry/emit.mjs
az-zhu: Verified against the producers and fixed 5 of the 6 rows: `metadata.clarify_mode` (wizard/full/fast_path) with `migration_type: ai-only` → AI_ONLY; Heroku `config.addon_service` feeds hasDatabase; `defer_for_evidence` → DEFER; billing `metadata.pricing_source`; `current_costs.source: preferences` → USER_PROVIDED (it is the customer's stated band). `sourceProvider` is now derived from the run's artifacts (inventory, else `ai_source`) instead of the owning skill, so an AI-only run reports OPENAI/ANTHROPIC, `both` is omitted, and RUN_STARTED carries no provider until Discover has written something. Two producer-shaped tests added.

advisor/plugins/aws-startup-advisor/hooks/telemetry/emit.mjs
az-zhu: Confirmed in the Claude Code docs: SessionEnd hooks share 1.5 s and a plugin hook's `timeout` does not raise it. The sweep now bounds every request to what is left of that budget (minus room for the snapshot write) and leaves runs it cannot reach untouched for the next trigger; a request cut short advances the snapshot, so nothing is repeated. Removed the misleading `timeout: 10` from hooks.json. Test runs the sweep against a 1.9 s-slow stub and checks the lock, the snapshot and no re-send.

PR #288, opened 2026-09-12
feat(heroku-to-aws): make report validation blocking + harden it

herosjourney:
## Problem

Heroku migrations produce a `migration-report.html` — the summary a stakeholder reads. There's a checker that's supposed to make sure that report is complete and sound, but it wasn't actually enforcing anything: it was **optional** (a bad report didn't stop anything), it had **no tests**, and it only looked at a handful of the report's sections. On top of that, two of the skill's own instruction files disagreed about whether the check even blocks — one said "optional, never halt," the other said "fail the gate."

The practical result: a Heroku user could ship a report that's missing its required cost-optimization section, or shows a verdict signalled by color alone (a problem for colorblind readers), or has an empty section where content should be — and nothing would catch it. This PR makes the check actually block, hardens what it inspects, gives it a test suite, and resolves the contradiction so the two files tell one story.

`validate-heroku-migration-report.py` was optional/non-blocking, had zero tests, and checked only three section IDs + the footer. `generate-report.md` said "optional / never-halt" while `generate-assemble.md` said "GATE_FAIL" — a direct contradiction. A report could ship missing `cost-optimization`, carrying a color-only verdict pill, or with an empty section, with no gate. Review item **P1-B**.
</details>

## Solution

Harden the Heroku report validator, make it blocking, and add the test suite it never had —
scoped to what the intentionally-thin one-pager actually emits.

**Validator (`validate-heroku-migration-report.py`), Python-stdlib only:**
- Requires `cost-optimization` **and** enforces it is **non-empty** (a table or the explicit
  no-eligible-commitment sentence, never a blank section — matching the skill's Step 3 contract).
- **Typography-first verdict:** `decision-summary` must **not contain** any `badge-verdict-*`
  pill; when `estimation-infra.json` declares `recommendation.outcome`, it must render a
  `verdict-headline`.
- **Accessibility (safe subset):** `<html lang>`; every `<th>` declares `scope=col|row`; any
  `<figure>` carries `aria-label` + `<figcaption>`.

**Emitter and gate now agree (a11y).** `generate-report.md` was updated so the report *template*
instructs the a11y attributes the validator requires — every `<th scope>`, and figure labels if a
figure is emitted. Without this the blocking gate would have false-failed a report that followed
the template (the tables emit `<th>` with no scope). `<html lang>` was already in the skeleton.

**Where the validator runs (round-2 fix — mirrors #287).** Generate dispatches under
`_exec._agent: rw`, which excludes shell/Bash — the dispatched worker (fragments **and**
assembler) cannot run the validator. An earlier revision put the `python3` invocation in
`generate-assemble.md` and called it "the main window," but the assembler is dispatched, so
"blocking" was prose on a path that can't execute. Corrected to the three-step split:

- **Worker (assembler + `generate-report.md`, dispatched, no shell):** authors
  `migration-report.html` and does the in-fragment Step 3 self-check + one-repair; it never runs
  the validator.
- **"Finish Generate" (main window, BEFORE the gate):** `generate.md` runs
  `validate-heroku-migration-report.py` (`$PLUGIN_ROOT`-anchored) over the report. On `REPORT_OK`
  it **stamps a durable `report-validation-status.json`** (temp status sidecar, not a `_produces`
  artifact); on `REPORT_FAIL` it emits `GATE_FAIL` **and pastes the validator's `errors[]`
  verbatim** so recovery is actionable. This step edits nothing else.
- **`_postconditions` gate (main window, READ-ONLY):** the section-id `_assert` plus a new
  `_assert` that `report-validation-status.json` has `report_status == REPORT_OK`. Runs no
  validator, edits nothing (per `INTERPRETER.md` § `_postconditions`). The durable stamp means the
  assert reads an artifact, not conversation memory — parity with #287's `validation-report.json`.

**Recovery** on `GATE_FAIL` is a hand-edit of the report from the pasted errors + a direct
validator re-run (or a maintainer re-running Generate for a clean rebuild) — not a blind "re-run
Generate," which re-dispatches the shell-less worker and re-authors the report.

**Blocking wiring resolved the contradiction.** `generate-report.md` and `generate-assemble.md`
now agree the gate is required/blocking and read-only; `generate.md` `_postconditions` gained
`cost-optimization` in the section-id `_assert` plus the durable `report-validation-status.json`
`REPORT_OK` `_assert`.

**Tests:** `tests/test_validate_heroku_migration_report.py` (both trees, **14 cases**) — compliant
report (with and without a declared recommendation) passes; failures fire for missing/empty
`cost-optimization`, missing footer, `badge-verdict-*` pill, `<th>` without `scope`, missing
`<html lang>`, missing `verdict-headline` when an outcome is declared, figure missing `aria-label`,
and (new) figure with `aria-label` but no `<figcaption>`; plus the figure-happy-path and
≥2-scenario paths.

**Deliberately NOT ported** (would false-fail the thin report): decision-before-TOC, single-`<h1>`,
per-table `<caption>`, glossary/appendix set.

## Type of Change

- [x] Bug fix
- [x] Guardrail/CI update

## Team Folder

- [x] `advisor/`
- [x] `migrate/`
- [ ] Other: ___

## Verification

- `cross-plugin-drift.ts`: **OK** (273 identical, 25 allowlisted — no new drift)
- `test_validate_heroku_migration_report.py` (both trees): **13 passed** each
- `frontmatter-validator` (both trees): OK
- `markdownlint-cli2`: 0 errors; `dprint check`: clean; `tsc --noEmit`: OK both trees
- False-fail probe: a template-shaped report (with `<th scope>`, non-empty cost-optimization) →
  exit 0; a bare `<th>` → exit 1 with the actionable message (gate matches emitter)
- `mise run build` (full composite incl. `bandit`/`semgrep`/`gitleaks`/`checkov`/`grype`):
  **PASS, exit 0** — run against a fresh detached worktree at this branch's commit (not my working
  checkout, which has an untracked, unrelated `migration-watchdog/` directory failing an unrelated
  markdownlint check). checkov 276/0; gitleaks no leaks; grype no vulnerabilities.

Not run locally: no browser a11y tooling (Vale/Pa11y/axe) — the validator is a dependency-free
static HTML reader, matching the plan's "Python 3 only" constraint.

## Checklist

- [x] I have updated documentation if needed (`generate-report.md` a11y authoring rules + blocking wiring; `generate-assemble.md` main-window gate)


herosjourney:
## Follow-up review + fixes (pushed as `b8f4260`)

A review pass over this PR found the blocking flip had real problems — a gate stricter than the
emitter, the same interpreter conflicts as the tf-policy-gate PR, and test gaps. All fixed.
Structured by what the review found and how it was addressed:

### Blocking issues found

**1. The a11y rules would false-fail a report that follows the template.** The validator required
`scope=col|row` on every `<th>` and `aria-label`+`<figcaption>` on every `<figure>`, but
`generate-report.md` never told the agent to emit them — and the report emits tables
(`exec-costs`, `cost-optimization`, `what-if-scenarios`). An agent following the template would
produce `<th>Tier</th>` with no scope and the now-blocking validator would `REPORT_FAIL`. Same
"gate stricter than emitter" mistake I'd consciously avoided for `<caption>`.
→ **Fix:** added the a11y requirements to the emitter (`generate-report.md` Step 2 skeleton + the
Step 3 self-check): every `<th>` declares `scope`, any `<figure>` carries `aria-label` +
`<figcaption>`. Now the gate matches what the template produces — verified with a probe: a
template-shaped report passes, a bare `<th>` fails.

**2. The completion gate said "GATE_FAIL … and repair," which the interpreter forbids.**
`INTERPRETER.md` § `_postconditions`: a gate must not modify artifacts. And the paragraph above
the new blocking gate still said "repair the HTML and continue," reading as a second
continue-on-fail path.
→ **Fix:** the one-repair-then-rewrite lives only in the `generate-report.md` in-fragment Step 3
self-check; the assembler now emits `GATE_FAIL` and **halts** (tells the user to re-run Generate),
editing nothing. Step 3 was renamed and split into "in-fragment self-check (repair here)" vs
"blocking validation (main window)."

**3. The required validator couldn't run under `_exec: rw`, and the path was cwd-relative.** Same
host gap as the tf-policy-gate PR — the dispatched worker has no shell, so a "required, blocking"
validator invoked in the fragment is just prose. And `generate-report.md` used
`python3 scripts/…` (cwd-relative, misses the plugin script even on inline hosts).
→ **Fix:** the validator runs in the **main window** completion gate (`generate-assemble.md` +
`generate.md` `_postconditions`, never dispatched — `INTERPRETER.md` § `_exec` step 4), the only
place with a shell; the worker only writes the HTML. Path corrected to
`"$PLUGIN_ROOT/scripts/validate-heroku-migration-report.py"`.

### Should-fix issues found

**4. Docstring vs code on `badge-verdict-*`.** The docstring said "as the sole verdict carrier" but
the code (correctly) fails on *any* `badge-verdict-*`. Three sentences disagreed.
→ **Fix:** aligned docstring, code comment, and `generate-report.md` Step 3 to "must not contain
`badge-verdict-*`"; Step 3 now also requires a `verdict-headline` when `recommendation.outcome` is
set, matching the validator.

**5. Test gaps for rules the PR made blocking.** No figure-a11y test, no happy-path test with a
declared `recommendation.outcome`, and empty `cost-optimization` passed (only the id was checked).
→ **Fix:** the validator now enforces **non-empty** `cost-optimization`; added tests for
figure-without-aria-label (fails), figure-with-label (passes), good-report-with-recommendation
(passes), and empty-cost-optimization (fails). 9 → 13 cases. Renamed Step 3 from "Soft gate."

**6. Corrupt `estimation-infra.json` fails open silently.**
→ **Fix:** added a one-line comment documenting the intentional fail-open-on-ambiguity.

### Verification (`b8f4260`, fresh detached worktree)

`mise run build` PASS (checkov 276/0, gitleaks no leaks, grype no vulns), `cross-plugin-drift.ts`
OK (273 identical), `test_validate_heroku_migration_report.py` 13 passed both trees, all touched
files twin-identical, false-fail probe confirms the gate matches the emitter. Not run locally:
browser a11y tooling — the validator is a dependency-free static reader.

herosjourney:
Verified additional issues and each point against the code at `b8f4260` before changing anything — all four hold, and #1 is the same host-gap as the first #287 pass, one file over. Fixed in `2868f78` (both trees, byte-identical). Structured as **finding → what I changed**.

### 1. Should-fix — validator was still in the worker, not the main window (agreed, blocking)
Confirmed: the `python3 validate-heroku-migration-report.py` call lived only in `generate-report.md` (fragment) and `generate-assemble.md` (assembler) — both dispatched to the shell-less `rw` worker. `generate.md` `_postconditions` only asserted section-ids + footer *prose*; it never ran the script or asserted `REPORT_OK`. So on `_exec: rw` the new checks (no `badge-verdict-*`, `verdict-headline`, `<th scope>`, non-empty `cost-optimization`, `<html lang>`) never executed, and a report with the four section-ids + footer passed the only main-window check. The assembler even labelled itself "the main window," which is wrong — the assembler is dispatched. Mirrored the #287 split exactly:

- **`generate.md`** — new main-window **"Finish Generate in the main window (report validation)"** step that runs `validate-heroku-migration-report.py` **after the worker returns and before** `_postconditions`. Added a read-only `_assert` that this step printed `REPORT_OK` (not just that the section-ids exist). Scope Boundary carves the main-window validate out of "Nothing else."
- **`_postconditions`** is read-only: the section-id `_assert` + the `REPORT_OK` `_assert`, no script run, no edits.
- **`generate-assemble.md`** — dropped the false "In the main window … run the report validator" claim; it now states the assembler is dispatched and points at the finish step; halts read-only on `GATE_FAIL`.
- **`generate-report.md`** — points at the finish step for authoritative execution.

### 2. Nit — GATE_FAIL should quote the validator errors (agreed)
Right — the fragment has no shell to re-run the validator, so a bare "re-run Generate" is not actionable. The finish step now **emits `GATE_FAIL` and pastes the validator's `errors[]` verbatim** (missing scope id, empty `cost-optimization`, banned pill class, `<th>` without `scope`, missing `<html lang>`, etc.). I also fixed the recovery copy the same way #287 was fixed: recovery is a hand-edit from the pasted errors + a **direct** validator re-run, not a full Generate re-dispatch (which re-authors the report under the shell-less worker and wipes the fix).

### 3. Nit — skeleton had no `<th scope="col">` example (agreed)
The HTML skeleton showed `<section id="exec-costs">…</section>`. Agents copy skeletons, so `<html lang="en">` got followed but `<th scope>` might not. The `exec-costs` skeleton section now shows a sample table with `<th scope="col">` header cells and a comment to copy the pattern for `cost-optimization` / `what-if-scenarios`.

### 4. Nit — no test for figure-without-`<figcaption>` (agreed)
Added `test_figure_without_figcaption_fails` — a `<figure aria-label="…">` with no `<figcaption>` now asserts exit 1 + a `figcaption` error, symmetric to the existing missing-`aria-label` case. Test count 13 → 14 (both trees).

### Verification (`2868f78`, both trees)
- `cross-plugin-drift.ts`: **OK** (273 identical, 25 allowlisted)
- `test_validate_heroku_migration_report.py`: **14 passed** advisor, **14 passed** migrate
- `mise run build` against a fresh detached worktree at this commit: **PASS, exit 0** (checkov 276/0, gitleaks no leaks, grype no vulnerabilities, `fail 0` across DSL suites)
- Fail-closed preserved: if the finish step can't run the validator (no shell), it does not record `REPORT_OK`, so the read-only `_assert` fails closed.

No change to the a11y/emitter alignment, the assembler-no-longer-edits-HTML property, the `$PLUGIN_ROOT` path, the `badge-verdict-*` wording, or the not-ported set — those stay as you approved.

herosjourney:
Fixed 3 additional nits.

### 1. `REPORT_OK` was conversation-only (agreed — added parity)
Correct: #287 stamps a durable `policy_status` into `validation-report.json`, but #288's finish step only "recorded" a print and the `_assert` read prose. Since the validator only emits to stdout/stderr, I added the one-line stamp you suggested: the Finish Generate step now writes `report-validation-status.json` (`{"report_status": "REPORT_OK"}`, or `REPORT_FAIL` with `errors[]`), and the read-only `_postconditions` `_assert` now checks that file's `report_status == REPORT_OK`. So the assert reads an artifact, not conversation memory — parity with #287. It's a temp status sidecar (not in `_produces`), and if the validator can't run the stamp is absent/`not_run`, so it still fails closed. `generate-report.md` and `generate-assemble.md` reference the stamp consistently, and recovery now says re-run + re-stamp.

### 2. Assembler header was stale (agreed)
Right — the opener still said the assembler "enforces the completion handoff gate." Rewrote it: the assembler runs the cross-artifact checks and emits the handoff signal, but the **report validator runs later in the main-window "Finish Generate" step** (the assembler is dispatched and has no shell). Body was already correct; this aligns the opener.

### 3. Two `GATE_FAIL` emitters (agreed it's redundant — leaving it)
Agreed it's fine and fail-closed. The redundancy is deliberate defense-in-depth: the finish step emits `GATE_FAIL` on `REPORT_FAIL`, and the read-only `_postconditions` `_assert` independently blocks on a non-`REPORT_OK` stamp — so a skipped/half-run finish step still can't complete. I'd rather keep both than rely on either alone, so no change here.

### Verification (`2bce3f5`, both trees)
- `cross-plugin-drift.ts`: **OK** (273 identical, 25 allowlisted)
- `test_validate_heroku_migration_report.py`: **14 passed** (both trees)
- `mise run build` against a fresh detached worktree at this commit: **PASS, exit 0** (checkov 276/0, gitleaks no leaks, grype no vulnerabilities)

No validator-code or test change in this revision — it's the durable-stamp wiring + the assembler-opener wording.

advisor/plugins/aws-startup-advisor/skills/heroku-to-aws/references/phases/generate/generate.md
```diff
@@ -118,4 +120,47 @@ FORBIDDEN — Do NOT include ANY of:
...
+This is **leftover Generate work**, not a gate check: the dispatched `_exec._agent: rw` worker
+that assembled the report has no shell (`INTERPRETER.md` § capability tiers — `rw` excludes Bash),
+so it could not run the report validator. The interpreter finishes that work in the MAIN window
+(the only place with a shell), **after the worker returns and before running `_postconditions`**.
+
+1. Run the report validator (**required, blocking**):
+
+   ```
+   python3 "$PLUGIN_ROOT/scripts/validate-heroku-migration-report.py" \
+     "$MIGRATION_DIR/migration-report.html" --migration-dir "$MIGRATION_DIR"
```
leon1418: [🤖 AI review 🤖] **[P1] Ship the mandatory validator with standalone skills**

The documented `npx skills add ... --skill heroku-to-aws --agent <agent>` installation in `advisor/README.md` copies the skill directory, but this script lives outside it under the plugin-level `scripts/` directory. Making this invocation mandatory therefore blocks Generate for supported skill-only installations, even when Python and a valid report are available. A skill-directory packaging probe contained no validator, and invoking the expected script exited 2 with “No such file or directory.” Please bundle the validator inside the installed skill, resolve its path from the skill root, and cover that installation layout in a regression test. Apply the corresponding change in both plugin trees.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
```diff
@@ -54,6 +171,19 @@ def validate(html: str, migration_dir: Path | None) -> list[str]:
     if "draft for review" not in html.lower():
         errors.append('footer must contain "draft for review" disclaimer')
 
+    # cost-optimization must not be a blank section (generate-report.md Step 3 item 6:
+    # a table, or the explicit no-eligible-commitment sentence — never empty).
+    if counts.get("cost-optimization", 0) >= 1:
+        body = _section_html(html, "cost-optimization") or ""
+        if not re.sub(r"<[^>]+>", "", body).strip():
```
leon1418: [🤖 AI review 🤖] **[P2] Require substantive cost-optimization content**

Stripping tags and checking for any remaining text accepts `<section id="cost-optimization"><h2>Cost Optimization Opportunities</h2></section>`, as well as a table containing only column headers and an empty `<tbody>`. Both probes returned exit 0 / `REPORT_OK` in both plugin copies, including with a populated `optimization_opportunities` input. Neither satisfies the documented opportunities-table-or-no-eligible-commitment contract. Please require populated opportunity rows or the explicit no-eligible statement, excluding headings and table headers from the content check, and add these cases to the tests.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
```diff
@@ -40,6 +56,107 @@ def _section_counts(html: str) -> dict[str, int]:
...
+        errors.append("accessibility: <html> must declare a valid lang attribute")
+
+    body = re.sub(r"<style\b.*?</style>", "", html, flags=re.IGNORECASE | re.DOTALL)
+
+    for index, table_match in enumerate(
+        re.finditer(r"<table\b[^>]*>.*?</table>", body, re.IGNORECASE | re.DOTALL), 1
+    ):
+        table = table_match.group(0)
+        for th in re.findall(r"<th\b[^>]*>", table, re.IGNORECASE):
+            if not re.search(r"\bscope=[\"'](?:col|row)[\"']", th, re.IGNORECASE):
```
leon1418: [🤖 AI review 🤖] **[P2] Accept valid HTML attribute syntax**

Replacing the passing fixture's `<th scope="col">` with either `<th scope = "col">` or `<th scope=col>` makes the validator exit 1 in both trees, although both headers declare the same valid scope. The `lang` and `aria-label` regexes have the same quoted-value/no-whitespace restriction. Since this validator now blocks Generate, valid HTML formatting can prevent completion. Please parse the attributes (the standard-library `HTMLParser` normalizes these forms), or otherwise accept equivalent legal syntax, and add coverage for these variants in both copies.

leon1418 (changes requested):
[🤖 AI review 🤖]

Reviewed head `d35f4b8bb65d673d541ed03ab9f7f0346abc32d0`, including the follow-up explanations and both plugin copies. Requesting changes for the three inline findings:

- **P1:** The newly mandatory validator is not included in the documented skill-only installation.
- **P2:** Heading-only cost-optimization sections and tables without data rows still pass.
- **P2:** The blocking accessibility checks reject valid HTML attribute syntax.

Validation: all 28 added tests passed across both trees; all five changed mirror pairs are byte-identical; cross-plugin drift and both Heroku frontmatter checks passed; current GitHub checks are green. Additional probes reproduced the validator gaps in both copies, and a skill-directory packaging probe confirmed the missing script. These cases are not covered by the added tests.

advisor/plugins/aws-startup-advisor/skills/heroku-to-aws/references/phases/generate/generate.md
herosjourney: Confirmed with a real install: checked a standalone `--skill heroku-to-aws` install on disk (predates this PR) and there is no `scripts/` directory and no validator file anywhere under the installed skill. $PLUGIN_ROOT is also never defined inside the skill itself. Making this invocation mandatory does block Generate for that install path. Will bundle the validator inside the skill directory, resolve its path relative to the skill root instead of $PLUGIN_ROOT, and add a regression test covering that layout. Both trees.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
herosjourney: Confirmed via the full CLI, not just the isolated check: a heading-only `<section id="cost-optimization"><h2>Cost Optimization Opportunities</h2></section>` and a table with header cells but an empty `<tbody>` both return `REPORT_OK`, including with a populated `optimization_opportunities` input. A control case with a real opportunity row correctly requires the section's other structural rules (e.g. `<caption>`) once those are met. Will exclude headings and table headers from the content check and require populated rows or the explicit no-eligible-commitment statement, with tests. Both trees.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
herosjourney: Confirmed. `<th scope="col">` passes; `<th scope = "col">` (spaces around `=`) and `<th scope=col>` (unquoted) both fail the same validator that's now blocking Generate, even though all three are valid, equivalent HTML. Will parse attributes with the standard-library HTMLParser (or an equivalent normalizer) instead of a literal-syntax regex, and add these variants as regression cases for scope, lang, and aria-label. Both trees.

herosjourney:
Thanks — reproduced all three before replying:

1. **Validator doesn't ship with a standalone skill install.** Checked a real `--skill heroku-to-aws` install on disk: no `scripts/` directory, no validator file anywhere under it, and `$PLUGIN_ROOT` is never defined inside the skill. This is a real blocker for that documented install path, not a hypothetical.
2. **Cost-optimization content check accepts decorative-only markup.** Confirmed through the full CLI: a heading-only section and a table with headers but an empty `<tbody>` both pass `REPORT_OK`; a control case with an actual opportunity row behaves correctly once the section's other structural rules are met.
3. **Attribute regexes reject valid HTML syntax.** Confirmed: `scope="col"` passes, `scope = "col"` and `scope=col` (both valid, equivalent HTML) fail the now-blocking validator.

All three will be fixed in both trees: bundle the validator inside the skill and resolve its path from the skill root, exclude headings/table headers from the content check and require real rows or the explicit no-eligible statement, and parse attributes properly instead of matching literal syntax. Replied inline on each with the specific fix.

advisor/plugins/aws-startup-advisor/skills/heroku-to-aws/references/phases/generate/generate.md
leon1418: [🤖 AI review 🤖] Verified the packaging fix at `8ddd56436b48e1c5d166e9f6bb1d7d150b916188` in both plugin copies. The validator now lives under `skills/heroku-to-aws/scripts/`, and `SKILL.md` plus both invocation sites resolve it from `<SKILL_BASE>`. I copied each skill tree into an isolated skill-only layout with no plugin-level scripts and ran its packaged validator from an unrelated working directory; both returned exit 0 / `REPORT_OK`. This P1 finding is addressed.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
leon1418: [🤖 AI review 🤖] **[P2] The content check still accepts an empty table with the template's required note.**

Rechecked `8ddd56436b48e1c5d166e9f6bb1d7d150b916188` in both copies. Heading-only sections and bare header-only tables now fail as intended. However, an empty `<tbody>` followed by the exact credits disclaimer required by `generate-report.md` lines 114–115 still returns exit 0 / `REPORT_OK`, including with a populated `optimization_opportunities` input:

```html
<table><thead><tr><th scope="col">Optimization</th></tr></thead><tbody></tbody></table>
<p>Activate credits don't cover RI/Savings Plan upfront costs — they apply only to the ongoing discounted hourly rate.</p>
```

In the relocated `skills/heroku-to-aws/scripts/validate-heroku-migration-report.py`, `_validate_optimization_content()` still accepts any text left after stripping headings and headers (lines 127–132). Please require a populated opportunity data row or the explicit no-eligible-commitment statement, rather than letting explanatory prose satisfy the requirement. Add this template-shaped empty-table regression in both copies. The original content finding remains open.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
leon1418: [🤖 AI review 🤖] **[P2] Apply attribute normalization to the verdict checks too.**

Verified `8ddd56436b48e1c5d166e9f6bb1d7d150b916188` in both copies: the reported `scope`, `lang`, and `aria-label` variants now pass, while missing scope/figure labels still fail. The same formatting-dependent behavior remains in `_validate_verdict()` in the relocated `skills/heroku-to-aws/scripts/validate-heroku-migration-report.py` (lines 150–171): with `recommendation.outcome` declared, changing only `class="verdict-headline"` to `class = "verdict-headline"` or `class=verdict-headline` makes an otherwise passing report fail with “has no verdict-headline element.” Conversely, adding `<span class=badge-verdict-go>Go</span>` alongside the normal headline returns `REPORT_OK`, bypassing the badge ban.

Please use parsed, tokenized `class` attributes for both verdict checks, and add these acceptance/rejection cases in both mirrors. The accessibility-specific fixes are confirmed; this related gap still lets equivalent HTML change whether Generate completes.

leon1418 (changes requested):
[🤖 AI review 🤖]

Re-reviewed exact head `8ddd56436b48e1c5d166e9f6bb1d7d150b916188` against base `0f067dfe8b4da3f9cca969f1c2022a75b4b7f154`, including all author replies, the full PR diff, both plugin trees, and the Generate worker/main-window handoff.

The standalone validator packaging P1 is fixed. I replied in all three original threads without creating duplicate findings. Two P2 concerns remain:

- A table with no opportunity rows still passes when accompanied by the credits disclaimer that the report template requires.
- The accessibility attributes are normalized, but verdict `class` attributes still produce false failures for valid syntax and miss an unquoted banned badge.

Local validation: **46/46 tests passed**; isolated skill-only validator execution passed for both copies; all six changed surviving mirror pairs are byte-identical and the old script is removed in both trees; cross-plugin drift (**277 identical, 25 allowlisted**), both Heroku frontmatter checks, both vendored-shared checks, and `git diff --check` passed. I also inspected the Heroku allowlisted interpreter/schema differences and ran 40 targeted CLI probes; the remaining failures are reproduced in both copies. No full agent-driven migration replay was run.

Current GitHub CI is green, separately from those local results. Retaining **Request changes** until the two remaining validator gaps are addressed.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
herosjourney: Confirmed, and fixed in `4dddc4a1`, both trees.

The root cause was the check being "any leftover text after stripping headings/`<th>`" rather than a positive check for real content. That let column-header text, the credits disclaimer, or any other prose satisfy it even with an empty `<tbody>`.

Replaced `_validate_optimization_content()` with a positive test: pass only if either (a) a `<tbody>` row has a `<td>` with non-empty text (a real opportunity row), or (b) the exact no-eligible-commitment sentence ("no 1-year/3-year commitment product applies to this architecture") appears in the section text. Nothing else — headings, `<th>` text, disclaimer prose, or any other filler — can satisfy it.

Added regression tests for your exact repro (`test_empty_tbody_with_credits_disclaimer_fails`) and a symmetric case (`test_th_only_row_in_tbody_does_not_count_as_opportunity_row`, a `<tbody>` row made only of `<th>` cells). Test count 29 in both trees; the existing `test_cost_optimization_with_real_opportunity_row_passes` control case still passes.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
herosjourney: Confirmed, and fixed in `4dddc4a1`, both trees.

Both `_validate_verdict()` checks (`badge-verdict-*` ban and `verdict-headline` requirement) were still matching `class="..."` with a literal regex, unlike the `<th scope>`/`lang`/`aria-label` checks that already went through the stdlib `HTMLParser`. Added a `_class_tokens()` helper that parses every tag in a fragment via the same `_collect_tags()` machinery and returns the set of whitespace-split `class` tokens, then switched both verdict checks to test set membership instead of regex-matching the raw attribute string. `class="a b"`, `class = "a b"`, and `class=singletoken` all normalize to the same token set now.

Added your two repro cases as regression tests: `test_verdict_headline_with_spaces_around_equals_passes` / `test_verdict_headline_unquoted_passes` (equivalent HTML no longer flips the outcome-declared verdict-headline requirement), and `test_badge_verdict_pill_with_spaces_around_equals_fails` / `test_badge_verdict_pill_unquoted_fails` (a badge alongside a normal headline is caught regardless of attribute formatting — no more bypass via unquoted `class=badge-verdict-go`). Test count 29 in both trees.

herosjourney (commented):
## Follow-up fixes (pushed as `4dddc4a1`)

Both remaining P2s from the last review round are fixed in both plugin trees:

**1. Cost-optimization content check accepted decorative-only markup.** `_validate_optimization_content()` checked for "any text left after stripping headings/`<th>`", so an empty `<tbody>` paired with the credits disclaimer (or any other prose) satisfied it. Replaced with a positive check: pass only on a real `<tbody>` row with non-empty `<td>` text, or the exact no-eligible-commitment sentence. Nothing else counts. Replied inline with the specific repro coverage.

**2. Verdict checks weren't using tokenized `class` attributes.** `_validate_verdict()` still matched `class="..."` literally, unlike the scope/lang/aria-label checks that already went through `HTMLParser`. Added `_class_tokens()` (same `_collect_tags()` machinery) and switched both the `badge-verdict-*` ban and the `verdict-headline` requirement to set-membership checks. Replied inline with the specific repro coverage.

**Also merged current `main`** (this PR's branch had fallen behind — main since gained the `rwx` capability tier + Terraform policy gate on this same phase, and the resulting `generate.md` conflict was resolved by combining both main-window exceptions in the Scope Boundary paragraph: the policy checker's scoped shell during assembly, and the report validator's separate main-window "Finish Generate" step).

### Verification (`4dddc4a1`, both trees)
- `test_validate_heroku_migration_report.py`: **29 passed** (14 → 29; added 6 regression tests for the two P2s plus retained coverage)
- `cross-plugin-drift.ts`: **OK** (283 identical, 27 allowlisted)
- `mise run lint:frontmatter`: OK (both trees)
- `mise run fmt:check` (dprint): clean
- `mise run lint:md` (markdownlint-cli2): 0 errors across 896 files
- Manually reproduced both reviewer repro cases against the pre-fix validator to confirm they failed as described, then confirmed the fix resolves them

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
leon1418: [🤖 AI review 🤖] **[P2] Count parsed, nonblank opportunity cells rather than raw row markup.**

Rechecked `4dddc4a14098ae1e4eb71a46cd8342215e2280ce` in both mirrors. The earlier empty-table-plus-credits-disclaimer repro is fixed. The new `_has_populated_opportunity_row()` still treats raw HTML source as rendered content, though. With the other report sections valid, both of these tables return exit 0 / `REPORT_OK` despite having no visible opportunity data:

```html
<table><tbody><tr><td>&nbsp;</td></tr></tbody></table>
<table><tbody><!-- <tr><td>RDS Reserved Instance</td></tr> --></tbody></table>
```

The first passes because `&nbsp;` (also `&#160;`) is not decoded before `.strip()`; the second passes because the regex searches inside comments. Conversely, a populated `<table><tr><td>RDS Reserved Instance</td></tr></table>` returns exit 1 solely because the source omits the optional `<tbody>` tags.

Please parse actual table rows/cells, ignore comments, decode character references, and require nonblank cell text; accept rows with an implicit tbody as well. Add these acceptance/rejection cases in both copies. This is the remaining gap in the original substantive-content finding (current script lines 125–138).

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
leon1418: [🤖 AI review 🤖] Verified fixed at `4dddc4a14098ae1e4eb71a46cd8342215e2280ce` in both mirrors. I reproduced the previous behavior at `8ddd56436b48e1c5d166e9f6bb1d7d150b916188`, then confirmed the new parser accepts spaced and unquoted `verdict-headline` classes when an outcome is declared and rejects an unquoted `badge-verdict-go` alongside a normal headline. Additional checks passed for multiple class tokens, a headline in a nested section, and rejecting a commented-out headline, a different class token, or a headline outside decision-summary. The prior scope/lang/aria-label cases still behave correctly. This P2 finding is addressed.

leon1418 (changes requested):
[🤖 AI review 🤖]

Re-reviewed exact head `4dddc4a14098ae1e4eb71a46cd8342215e2280ce` against base `b30cb39be02d2eae78cb8c482f4018435d4dc988`, including the author's latest replies, the current diff, both plugin copies, and the merged `rwx` policy-checker/main-window report-validation integration.

The standalone packaging P1 remains fixed. The verdict-class P2 is now fixed, and the earlier credits-disclaimer repro is fixed. I replied in the two existing P2 threads. **One P2 content-validation concern remains:** the new regex row check accepts blank character-reference cells and commented-out rows, while rejecting a populated table whose source omits optional tbody tags. The original content thread contains the reproductions and suggested correction; no duplicate thread was created.

Local validation: **58/58 tests passed**; all eight before/after checks reproduced the earlier bugs and verified the claimed corrections across both mirrors; isolated skill-only validator execution passed in both copies; 64 scoped CLI probes covered original and additional cases. Cross-plugin drift (**283 identical, 27 allowlisted**), both Heroku frontmatter checks, both vendored-shared checks, and `git diff --check` passed. The changed validator/tests/Generate files match across mirrors; the allowlisted SKILL.md difference is the existing advisor-only offers block. No full agent-driven migration replay or deployment was run.

Current GitHub CI is green, separately from those local checks. Retaining **Request changes** for the remaining content-validation gap.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
herosjourney: Fixed and pushed at `ab070aa7`, both trees.

Root cause across all three rounds on this same finding has been the same shape: matching raw HTML source with regex instead of parsing actual structure. This round's three repros all come from that:

1. `&nbsp;`/`&#160;` not decoded before `.strip()` — regex sees literal markup, not rendered content.
2. The regex searches inside comments — no concept of "comment," so a commented-out `<tr>` still matches.
3. Requiring an explicit `<tbody>` — a `<table><tr><td>` with no `<tbody>` is valid HTML (implicit tbody), but the regex only looked for the literal tag.

Replaced `_has_populated_opportunity_row()` with an `HTMLParser` subclass (`_OpportunityRowParser`), following the same pattern already used elsewhere in this file for `_AccessibilityParser` / `_TagAttrCollector`. It tracks table/tr/td depth directly (no tbody dependency), uses `convert_charrefs=True` so entities are decoded before any text is seen, and never sees comment text at all since `HTMLParser` routes comments to `handle_comment`, not `handle_data`. `<th>` cells are still excluded from counting as data.

Verified against your three new repros (all now correctly fail/pass as expected) plus the two previously-fixed cases (heading-only, header-only table, empty-tbody-plus-disclaimer) and the control case (real row with explicit tbody) — all still behave correctly. Added 4 new regression tests in both trees for this round's repros: `test_nbsp_only_opportunity_cell_fails`, `test_numeric_charref_only_opportunity_cell_fails`, `test_commented_out_opportunity_row_fails`, `test_opportunity_row_without_explicit_tbody_passes`. Test count 33 in both trees.

Also merged `origin/main` (was behind after #290) — clean, no conflicts, no code overlap.

leon1418: [🤖 AI review 🤖] **[P2] Finalize populated cells when their end tags are omitted.**

Verified `b62d8c67aeddb7c9fa2fd36e7102a60ca71cb8ab` in both mirrors: the four new regressions now behave correctly, and the earlier packaging, disclaimer, and verdict-class fixes remain intact. One table-boundary case is still rejected:

```html
<table><tr><td>RDS Reserved Instance</tr></table>
```

Placed in an otherwise passing report's cost-optimization section, this returns exit 1 / `REPORT_FAIL` with “has no substantive content”; adding the optional `</td>` changes it to exit 0. The HTML Standard permits a td end tag to be omitted at the end of its parent row or before the next td/th, so populated compact tables should not halt Generate.

`_OpportunityRowParser` sets `found` only on an explicit `handle_endtag("td")` (current lines 164–168). It does not finalize the current cell when the row ends or a new cell starts, and it increments `_td_depth` as though adjacent implicitly closed cells were nested. Please handle those implicit cell boundaries (or recognize nonblank data while a valid table/row/cell is active), clear the cell state at row/table boundaries, and add positive tests with omitted td end tags while retaining the blank/comment regressions. This is the remaining case in the original content-validation thread.

leon1418 (changes requested):
[🤖 AI review 🤖]

Re-reviewed exact head `b62d8c67aeddb7c9fa2fd36e7102a60ca71cb8ab` against base `ade57adaf9d776d18d316d28d3cee3f746582ced`, including the author reply, current diff, both mirrors, and Generate's report-validator integration.

The latest entity-only, commented-row, and implicit-tbody cases are fixed. Packaging, substantive-content controls, and parsed verdict classes retain their earlier fixes. **One P2 remains:** a populated table with a valid omitted td end tag is rejected because the new parser only commits cell content on an explicit `</td>`. I added the reproduction and suggested correction to the original content thread; no duplicate thread was created.

Local validation: **66/66 tests passed**; eight before/after comparisons reproduced and verified the newly claimed fixes in both copies; isolated skill-only validator execution passed in both copies. Of 64 scoped CLI probes, only the omitted-td-ending case remains incorrect (once per mirror). Cross-plugin drift (**283 identical, 27 allowlisted**), both Heroku frontmatter checks, both vendored-shared checks, and `git diff --check` passed. I checked the intentional Heroku SKILL.md/interpreter/schema differences and confirmed the main-window report-validation placement. No full agent-driven migration replay or deployment was run.

Current GitHub CI is green, separately from local validation. **Recommendation: Request changes** for the remaining parser boundary case.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
herosjourney: Fixed and pushed at `e3ca4d23`, both trees.

Confirmed: `_OpportunityRowParser` only registered a populated cell on an explicit `handle_endtag("td")`. The HTML Standard permits omitting a `<td>` end tag immediately before the next `<td>`/`<th>` or before the parent `<tr>`/`<table>` closes — your exact repro (`<table><tr><td>RDS Reserved Instance</tr></table>`) never fires that explicit close at all, so real content silently registered as empty.

Replaced the depth-counter approach with a single `_close_cell()` finalization step, invoked whenever a new `<td>`/`<th>` starts (closing any sibling cell still open in the row), when a `<tr>` opens (closing any cell that leaked from an earlier row), and on every actual close of `<td>`/`<th>`/`<tr>`/`<table>` — matching how a real HTML parser handles implicit cell closure.

Verified against your exact repro plus adjacent edge cases: sibling cells sharing omitted end tags in the same row, `<th>`/`<td>` boundary interaction, and cross-row leakage in both directions (a cell opened in one row must not be finalized against a later row's content). All 9 previously-fixed regressions from the prior three rounds on this function still hold. Added 3 new regression tests (2 positive, 1 blank-cell control) — confirmed the 2 positive tests fail against the pre-fix code with the exact `REPORT_FAIL` you reported, and pass against the fix. 36/36 tests passing in both trees (up from 33).

leon1418: [🤖 AI review 🤖] Verified fixed at `e3ca4d232e9592c742e40dcaf30903f0e876ca90` in both mirrors. I reproduced the omitted-cell-end, sibling-cell, and row-transition failures at `b62d8c67aeddb7c9fa2fd36e7102a60ca71cb8ab`, then confirmed the current blocking CLI accepts those valid populated tables. Blank/entity-only cells, commented rows, heading-only/header-only sections, and the empty-table-plus-disclaimer cases still fail. Controls also confirm that blank cells do not inherit text from a following header, another row, or prose outside the table. This closes the remaining substantive-content finding; the packaging and verdict-class fixes remain intact.

leon1418 (commented):
[🤖 AI review 🤖]

Completed the follow-up review of exact head `e3ca4d232e9592c742e40dcaf30903f0e876ca90` against base `ade57adaf9d776d18d316d28d3cee3f746582ced`. The optional-td-end finding is fixed, and all previously reported packaging, content-validation, and attribute/class findings remain addressed. I found no remaining substantive concerns and replied in the original content thread.

Validation: **72/72 tests passed** across both trees; **92/92 scoped CLI probes** matched their expected outcomes; ten before/after comparisons reproduced the prior failures and verified the corrected behavior and negative controls. Isolated skill-only validator execution passed for both copies. Cross-plugin drift (**283 identical, 27 allowlisted**), both Heroku frontmatter checks, both vendored-shared checks, and `git diff --check` passed. The intentional SKILL.md/interpreter/schema differences and the main-window report-validation integration were checked. No full agent-driven migration replay or deployment was run.

Current GitHub CI is green, separately from local validation. **Recommendation: Approve, for the user to submit.** This is a COMMENT follow-up only; I have not submitted approval, dismissed prior reviews, or resolved conversations. The existing blocking review state still requires the user's action.

leon1418 (approved):

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/phases/generate/generate-report.md
```diff
@@ -191,36 +206,47 @@ stakeholders, not a design system.
...
-   pill badges as the sole verdict carrier.
+3. `decision-summary` must **not contain** any `badge-verdict-*` class; when
+   `recommendation.outcome` exists it must render a `verdict-headline` element.
 4. When `recommendation.decision_basis` exists: contains `decision-basis`.
 5. If `scenarios/index.json` has ≥2 scenarios, contains `what-if-scenarios`.
-6. `cost-optimization` is non-empty — either the table or the explicit
-   no-eligible-commitment sentence, never a blank section.
+6. `cost-optimization` is non-empty — the table or the explicit no-eligible-commitment
+   sentence, never a blank section.
+7. Every `<th>` declares `scope="col"`/`"row"`; any `<figure>` has `aria-label` + `<figcaption>`.
```
leon1418, line 223: [🤖 AI review 🤖] **[P2] Teach the report worker the newly blocking currency rule**

The merged validator now runs `_validate_currency_formatting()` on every report, but Step 2 and this in-fragment self-check never require whole-dollar monthly formatting. A report rendering an estimate as `Balanced Est. $112.34/mo` satisfies every listed self-check, then fails the required Finish Generate step solely on currency formatting. I verified that exact report passes the previous reviewed head, fails the current head in both copies, and passes again when rendered as `$112/mo`.

The dispatched worker cannot run the report validator, and the completion gate deliberately does not repair HTML, so this mismatch sends otherwise structurally compliant reports into manual GATE_FAIL recovery. Add the monthly rounding rule and its small-total/per-unit precision exceptions to the authoring instructions and self-check in both mirrors, so the worker produces the format the gate now requires.

leon1418 (commented):
[🤖 AI review 🤖]

Reviewed exact head `2ecdcfcd06c10b0a3ce5e2fd6f6882bc7256e5b9` against base `811fddafced0870173dc9c1a66749a9143baa040`, including the currency-validation merge, relocated tests, standalone skill layout, both plugin copies, and the current discussions. The previously reported packaging, content, optional-td-ending, and attribute/class issues remain fixed.

One integration follow-up is inline: the now-mandatory currency check requires whole-dollar monthly values, but the dispatched Heroku report writer and its self-check do not carry that rule. Please align the producer instructions with the gate. **Recommendation: Request changes for this P2 instruction/validator mismatch.**

Validation: 158 focused tests passed; all 122 Heroku CLI probes and four isolated skill-packaging checks passed. Of 74 GCP currency probes, two expose an implicit-cell-boundary behavior already present in the byte-identical merged-base GCP validator; I am not attributing it to this PR. The relocated Heroku currency helpers match main, and the prior structural/verdict helpers are unchanged. Cross-plugin drift (284 identical, 27 allowlisted), both Heroku frontmatter checks, both vendored-shared checks, and `git diff --check` passed. GitHub CI is green separately from local validation. No full agent-driven migration replay or deployment was performed.

This is a COMMENT review; the existing human approval and historical reviews are left untouched.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/phases/generate/generate-report.md
leon1418: [🤖 AI review 🤖] Rechecked `571f7daaaa419e221e6d4c9fa5404ea9d5c19778` under the combined OCR-assisted/general/repository review. This currency-authoring mismatch still applies: the report instructions and self-check remain unchanged, `$112.34/mo` fails the required CLI check, and `$112/mo`, small totals, and per-unit rates pass. The issue is the missing instruction before the dispatched worker returns, not the intentionally preserved currency-validation rule. Keeping the follow-up in this original thread.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
leon1418: [🤖 AI review 🤖] **[P2] Decode the no-commitment fallback as rendered text.**

The earlier opportunity-row fixes still pass at `571f7daaaa419e221e6d4c9fa5404ea9d5c19778`. The fresh general review found a separate fallback-path case, which I independently reproduced in both mirrors:

```html
<p>No 1&#45;year/3&#45;year commitment product applies to this architecture.</p>
```

This renders the exact required sentence, but the blocking CLI returns `REPORT_FAIL`; `&nbsp;` between the words fails too. The literal paragraph passes. Conversely, placing that literal paragraph only inside an HTML comment or `<template>` returns `REPORT_OK` despite no rendered optimization content.

The raw tag-strip/literal match in `skills/heroku-to-aws/scripts/validate-heroku-migration-report.py:358–361` does not decode entities or exclude inert/comment content. Extract decoded rendered text, preserving inline adjacency and normalizing whitespace, before checking the fallback. Add encoded-sentence and commented/inert-fallback regressions in both copies. This is an additional supported case in the original content-validation thread, not a regression in the fixed row parser.

leon1418 (changes requested):
[🤖 AI review 🤖]

Completed `startups-hybrid-v1` review of exact head `571f7daaaa419e221e6d4c9fa5404ea9d5c19778` against base `ec681ba5d49b039dbf1ef9d87d3304205b824f3d`: actual OCR v1.12.5 delegation, one fresh general reviewer, the parent repository review, and independent candidate verification. All 18 Git inventory entries were covered, including Markdown and deletions. The canonical evidence gate passed.

Two verified P2 concerns remain, updated in the original threads:

1. The dispatched report writer still lacks the monthly rounding instruction required by the blocking currency gate.
2. The no-commitment fallback rejects valid encoded text and can accept content present only in comments/inert templates.

Local validation: the 108 authored tests passed in both independent perspectives; the parent ran 34 CLI/packaging checks, the general reviewer ran 48 focused base/head CLI probes and four packaging checks, and the parent independently ran 14 fallback verification cases. Drift, frontmatter, vendored-source, and diff checks passed. Current GitHub CI is green separately from those results. No full agent-driven migration replay, browser-accessibility run, or deployment is claimed.

**Recommendation: Request changes** for the two P2s. No approval, dismissal, merge, or conversation-resolution action was performed.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
herosjourney: Fixed at `e538f368`, both trees. Replaced the raw `re.sub("<[^>]+>", " ")` tag-strip in `_validate_optimization_content` with rendered-text extraction via the existing `_DecodedTextParser` — entities decoded (`convert_charrefs=True`), `<script>/<style>/<template>` excluded, comment text never seen by the tokenizer — then whitespace-normalized (including NBSP from `&nbsp;`) before the sentence check. So it now uses the same parser the currency gate uses to decide what "rendered" means. Verified against your exact repros: `No 1&#45;year/3&#45;year …` PASSES, `&nbsp;`-separated PASSES, the literal PASSES; the sentence present only inside an HTML comment or `<template>` FAILS. Four regressions added (entity-hyphen, nbsp, comment-only, template-only); the entity/nbsp/template ones verified to fail against the pre-fix parser.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/phases/generate/generate-report.md
herosjourney: Fixed at `e538f368`, both trees. The report worker now carries the blocking whole-dollar rule in two places: (1) a new **"Whole-dollar monthly figures (blocking)"** authoring rule in generate-report.md's cost-labeling section — round every monthly-scale ($2+) figure to a whole dollar when writing exec-costs, the cost one-liner, and the what-if table; cents only for sub-$2 or explicit per-unit rates — and (2) a new Step-3 self-check item 8 telling the worker to scan every $ figure and round any cents before returning, so it self-corrects in-fragment rather than only discovering the failure at the shell-less main-window gate. `.34/mo` is now caught during authoring; `/mo`, small totals, and per-unit rates are unaffected.

leon1418: [🤖 AI review 🤖] Verified addressed at `e538f368cb388fa36e6dbbb0c4e8cd52c58650d2` in both mirrors. The authoring section and Step 3 now require whole-dollar monthly figures at $2+, preserve the small-total/per-unit exceptions, and tell the worker to correct formatting before returning. Current source checks and CLI controls confirm the guidance matches the blocking validator. No remaining concern on this currency-authoring finding.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
leon1418: [🤖 AI review 🤖] **[P2] Exclude inert content from the opportunity-row branch too.**

The decoded fallback fix is verified at `e538f368cb388fa36e6dbbb0c4e8cd52c58650d2`: encoded text passes, and a fallback sentence present only in a comment/template fails. The fresh general review and my independent CLI verification found a separate path that still returns `REPORT_OK` with no rendered optimization content:

```html
<section id="cost-optimization"><template><table><tr><td>RDS Reserved Instance</td></tr></table></template></section>
```

A visible table whose cell contains only a `<template>` or an `application/json` script also passes. `_OpportunityRowParser` does not track inert ancestors, and `_validate_optimization_content()` returns immediately when it reports a populated row, before the corrected decoded-text fallback runs. Please exclude template/script/style subtrees from row/cell acceptance and add inert-table and inert-only-cell negative controls in both mirrors. Keep the existing visible-row, comment, entity, and optional-td cases. This is incomplete enforcement of the new content gate; the base had no such content requirement.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
leon1418: [🤖 AI review 🤖] **[P2] Require the declared verdict headline outside inert templates.**

The earlier class-formatting fixes still pass at `e538f368cb388fa36e6dbbb0c4e8cd52c58650d2`. An additional rendering case is now verified in both mirrors: with a supported Estimate declaring `conditional_go`, decision-summary containing only `<template><p class="verdict-headline">Go, with conditions</p></template>` returns `REPORT_OK`, although it renders no verdict. Removing the headline altogether correctly fails.

`_class_tokens()` collects the class from template descendants, so `_validate_verdict()` treats it as the required rendered headline. Exclude inert template content when establishing headline presence and add this negative case, while retaining normal quoted/spaced/unquoted class support and the badge checks. This concerns the new rendering guarantee, not a regression in the corrected attribute syntax.

leon1418 (changes requested):
[🤖 AI review 🤖]

Completed `startups-hybrid-v1` at exact head `e538f368cb388fa36e6dbbb0c4e8cd52c58650d2` against base `ec681ba5d49b039dbf1ef9d87d3304205b824f3d`: OCR v1.12.5 delegation, one fresh general review, parent repository review, independent candidate verification, and the canonical gate. All 18 Git entries were covered. The actual child's initial-context disclosure and recorded tool audit passed the isolation checks.

The currency-authoring and decoded-fallback fixes are verified. **Two P2 concerns remain:** inert-only opportunity rows can satisfy the content check, and an inert-only verdict headline can satisfy the declared-outcome check. Both were independently reproduced with supported Estimate inputs and updated in the original threads. The base lacked these positive-content requirements; this is incomplete enforcement of the newly introduced guarantees.

Validation: 116 head tests passed in both perspectives; the general reviewer also ran 36 base currency tests. The parent ran 54 CLI/packaging checks and 26 authoritative candidate-verification executions. Drift, frontmatter, vendored-source, and diff checks passed. GitHub CI is green separately from local results. No full agent-driven migration replay, browser-accessibility run, or deployment was performed.

**Recommendation: Request changes** for the inert-content cases. No approval, dismissal, merge, or conversation-resolution action was performed.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
herosjourney: Fixed at ab95346f, and this time as the whole class rather than one parser. _OpportunityRowParser now tracks inert ancestors (script/style/template) with a sticky _inert_depth, mirroring _DecodedTextParser: table/row/cell structure and text inside an inert subtree are skipped, so neither a table nested in a <template> nor a visible cell whose only content is a <template>/<script> registers as a populated opportunity row. Your exact repro (a template-wrapped RDS Reserved Instance table) now returns REPORT_FAIL, and a visible populated row still passes.

Added negative controls in both mirrors: an inert-only opportunity table and an inert-only cell, alongside the retained visible-row, empty-tbody, header-only, entity-only, and commented-out cases. Both new tests fail against the pre-fix script and pass after.

Same sweep hardened the two sibling checks that shared this blind spot (see the verdict-headline thread) and _AccessibilityParser (a <th>/<figure> inside a <template> is no longer audited). Mirrored byte-identical across the advisor and migrate trees; drift/fmt/lint:md/bandit all green.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
herosjourney: Fixed at ab95346f. The root cause was that _class_tokens reads from _collect_tags, and _TagAttrCollector had no inert tracking, so a verdict-headline declared inside a <template> was collected as a rendered class. _TagAttrCollector now skips tags inside inert subtrees (script/style/template), mirroring _DecodedTextParser, so _class_tokens and therefore _validate_verdict only see rendered elements. Your repro (decision-summary containing only a templated verdict-headline, with an Estimate declaring conditional_go) now returns REPORT_FAIL; a visible verdict-headline still satisfies the check, and the badge-verdict-* ban plus normal quoted/spaced/unquoted class support are unchanged.

Added a verdict-headline-only-in-template negative control (with the recommendation.outcome estimate) in both mirrors; it fails pre-fix and passes after.

This was part of one sweep that fixed the same inert blind spot in all three rendered-content checks in the file (opportunity rows, verdict headline, accessibility) so a fourth instance of the class should not surface. Byte-identical across both trees; all local gates green.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
leon1418: [🤖 AI review 🤖] **[P2] Preserve the enclosing context when locating report sections.**

Verified your inner-template/table/cell fixes at `ab95346fcad2bc3f432489d09a7c1696a5b9f7e3` in both mirrors. The original repros now fail, while visible rows, decoded fallback text and optional cell endings still pass.

The complete review found an upstream gap: wrapping the entire section in a template or comment still returns exit 0 / `REPORT_OK`:

```html
<template><section id="cost-optimization"><p>No 1-year/3-year commitment product applies to this architecture.</p></section></template>
```

`_section_counts()` and `_section_html()` scan raw source; extraction removes the enclosing inert/comment context before the corrected content parser runs. The same scanner accepts `data-id="cost-optimization"`, rejects a real `id=cost-optimization` or `id = "cost-optimization"`, and falsely reports a duplicate when a commented copy accompanies the visible section. The skeleton's commented `what-if-scenarios` opening tag also satisfies the two-scenario check without a real comparison section. These cases were reproduced through both blocking CLIs.

Please derive section identities/counts and fragments from parsed document structure, preserving inert ancestry and using the actual `id` attribute. Apply this to both skill-local validators and add whole-section comment/template, false-ID, legal-attribute and conditional-section controls in both test copies, retaining the existing positive and negative cases. This is one section-scanner correction, also covering the verdict thread.

The same ancestor-level gap existed at `e538f368`; I missed it in the previous review. It was not introduced by your latest fix.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
leon1418: [🤖 AI review 🤖] Verified the original inner-template headline repro is fixed at `ab95346fcad2bc3f432489d09a7c1696a5b9f7e3` in both mirrors, including the quoted/spaced/unquoted class controls.

**[P2] The shared section-extraction gap still affects this path:** with a supported `conditional_go` Estimate, `<template><section id="decision-summary"><p class="verdict-headline">Go, with conditions</p></section></template>` returns `REPORT_OK`, although the entire section is inert. `_section_html()` strips away the template ancestor before `_class_tokens()` sees the fragment.

This is covered by the single document-context section-parser correction in the original cost-optimization thread; it does not require another class-parser rewrite. I reproduced it at `e538f368` too and missed it in that review. The inner-template fix itself is working.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/SKILL.md
```diff
@@ -100,7 +107,7 @@ uvx --version 2>/dev/null || echo "UVX_MISSING"
 - If both are present: proceed without nagging. Live pricing still depends on
   the `awspricing` MCP being configured.
 - Soft-warn once if `python3` is missing (Heroku report validation at Generate
-  uses `$PLUGIN_ROOT/scripts/validate-heroku-migration-report.py`). Generate can
+  uses `<SKILL_BASE>/scripts/validate-heroku-migration-report.py`). Generate can
```
leon1418: [🤖 AI review 🤖] **[P2] Align the missing-Python fallback with the mandatory gate.**

When the cold-start check finds no `python3`, this paragraph still says Generate can complete even if validation did not occur. The new `generate.md` Finish Generate step and its `report-validation-status.json` postcondition require the opposite: absent/not_run validation blocks completion. The skill entry and phase therefore give conflicting completion instructions for the same tooling-failure path.

Please update this fallback in both SKILL.md copies: earlier phases may continue after the warning, but Generate remains incomplete until the report validator runs successfully. Keep the existing fail-closed phase gate. This is a verified instruction mismatch; I am not claiming an observed runtime bypass. Both texts already coexisted at `e538f368`, and I missed this consumer in the previous review.

leon1418 (changes requested):
[🤖 AI review 🤖]

Completed `startups-hybrid-v1` with convergence v1 at exact head `ab95346fcad2bc3f432489d09a7c1696a5b9f7e3`, base `ec681ba5d49b039dbf1ef9d87d3304205b824f3d`: OCR v1.12.5 delegation, one fresh general reviewer, separate parent review and impact maps, candidate verification, and the canonical evidence gate. All 18 Git entries were covered.

The two previously reported inner-inert cases are fixed. Packaging, currency guidance, decoded fallback text, optional cell endings and class handling retain their fixes. **Two consolidated P2 corrections remain:**

1. **Section parsing:** in both skill-local validators and structural test copies, use actual parsed section IDs with document/comment/inert context. Invisible sections and `data-id` must not satisfy the gate; real quoted/spaced/unquoted IDs must pass; commented duplicates and the what-if skeleton comment must not count. The original content and verdict threads carry the reproductions.
2. **Missing-Python instructions:** update the tooling fallback in both SKILL.md copies so it no longer promises Generate completion without successful report validation. Earlier phases can continue; the existing missing/not_run gate stays blocking. This is inline.

Both gaps already existed at `e538f368`; I missed them in the prior review. Neither is attributed to the latest inert-parser fix.

Local evidence: 124 authored tests passed in each perspective; the parent ran 54 CLI/packaging checks and 76 authoritative before/after boundary executions with preserved inputs. The fresh reviewer separately reproduced the section issue. Drift, frontmatter, vendored-source and diff checks passed. Current GitHub CI reports nine successful checks, separately from local results. No full agent-driven migration, missing-Python host replay, browser accessibility suite or deployment was performed.

**Recommendation: Request changes.** GitHub also reports merge conflicts, a separate landing blocker. Approval remains for the user; no approval or review dismissal was performed.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
herosjourney: Fixed at aceed8eb. Section identity/count/fragment now come from a stdlib HTMLParser (_SectionParser), never a raw-source regex, so the enclosing context is preserved:

- A section wrapped in an inert subtree (script/style/template) is not counted and its inner HTML is not returned — your template-wrapped cost-optimization repro now returns REPORT_FAIL.
- A section that exists only inside an HTML comment is not counted (comments are a distinct token the parser never re-tokenizes), so a commented duplicate no longer inflates the count and the skeleton comment no longer satisfies a gate.
- The commented what-if-scenarios opening tag no longer satisfies the two-scenario check.
- The real id attribute is read regardless of spelling (quoted, unquoted id=cost-optimization, spaced id = "x"), and data-id is NOT read as the section id.

_section_html was rewritten on the same parser, so it returns the inner HTML of the first RENDERED section and None for a comment/inert-wrapped one — this is what also closes the verdict thread.

This landed as part of reconciling the branch with main (which had independently moved to a parsed _SectionOpenTagCollector and the --mode/decision-basis logic); the merge keeps this skill-relocated validator and folds in main mode logic. Added section-parser regressions (template-wrapped, commented, data-id, unquoted/spaced id, commented-dup, commented-what-if) plus decision-mode/decision-basis tests; 55 co-located tests pass per plugin, byte-identical across trees, full build green (drift/fmt/lint:md/lint:frontmatter/lint:types/bandit).

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
herosjourney: Fixed at aceed8eb by the same single section-parser correction (no separate class-parser change needed, as you noted). _section_html now derives the decision-summary fragment from _SectionParser, which excludes inert subtrees, so a decision-summary whose entire content lives inside a <template> yields None — _class_tokens never sees the templated verdict-headline. Your repro (`<template><section id="decision-summary"><p class="verdict-headline">Go, with conditions</p></section></template>` with a supported conditional_go Estimate) now returns REPORT_FAIL; a visible verdict-headline still passes, and the quoted/spaced/unquoted class controls are unchanged. Regression added: test_verdict_headline_only_in_template_fails_via_section_parser.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/SKILL.md
herosjourney: Fixed at aceed8eb in both SKILL.md copies. The missing-python paragraph no longer says Generate can complete without validation. It now reads: earlier phases (Discover to Estimate) may continue after the warning, but Generate does not complete until the report validator runs successfully — install python3 before Generate — and explicitly ties to generate.md Finish-Generate + the report-validation-status.json postcondition, stating absent/not_run validation is a blocking GATE_FAIL, not a pass. The two texts now agree.

leon1418 (approved):

PR #290, opened 2026-09-13
feat(llm-to-bedrock): accept Decide-complete as sufficient Assess handoff

herosjourney:
## Problem

`llm-to-bedrock` handles a narrow job: rewrite an app's OpenAI/Gemini/Anthropic calls to use Amazon Bedrock. It doesn't do its own analysis of the codebase — it hands that "assess the app" step to the `gcp-to-aws` skill, waits for it to finish, then reads a few files it produced and starts rewriting.

The problem was in *how it decided the assess step was finished*. It waited for `gcp-to-aws` to fully **generate infrastructure** (all the Terraform) — but generating infrastructure is optional in `gcp-to-aws`, and it's completely irrelevant to an SDK rewrite. A normal assessment run stops earlier, at the point where it has produced its recommendation, without generating any infrastructure.

Two things went wrong as a result:

1. **It demanded a step that isn't needed.** Everything `llm-to-bedrock` actually reads is ready as soon as the design work is done — none of it comes from the infrastructure-generation step. So insisting on that step forced users through infra generation they'd never use for a pure code rewrite.
2. **It looped pointlessly.** A run that correctly stopped at the recommendation looked "not finished" to `llm-to-bedrock`, so it kept re-invoking `gcp-to-aws` (up to 6 times) trying to reach a state that run was never going to reach.

An earlier PR (#260) fixed the surrounding wiring; this PR is the remaining piece — letting "the recommendation is done" count as a finished assessment.

<details><summary>In code, for reviewers</summary>

`llm-to-bedrock` delegates Assess to `gcp-to-aws` and waits at its **A2 gate**. A2 required `phases.generate == "completed"`, but `gcp-to-aws`'s Generate is opt-in; a normal AI-path run ends at the decision pack (`run_mode: "decide"`, `current_phase: "complete"`, `phases.generate: "pending"`). The three artifacts Execute consumes — `aws-design-ai.json` (Design), `ai-workload-profile.json` (Discover), `preferences.json` (Clarify) — are all present once Design completes; nothing it reads comes from Generate. So the gate over-required, and an AI-path run that stopped at the decision pack read as "not done," re-invoking `gcp-to-aws` up to its 6× cap. #260 landed the explanation + presence-check + manual-handoff; this PR is the decoupling — accept Decide-complete as a sufficient handoff.
</details>

## Solution

[...]

PR #291, opened 2026-09-13
feat(heroku-to-aws): decision gate — decide is default, Generate is opt-in

herosjourney:
## Problem

When someone runs a migration, the thing they actually want first is the **decision**: should I move to AWS, and what will it cost? Generating all the Terraform and migration scripts is the second step — and only worth doing once they've decided to go ahead.

`gcp-to-aws` already works this way: after it estimates costs, it stops and shows a decision summary, and it only generates the infrastructure files if the user explicitly asks. `heroku-to-aws` did the opposite — as soon as it finished estimating, it **automatically barrelled ahead into generating everything**, with no "do you want to proceed?" step.

That's a problem for two reasons:

1. **It does work the user didn't ask for.** A Heroku user gets pushed through full infrastructure generation just to *see* the recommendation, when the recommendation and the cost picture were the whole point.
2. **It sets a bad precedent for the next skill.** We're about to add an Azure migration skill, and new skills get built by copying an existing one. If Azure copies Heroku's "auto-generate everything" behavior, we end up with a third skill that behaves differently from gcp — and the inconsistency compounds. Fixing Heroku now means Azure copies one consistent "decide first, generate on request" flow.

`gcp-to-aws` stops at a Decision Gate and records `run_mode: "decide"` + `current_phase: "complete"` (with `phases.generate: "pending"`), entering Generate only on explicit opt-in. `heroku-to-aws` instead declared `_advances_to: generate` and its post-Estimate offer was only `[A] workshop / [B] proceed toward Generate`, with B setting `current_phase → generate` unconditionally — no `run_mode`, no decision pack, no opt-in.
</details>

## Solution

Port the gcp decide-default into heroku-to-aws (translated to the DSL idiom):

- **`estimate-assemble.md`** — the post-Estimate offer becomes a 3-option **Decision Gate**: `[A] Done for now` (decide-complete) / `[B] Explore what-ifs` (existing workshop sidebar) / `[C] Generate Terraform and migration scripts`. The workshop returns to the gate rather than falling through to Generate. A decide-complete **resume offer** handles the warm-start case.
- **`estimate.md`** — documents the `_advances_to: generate` exception: choice **A** overrides the default next-phase by setting `current_phase: "complete"` (leaving `phases.generate: "pending"`), the same class of documented exception as the inner-workshop-reprice `HANDOFF_OK` skip. Choice **C** follows `_advances_to` normally.
- **`generate.md`** — a **CONSENT GUARD** at the top of the phase (mirroring gcp): Generate runs only when `run_mode == "decide_and_execute"` or the current-turn message is an explicit Execute request (which sets `run_mode` first); otherwise it STOPs and re-presents the gate. Opt-in is enforced at the execution point, not just in SKILL.md prose.
- **`SKILL.md`** — decide-complete + resume rows and the "Generate is opt-in (HARD RULE)".
- **`report-decision-core.md`** + **`validate-heroku-migration-report.py`** — a shared decision-mode report renderer spec and a new `--mode {full,decision}` arg so the decision pack (`decision-report.html` + `DECISION.md`) validates in decision mode (`decision-cta` instead of `next-steps`). Tests 13 → 15.
- **`run_mode`** added to the **canonical** `shared/state/phase-status.schema.json` (both plugins) + vendored via `shared:sync` — an optional enum `["decide","decide_and_execute"]`. This also advances the shared-schema consolidation (previously `run_mode` lived only in gcp's private prose schema).
- New golden fixture `heroku-decision-gate/after-decide-complete` (the decide-complete tuple: `current_phase=complete`, `run_mode=decide`, `generate=pending`) + asserter.

**Cross-skill payoff:** Heroku now produces the same decide-complete tuple + `DECISION.md` that `llm-to-bedrock`'s Assess handoff (#290) keys on — so the AI-path handoff works identically regardless of source cloud, and Azure copies one decision-flow story.

**Scope:** decision gate + `run_mode` + opt-in entry only. Deliberately **de-bundled from P2-B** (artifact-inference state recovery), which stays a separate PR — it modifies the shared canonical `INTERPRETER.md` and is an independent concern.

## Type of Change
- [ ] New plugin/power/tool
- [ ] Bug fix
- [x] Guardrail/CI update

## Team Folder
- [x] `advisor/`
- [x] `migrate/`
- [ ] Other: ___

## Verification
- `cross-plugin-drift.ts`: **OK** (274 identical, 25 allowlisted — twins byte-identical mod the allowlisted namespace)
- `test_validate_heroku_migration_report.py` (both trees): **15 passed** each (13 → 15; adds the `--mode decision` section rules)
- Fixture asserters (`mise run fixtures:assert`, both trees): **PASS — 9 asserters**, incl. the new golden `heroku-decision-gate/check_expected_decide.py`
- Decide-complete state probed: choice A → `run_mode=decide`, `current_phase=complete`, `generate=pending`, `DECISION.md` present; choice C → `run_mode=decide_and_execute`, `current_phase=generate`; workshop returns to the gate
- `frontmatter-validator` / `markdownlint` / `dprint`: clean (both trees)
- `mise run build` (full composite incl. `bandit`/`semgrep`/`gitleaks`/`checkov`/`grype`): **PASS, exit 0** against a fresh detached worktree at this commit. checkov 276/0; gitleaks no leaks; grype no vulnerabilities.

Not run: no live cross-skill run; state transitions verified against constructed `.phase-status.json` fixtures.

## Merge order (please read)
**This PR restructures `heroku-to-aws` `generate.md` / `generate-assemble.md` / `SKILL.md` and adds a `--mode` arg to `validate-heroku-migration-report.py`.** The open PRs **#287** (tf-policy-gate) and **#288** (report-validation blocking) also touch those same files. This PR is the more foundational change (it establishes the decide-vs-execute flow those two refine), so **please merge this before #287/#288** — those two should then rebase onto this. The overlap is small: #288 touches the same report-validator invocation line (the `--mode full` argument) and both edit the Generate completion region. Verified P2-A does **not** depend on anything in #287/#288.

**Rebase note for whoever rebases #287 onto this:** this PR adds a **CONSENT GUARD** at the top of `generate.md` and a "main-window exception" carve-out in its Scope Boundary. #287 (on its pre-P2-A base) also adds a Scope-Boundary "main-window exception" carve-out (for the tf policy reconcile) and its own main-window "Finish Generate" section. These compose fine logically — the consent guard runs first, then the policy reconcile, then the read-only gate — but on rebase, **keep this PR's CONSENT GUARD (do not revert it)** and **merge the two Scope-Boundary carve-outs into one paragraph** rather than leaving two duplicate "main-window exception" notes.

## Checklist
- [x] I have updated documentation if needed (the change is in the skill instructions themselves)


herosjourney:
More fixes.

### 1. No resume for "gate presented, user hasn't picked A/C" (agreed — real state-machine gap)
Confirmed by tracing the walk-away state (`current_phase=estimate`, `estimate=completed`, `workshop=completed`, `run_mode` absent) against the interpreter:
- Deferred-advance sidebar resume requires the sidebar `"pending"`/`"in_progress"` → workshop is `"completed"`, so it doesn't fire.
- Next rule: `current_phase` present → authoritative → loads `estimate.md` and **re-runs Estimate.** Exactly the opposite of decide-as-default, as you said.

Added the `Gate-presented resume (mandatory)` row to SKILL.md, same shape as the workshop-resume row: if `current_phase == "estimate"` AND `estimate == "completed"` AND `workshop == "completed"` AND `run_mode` absent → do **not** recompute Estimate; re-present the Decision gate (A/C only, workshop already resolved). So closing the laptop at the gate now resumes at the gate.

### 2. Choice-A validator omitted `--migration-dir` (agreed)
Confirmed: the validator's decision-mode pre-execution checks (`terraform/` / `generation-*.json` must not exist yet) are gated on `if migration_dir is not None`, and both invocation sites omitted the flag — so on a re-entry with Terraform already on disk a decide pack could still `REPORT_OK`. Classic fixture-doesn't-match-real-invocation: the asserter passed `--migration-dir`, so the test was green while the real prose invocation was under-armed.

Fixed both sites to pass `--migration-dir "$MIGRATION_DIR"` (`estimate-assemble.md` and `report-decision-core.md`), and fixed the `cwd`-relative `python3 scripts/...` in `report-decision-core.md` to the absolute `$PLUGIN_ROOT` form. The validator + tests already cover the behavior (`test_decision_mode_rejects_terraform_dir_on_disk`, and `test_decision_mode_without_migration_dir_skips_disk_checks` documents the skip) — the bug was purely that the prose never armed it.

### Non-blocking comments — noted, agreed
- **`_advances_to: generate` vs decide-default:** agreed it's a documented exception + consent guard, not a mechanical default. Recorded in the Azure-prep issue (#292): a third skill should copy the **consent guard + resume rows**, not the `_advances_to` line, or it re-inherits "Generate is the mechanical default."
- **llm-to-bedrock still invokes gcp, not Heroku:** correct — the tuple match is what makes it Azure-ready, not a claim that the current AI path is source-agnostic. I won't overstate that.
- **Duplicate "Phase 4 of 6 complete":** left as-is (minor).
- **`report-decision-core.md` doesn't require `cost-optimization` in decision mode:** intentional for the thin decide pack; full Generate still requires it.

### Verification (`193d2d2`, both trees)
- `cross-plugin-drift.ts`: OK (274 identical, 25 allowlisted)
- `mise run build` on a fresh detached worktree: PASS, exit 0 (checkov 276/0, no leaks, no vulns); validator tests 15 passed both trees; `heroku-decision-gate` golden asserter PASS both trees.

Merge-before-#287/#288 note stands.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/SKILL.md
```diff
@@ -199,30 +199,65 @@ contract). Both are `_kind: sidebar` — off-backbone, trigger-entered, never
...
   post-Estimate workshop offer from `estimate-assemble.md`. If `"in_progress"`,
-  load `references/phases/workshop/workshop.md`. Generate must wait until
-  `phases.workshop == "completed"` (entered+exited or declined).
+  load `references/phases/workshop/workshop.md`. The Decision gate — and
+  therefore Generate — must wait until `phases.workshop == "completed"`
+  (entered+exited or declined).
+
+- **Gate-presented resume (mandatory):** If `current_phase == "estimate"` AND
+  `phases.estimate == "completed"` AND `phases.workshop == "completed"` AND
+  `run_mode` is **absent**, the user reached the Decision gate but has not yet
```
leon1418, line 268: [🤖 AI review 🤖] **[P2] Handle populated `run_mode` when reopening the decision gate**

After choice A, a later what-if session retains `run_mode: "decide"`. Following the canonical workshop exit leaves `current_phase: "estimate"`, `phases.estimate: "completed"`, and `phases.workshop: "completed"`. This rule requires `run_mode` to be absent, so it misses that valid state; the interpreter then uses the authoritative `current_phase` and reloads Estimate instead of presenting the gate. Neither workshop re-entry nor its downstream reset clears `run_mode`. Please reset the gate state when reopening it, or allow the retained value here, and add a decide → what-if → gate → session-resume case in both plugins.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/phases/estimate/estimate-assemble.md
```diff
@@ -130,24 +132,109 @@ update — offer:
...
+- **A** → Then:
+  1. **Render the decision pack:** load
+     `references/shared/report-decision-core.md` and render it in **decision**
+     mode — write `$MIGRATION_DIR/decision-report.html` and
+     `$MIGRATION_DIR/DECISION.md` per that file's decision-mode rules (no
+     appendices, no Terraform, CTA footer). Validate with
+     `python3 "$PLUGIN_ROOT/scripts/validate-heroku-migration-report.py" "$MIGRATION_DIR/decision-report.html" --mode decision --migration-dir "$MIGRATION_DIR"`
+     (absolute paths — cwd must not be load-bearing; `--migration-dir` is required
+     so the decision-mode pre-execution checks — no `terraform/` / `generation-*.json`
+     yet — actually run) and fix failures before presenting.
```
leon1418, line 207: [🤖 AI review 🤖] **[P2] Provide a decision-completion path after earlier generation**

The supported post-Generate workshop resets downstream phase statuses but retains `terraform/` and `generation-*.json`. Choosing A after repricing therefore always fails this mandatory validator, even when the decision HTML is valid: both retained artifacts trigger `REPORT_FAIL` in each plugin. Repairing the HTML cannot satisfy the instruction to fix failures before presenting, so the user cannot finish with the newly offered decision-only choice. Please handle this re-entry case before offering A, for example by creating a separate decision run while preserving the previous execution outputs. The pre-execution check itself should remain effective for fresh decision runs.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
```diff
@@ -40,17 +56,30 @@ def _section_counts(html: str) -> dict[str, int]:
     return counts
 
 
-def validate(html: str, migration_dir: Path | None) -> list[str]:
+def validate(html: str, migration_dir: Path | None, mode: str = "full") -> list[str]:
     errors: list[str] = []
     counts = _section_counts(html)
 
-    for sid in REQUIRED_SECTION_IDS:
+    required = [*COMMON_REQUIRED_SECTION_IDS, MODE_REQUIRED_SECTION_ID[mode]]
```
leon1418, line 63: [🤖 AI review 🤖] **[P2] Validate the conditional `decision-basis` section**

With `recommendation.decision_basis` present in `estimation-infra.json`, a decision report that omits `<section id="decision-basis">` still returns `REPORT_OK` in both plugins, including with `--migration-dir` supplied. The new decision renderer requires this section, but the validator never reads the estimate. Full Generate has a separate check in `generate-report.md` Step 3; the new decision path reuses Step 2 content rules and relies on this validator, allowing it to mark the run complete without the required evidence and assumptions. Please require `decision-basis` when the source field exists and add a missing-section regression case.

leon1418 (commented):
[🤖 AI review 🤖]

Reviewed head `193d2d24603db0dbc85102d9b45dae53a0ec9a4b`. Three P2 findings below apply to both plugin copies.

Validation: 15 report-validator tests and all 9 fixture asserters passed in each plugin; cross-plugin drift, frontmatter, vendored-shared checks, and `git diff --check` also passed. The findings were checked with targeted validator probes and traces of the documented DSL state transitions. No live migration was run.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/SKILL.md
herosjourney: Confirmed, and fixed in `b3bf3de7`, both trees.

Traced it further than the immediate symptom: the retained `run_mode` surviving a workshop re-entry is a real bug (fixed — `workshop.md` § Entry now resets `current_phase` to `"estimate"` and clears `run_mode` when re-entering from a resolved decide-complete state), but it's a downstream symptom of a bigger gap. `workshop-assemble.md`'s "When exiting to Generate" section still unconditionally set `current_phase → "generate"` on exit — it was simply never updated when the decision gate was added in this PR. That contradicts `workshop-invariants.md` § 2 (a skill that DEFINES a decision gate must re-present it, never auto-advance to Generate) and every other file already updated in this PR (`SKILL.md`, `estimate-assemble.md`, `workshop.md`'s own prose), all of which already say workshop exit returns to the gate.

Fixed `workshop-assemble.md` to route back to the Decision gate (matching the gcp-to-aws pattern exactly — options A/C only, verdict refreshed from the active scenario). Also fixed `workshop.md`'s "Decline without entering" section, which still referenced the old two-option Estimate offer and advanced straight to Generate — same stale-file issue.

Added the mandatory re-entry reset to `workshop.md` § Entry per your finding, and cross-referenced it from `SKILL.md`'s Warm-start bullet.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/phases/estimate/estimate-assemble.md
herosjourney: Confirmed, and fixed in `b3bf3de7`, both trees.

Root cause: `workshop-refresh.md`'s Stale Generate guard reset `phases.generate` to `pending` (existing `_re_entry_guard` behavior) but never deleted what Generate had written. Design/Estimate are overwritten by the reprice, so `terraform/` and `generation-*.json` from the previous design are stale regardless of what the user chooses next — and the decision-mode validator requires both absent, so choosing A always failed even against a perfectly valid `decision-report.html`, with no way back to decide-complete.

Fixed the guard to also delete the stale execution pack on user confirm: `terraform/` (recursive), `generation-*.json`, `MIGRATION_GUIDE.md`, `README.md`, `migration-report.html`, and `report-validation-status.json`. These never get archived into `scenarios/` (which stores JSON snapshots only, never Terraform) — they're just deleted, since they'll be freshly re-derived from whichever choice (A or C) the user makes at the next Decision-gate pass. If the user declines the re-entry guard, we stop before touching anything (preferences, Design/Estimate re-run, or the execution pack).

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
herosjourney: Confirmed, and fixed in `b3bf3de7`, both trees.

`decision-basis` was only checked structurally (right ID, right place in the skeleton) but the validator never cross-referenced `estimation-infra.json` to see whether Estimate actually declared `recommendation.decision_basis` — so a report that silently dropped the section still returned `REPORT_OK` even with `--migration-dir` supplied.

Added a check mirroring the existing `what-if-scenarios` pattern exactly: when `--migration-dir` is given, read `estimation-infra.json`, and if `recommendation.decision_basis` is present, require `<section id="decision-basis">` in the report. Added three tests: required-when-declared-and-missing (fails), not-required-when-absent (control, pre-extension artifacts), and present-and-required (passes).

herosjourney (commented):
## Fixes for the three P2 findings (pushed as `b3bf3de7`)

All three fixed in both plugin trees, replied inline on each thread:

**1. Handle populated `run_mode` when reopening the decision gate.** Traced past the immediate symptom to the real root cause: `workshop-assemble.md` still unconditionally advanced `current_phase` straight to `"generate"` on exit — it was never updated when the decision gate was added, contradicting `workshop-invariants.md` § 2 and every other file this PR touches. Fixed the assembler to route back to the Decision gate (gcp-to-aws pattern), fixed a stale two-option reference in `workshop.md`'s decline path, and added the mandatory re-entry reset (`current_phase` → `"estimate"`, clear `run_mode`) to `workshop.md` § Entry.

**2. Provide a decision-completion path after earlier generation.** The Stale Generate guard reset `phases.generate` to pending but never removed what Generate had written, so Decision-gate choice A always failed the decision-mode validator (which requires `terraform/`/`generation-*.json` absent) with no way back to decide-complete. Fixed the guard to delete the stale execution pack (`terraform/`, `generation-*.json`, `MIGRATION_GUIDE.md`, `README.md`, `migration-report.html`, `report-validation-status.json`) on user confirm — all of it is stale the moment the reprice overwrites Design/Estimate anyway.

**3. Validate the conditional `decision-basis` section.** The validator checked `decision-basis` structurally but never cross-referenced `estimation-infra.json`'s `recommendation.decision_basis`, so a report that silently dropped the section still passed. Added a check mirroring the existing `what-if-scenarios` pattern; 3 new tests.

### Verification (`b3bf3de7`, both trees)
- `pytest tests/test_validate_heroku_migration_report.py`: **18 passed** (15 → 18)
- `cross-plugin-drift.ts`: **OK** (282 identical, 27 allowlisted)
- `mise run fixtures:assert`: **PASS** (9 asserters, both trees) — including the `heroku-decision-gate` golden fixture, unaffected
- `mise run lint:frontmatter`: OK (both trees)
- `dprint check`: clean
- `markdownlint-cli2`: 0 errors on touched files

Also merged current `main` (this branch had fallen behind — main since merged #287, which touches the same `generate.md`/`phase-status.schema.json` files; resolved the schema conflicts by treating `run_mode` and #287's `run_id`/`owning_skill`/`initiated_by` as sibling additive properties).

migrate/plugins/migration-to-aws/skills/heroku-to-aws/SKILL.md
leon1418: [🤖 AI review 🤖] **[P2] Terminal re-entry is fixed; pending/in-progress Generate still retains the old gate state.**

Verified at `b3bf3de7641d9ffc2b592c7bbdd81ce0d225c5d3` in both trees: decide-complete → workshop → exit now reaches the gate. However, choice C first persists `current_phase: "generate"` and `run_mode: "decide_and_execute"` (while feedback can still be pending), and an interrupted Generate can retain that current phase with `phases.generate: "in_progress"`. The documented warm-start `compare scenarios`/`what if` path also accepts these states. Entry step 3 only resets `current_phase == "complete"`; step 4 keeps the current phase, and the assembler only says it stays at Estimate. Consequently these paths retain `generate` and the old execution consent, so a session resumed after workshop exit selects Generate rather than the unresolved Decision gate.

Please normalize every supported workshop re-entry to the pre-gate state, including Generate pending/in-progress, and cover both in the resume fixture. This result is a trace of the documented DSL transitions, not a live agent migration.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/phases/estimate/estimate-assemble.md
leon1418: [🤖 AI review 🤖] **[P2] The new cleanup is not reached on all paths that offer decision completion.**

At `b3bf3de7641d9ffc2b592c7bbdd81ce0d225c5d3`, `workshop.md` Entry step 2 already applies `reset_downstream_to_pending`. By the time Apply & reprice reaches `workshop-refresh.md`, its `phases.generate == completed` guard is false, so the newly attached cleanup/confirmation is not reliably reached. More directly, entering the supported Compare scenarios path and exiting without Apply never runs that refresh section at all, but still returns to the A/C gate with the old execution pack present. Choice A then fails the real CLI with both `terraform/ must not exist` and `generation-*.json must not exist`; reproduced on isolated copies for both plugins.

Please establish the decision-run/artifact boundary as part of accepted re-entry, covering compare-only, apply, exit, and session resume, before consuming the old Generate status. Preserve the previous outputs; the recursive-deletion regression is called out separately on the new cleanup lines.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
leon1418: [🤖 AI review 🤖] **[P2] The ordinary missing-section case is fixed, but the shipped skeleton placeholder still passes.**

Verified the new missing/present/legacy cases through the real CLI at `b3bf3de7641d9ffc2b592c7bbdd81ce0d225c5d3`. There is still a concrete bypass in both plugins: declare `recommendation.decision_basis`, omit the rendered section, and retain this exact line from `generate-report.md`'s skeleton:

```html
<!-- <section id="decision-basis"> when recommendation.decision_basis exists -->
```

With `--mode decision --migration-dir`, the validator returns `REPORT_OK ... optional=decision-basis`, although an HTML parser finds no such section. `_section_counts` counts the tag inside the comment, so an unexpanded template placeholder satisfies the new requirement while stakeholders receive no evidence/assumptions section. Please count actual HTML section elements, excluding comments, and add this shipped-placeholder regression alongside the three new tests.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/phases/workshop/workshop-refresh.md
```diff
@@ -56,6 +56,30 @@ If `.phase-status.json` has `phases.generate` (or later) `completed`, require
...
+leaves a decision-mode re-entry permanently blocked: `terraform/` and
+`generation-*.json` from the PREVIOUS design still sit on disk, and
+`validate-heroku-migration-report.py --mode decision` requires both absent
+(`report-decision-core.md` — decision mode is pre-execution). Design/Estimate
+are about to be overwritten by this reprice, so that stale execution pack no
+longer matches either artifact regardless of what the user chooses next. On
+user confirm of the guard above, also delete (do not archive into
+`scenarios/`, which stores JSON snapshots only, never Terraform):
+
+- `terraform/` (recursive)
```
leon1418, line 70: [🤖 AI review 🤖] **[P1] Preserve customer state and edits instead of recursively deleting `terraform/`**

This directory is not limited to reproducible generated files. `generate-terraform.md` Step 11 tells users to create `terraform.tfvars` and explicitly ignores `*.tfstate`/backups; `generate-docs.md` tells them to fill customer values, edit the baseline, and run `terraform apply` inside it. A later what-if reprice would therefore delete customer-entered configuration, manual `.tf` edits, and potentially the only local Terraform state along with the generated templates. None of those can be re-derived from Design/Estimate JSON.

The existing `_re_entry_guard` only asks to confirm a re-run, and the workshop sheet says outputs will be "marked stale"; neither authorizes destruction of those files. Please preserve the previous execution directory (for example, create an isolated decision run or retain an explicit archive outside `scenarios/`) and obtain separate informed consent for any actual deletion. A stale design is not evidence that its state or customer edits are disposable. This applies to both mirrors.

leon1418 (changes requested):
[🤖 AI review 🤖]

Re-reviewed `b3bf3de7641d9ffc2b592c7bbdd81ce0d225c5d3` against `b30cb39be02d2eae78cb8c482f4018435d4dc988`, including both plugin trees, the allowlisted state/interpreter differences, and all current discussion.

Requesting changes. The terminal decide-complete path and the ordinary missing decision-basis case are corrected, but the original threads still have substantiated follow-up cases: Generate re-entry retains old gate state, post-Generate decision completion misses cleanup paths, and the shipped commented-out basis placeholder passes validation. The new inline P1 finding covers destructive recursive cleanup of customer state and edits without deletion-specific consent or preservation.

Local validation: 18 report-validator tests and all 9 fixture asserters passed per plugin; cross-plugin drift (282 identical, 27 allowlisted), Heroku frontmatter, vendored-shared checks, and `git diff --check` passed. Additional real-CLI probes reproduced the remaining report failures/bypass; workflow results are traces of the documented DSL state transitions. GitHub independently reports all 9 check runs successful on this head. No deployment or live migration was run, and no customer migration outputs were deleted.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/phases/estimate/estimate-assemble.md
herosjourney: Fixed and pushed at `a5f45b6f`, both trees.

Root cause: the cleanup was gated behind `workshop-refresh.md`'s own `phases.generate == "completed"` check, but `workshop.md` § Entry step 2 already resets `phases.generate` to `"pending"` via `reset_downstream_to_pending` *before* `workshop-refresh.md` is ever reached — so that guard's condition could never observe "completed" on the very re-entry it was meant to catch. And the "Compare scenarios" loop branch never touches `workshop-refresh.md` at all, so it could return to the Decision gate with the old execution pack still on disk regardless of which path was taken.

Moved the guard (detection + confirm + deletion) into `workshop.md` § Entry step 2 itself — the one point every re-entry path (Apply & reprice, Compare scenarios, or exiting workshop without ever entering refresh) passes through before any branch is chosen. `workshop-refresh.md`'s own step 2 is now just a pointer confirming this already happened upstream, rather than a second check that can never fire.

This is bundled with the fix for the sibling P1 finding on `workshop-refresh.md` (recursive `terraform/` deletion) — the same guard now also only deletes the exact named files each Generate fragment declares, never `terraform/` wholesale.

No executable test exists for this DSL/prose contract in this repo (matching your own stated verification method for these files — tracing documented state transitions). Checked internal consistency between `workshop.md` and `workshop-refresh.md`, and confirmed the golden `heroku-workshop`/`heroku-decision-gate` fixture asserters still pass in both trees.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
herosjourney: Fixed and pushed at `a5f45b6f`, both trees.

`_section_counts()` matched `<section id="...">` via a raw regex, so the unexpanded skeleton placeholder comment from `generate-report.md` — `<!-- <section id="decision-basis"> when recommendation.decision_basis exists -->` — was counted as a real, present section even though it renders nothing. `_section_counts` counting a tag inside a comment is the same root cause as the `_has_populated_opportunity_row` bug fixed on PR #288 and the `_section_html` bug fixed on PR #289 this same review cycle: regex against raw markup instead of parsed structure.

Replaced `SECTION_OPEN`/`_section_counts` with an `HTMLParser` subclass (`_SectionOpenTagCollector`) that only counts real start tags — `HTMLParser` routes comment text to `handle_comment`, never re-tokenizing it as a tag, so a commented-out placeholder can never register as present.

Added a regression test using the exact placeholder text from `generate-report.md`. Confirmed it fails against the pre-fix code (`REPORT_OK ... optional=decision-basis`, matching your exact repro) and passes against the fix. 19/19 tests passing in both trees (up from 18).

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/phases/workshop/workshop-refresh.md
herosjourney: Fixed and pushed at `a5f45b6f`, both trees. Thanks for catching this — recursive deletion here was a real risk, not a nitpick.

You're right that `terraform/` is not purely-generated. Replaced the recursive `terraform/` delete with an explicit, named list of only the files each Generate fragment's own `_contributes:` frontmatter declares it writes: `main.tf`, `baseline.tf`, `variables.tf`, `outputs.tf`, `security.tf`, `beanstalk.tf`, `pipeline.tf`, `.gitignore`, `terraform.tfvars.example` (from `generate-terraform.md`), plus `eks.tf`/`kubernetes/` and `.github/workflows/deploy-eb.yml` when present. Every other file under `terraform/` — `terraform.tfvars`, any `*.tfstate*` file, `.terraform/` provider cache, or anything else the customer added or edited — is now explicitly left untouched, and the guard documents why (gitignored tfvars holds real values never regenerated by this skill; state files are often the only local record of what's actually deployed).

Also moved this guard from `workshop-refresh.md` (where it was gated behind a `phases.generate == "completed"` check that `workshop.md` Entry already defuses before that file is reached) up into `workshop.md` § Entry step 2, so it fires on every re-entry path — Apply & reprice, Compare scenarios, and exit-without-refresh — not just the one path that happened to call `workshop-refresh.md`. Details on that half of the fix are on the sibling thread on `estimate-assemble.md`.

No executable test exists for this DSL/prose contract in this repo — verified by tracing the documented state transitions and by rechecking the golden `heroku-workshop`/`heroku-decision-gate` fixture asserters still pass in both trees, matching your own stated verification method for this file.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/SKILL.md
leon1418: [🤖 AI review 🤖] **[P2] Still present at `a5f45b6fa1e9cb8156bc4f963a9567ca25d1be1b` in both mirrors.**

The expanded Entry step 2 now detects an interrupted Generate, but it only resets `phases.generate`. Step 3 still clears `run_mode` and changes `current_phase` only when the current phase is `"complete"`. Starting from `current_phase: "generate"`, `run_mode: "decide_and_execute"`, and Generate either `pending` or `in_progress`, workshop entry/exit therefore retains the old current phase and consent. The documented session-resume rules select Generate instead of the unresolved Decision gate. Adding the in-progress guard did not fix that transition.

Please normalize `current_phase` to `"estimate"` and clear `run_mode` on every accepted workshop re-entry, including these two states. I rechecked both cases against the current schema and documented DSL transitions; the terminal decide-complete path still works.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/phases/estimate/estimate-assemble.md
leon1418: [🤖 AI review 🤖] **[P2] Entry now covers Compare/Exit, but decision completion is still blocked by the preserved directory.**

Verified at `a5f45b6fa1e9cb8156bc4f963a9567ca25d1be1b`: moving the guard to Entry fixes the skipped-refresh path. However, the new contract deliberately leaves `terraform/` in place, including when it contains only customer files. `validate-heroku-migration-report.py` still rejects `(migration_dir / "terraform").exists()` without inspecting its contents. On isolated copies in both plugins, choice A returns `REPORT_FAIL ... terraform/ must not exist` after the named cleanup, with only preserved `terraform.tfvars`, `terraform.tfstate`, and `custom.tf` remaining. Even an empty retained directory fails.

Please align the decision-run layout and validator with preservation of prior outputs, for example by using a separate decision run that leaves the execution directory intact. Deleting the preserved customer files to satisfy this check would undo the safety fix.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
leon1418: [🤖 AI review 🤖] **Verified fixed at `a5f45b6fa1e9cb8156bc4f963a9567ca25d1be1b` in both mirrors.**

The real CLI now rejects the exact commented `decision-basis` placeholder from the shipped skeleton when the estimate declares the basis. It accepts a real section with quoted, unquoted, or multiline attributes and nested content; an ID on a different element or markup inside script text does not satisfy the requirement. The ordinary missing-section and legacy-estimate cases also behave as intended. All 19 report-validator tests pass per plugin. This finding is addressed; the remaining concerns are in the workshop re-entry and artifact-preservation threads.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/phases/workshop/workshop-refresh.md
leon1418: [🤖 AI review 🤖] **[P1] Named generated files can still contain customer edits; the preservation issue remains.**

At `a5f45b6fa1e9cb8156bc4f963a9567ca25d1be1b`, preserving `terraform.tfvars`, state, and unlisted files is an improvement. But the deletion list in `workshop.md` Entry still includes `baseline.tf` and `variables.tf`. The generated guide explicitly tells customers to replace the phone-number placeholders in `baseline.tf` and remove/comment variable blocks when opting out of the baseline. Those edited files keep their generated filenames. `_contributes` identifies original output paths, not whether their current contents are disposable. In a synthetic fixture, applying the new exact-name deletion rule removed both customer-edited files while correctly retaining tfvars/state.

There is no saved-output comparison or backup before deletion, and the existing re-entry confirmation still only covers re-running analysis; the workshop sheet promises files will be "marked stale". A Compare-only visit now reaches deletion before any design change. Please preserve the previous execution pack or archive it intact, and obtain separate informed consent before deleting customer modifications. No real migration outputs were touched in this verification.

leon1418 (changes requested):
[🤖 AI review 🤖]

Re-reviewed `a5f45b6fa1e9cb8156bc4f963a9567ca25d1be1b` against `ade57adaf9d776d18d316d28d3cee3f746582ced`, including all four prior findings, current discussion, full relevant contracts, and both plugin mirrors.

The commented decision-basis finding is verified fixed. Requesting changes for the three remaining findings, detailed in their original threads: P1 deletion of customer edits in named generated files; P2 decision validation rejecting the Terraform directory the cleanup deliberately preserves; and P2 pending/in-progress Generate re-entry retaining the old current phase and execution consent. No duplicate inline findings were added.

Local validation: all 38 report-validator tests and 18 fixture-asserter runs passed, along with Heroku frontmatter, vendored-shared checks, cross-plugin drift (282 identical, 27 allowlisted), and `git diff --check`. Additional real-CLI probes confirmed the parser fix and retained-directory failure. Re-entry traces and the exact-name deletion model used isolated synthetic fixtures; no live migration, deployment, or deletion of customer outputs occurred. GitHub separately reports all 9 check runs successful on this head.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/SKILL.md
herosjourney: Fixed and pushed at `ee6a59a9`, both trees.

Confirmed: the guard in `workshop.md` § Entry step 2 detected `current_phase == "generate"` with `phases.generate` `pending`/`in_progress` as stale, but the normalization step (step 3) that resets `current_phase`/`run_mode` only fired for the `current_phase == "complete"` terminal case — so those two states got a normalized `phases.generate` but a stale `current_phase`/`run_mode`, exactly as you described.

Widened both step 2's detection condition and step 3's normalization to cover all four states uniformly: `phases.generate` completed, `phases.generate` in-progress, `current_phase == "generate"` with `run_mode` set (regardless of what `phases.generate` currently is), and the terminal `current_phase == "complete"` case. Every one of these now gets `current_phase` reset to `"estimate"` and `run_mode` cleared on confirmed re-entry. Updated `SKILL.md`'s Warm-start bullet to reference the same four-condition list rather than re-deriving a narrower summary.

No executable test exists for this DSL/prose contract in this repo; verified by tracing both of your exact repro states (`phases.generate` pending and in-progress, both with `current_phase: "generate"` / `run_mode: "decide_and_execute"`) against the updated text, and confirmed the golden `heroku-workshop`/`heroku-decision-gate` fixture asserters still pass in both trees.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/phases/estimate/estimate-assemble.md
herosjourney: Fixed and pushed at `ee6a59a9`, both trees. You're right — the named-file deletion list still wasn't safe.

Confirmed: `baseline.tf` and `variables.tf` are exactly the two files `generate-terraform.md` tells customers to hand-edit (replace placeholder contact emails; comment out blocks to opt out of the baseline), and they keep their generated filename after being edited. `_contributes:` frontmatter identifies each fragment's ORIGINAL output path, not whether the CURRENT contents at that path are still disposable — there's no way to tell those apart from the filename alone, so no filename-based deletion rule can be made safe.

Took your repeated suggestion and stopped deleting anything: `workshop.md` § Entry step 2 no longer deletes any Generate-authored file. A previous execution pack (including any customer edits to it) is left completely untouched by workshop re-entry now.

This only works because the decision-mode validator's "pre-execution" check no longer keys off raw file presence — see the reply on the sibling `validate-heroku-migration-report.py` thread for that half of the fix (it now reads `.phase-status.json`'s `phases.generate` value instead of checking whether `terraform/`/`generation-*.json` exist on disk). With that change, a stale execution pack sitting on disk after a workshop reprice no longer blocks a fresh decision — so there's no remaining reason to delete anything at all.

Verified by tracing your synthetic-fixture repro (customer-edited `baseline.tf`/`variables.tf` alongside `tfvars`/state) against the updated text — nothing under `terraform/` is deleted now, customer edits or otherwise.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/phases/workshop/workshop-refresh.md
herosjourney: Fixed and pushed at `ee6a59a9`, both trees. Thanks for catching this a second time — you were right that the named-list approach wasn't actually safe.

Confirmed your exact repro: `baseline.tf`/`variables.tf` keep their generated filename after a customer hand-edits them (per `generate-terraform.md`'s own instructions), so `_contributes:`'s original-output-path list can't distinguish "still pristine" from "customer-edited" — no filename-based rule can.

Stopped deleting anything. `workshop.md` § Entry step 2 (the single authoritative guard, referenced from this file) no longer deletes any file Generate wrote — this file's own step 2 now just documents that fact rather than pointing at a deletion list. A previous execution pack, including any customer edits to it, is left completely untouched by workshop re-entry.

This is only possible because the decision-mode validator no longer requires the execution pack absent from disk at all — it now reads `.phase-status.json`'s `phases.generate` value instead (see the reply on the validator's own thread). Once "pre-execution" means "this cycle hasn't gone through Generate" rather than "no generated files exist anywhere," there's no remaining reason to delete or archive anything on re-entry.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
herosjourney: Follow-up note (this thread's own finding is still resolved — no new issue here): the decision-mode pre-execution check changed shape in `ee6a59a9` as part of fixing the sibling P1/re-entry threads.

It previously required `terraform/`/`generation-*.json` absent from disk. That made a valid decision, after a workshop reprice on a previously-executed run, permanently unable to pass — and it's also what motivated deleting the stale execution pack on re-entry, which turned out unsafe (customer-edited `baseline.tf`/`variables.tf` keep their generated filename). It now reads `.phase-status.json`'s `phases.generate` value instead: `"completed"`/`"in_progress"` fails (this cycle already went through Generate), `"pending"` or absent passes regardless of what files happen to still be on disk. Workshop re-entry no longer needs to delete anything as a result.

Added regression coverage for both directions (stale `terraform/` present with `phases.generate: "pending"` → pass; `phases.generate: "completed"` with no files at all → still fail) — flagging here in case it affects how you verify this thread going forward.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/SKILL.md
leon1418: [🤖 AI review 🤖] **Verified fixed at `ee6a59a98a76c06f7cb7afa23976dd715d2f7afe` in both mirrors.**

Entry step 3 now runs for the same accepted event as step 2 and resets both `current_phase` and `run_mode`. I traced pending, in-progress, and completed Generate states, plus terminal decide/execute states: accepted re-entry reaches `current_phase: "estimate"`, clears execution consent, resumes the workshop while active, and returns to the unresolved Decision gate on exit. Declined re-entry leaves the original state unchanged. These are schema-checked traces of the documented DSL, not a live migration run. The previously reported state-reset gap is addressed.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/phases/estimate/estimate-assemble.md
leon1418: [🤖 AI review 🤖] **[P2] The CLI accepts retained artifacts now; the report contract and fixture verifier still reject them.**

At `ee6a59a98a76c06f7cb7afa23976dd715d2f7afe`, the real CLI correctly accepts `phases.generate: "pending"` alongside the preserved execution pack. However, the same valid decide-complete run still fails `fixtures/heroku-decision-gate/check_expected_decide.py` with `terraform/ must not exist` and `generation-*.json must not exist` in both plugins. `report-decision-core.md` lines 26–27 also still declare that those artifacts do not exist in decision mode, and this choice-A instruction still describes file-absence checks. These consumers were not updated with the new current-cycle definition.

Please align the renderer instructions and fixture asserter with the new state-based boundary, and add a retained-execution-pack re-entry fixture. I reproduced the CLI PASS / fixture FAIL on the same isolated copy; the existing golden fixture passes because it contains no prior execution pack. No customer outputs need to be removed.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
leon1418: [🤖 AI review 🤖] **The commented decision-basis finding remains fixed at `ee6a59a98a76c06f7cb7afa23976dd715d2f7afe`.**

Both real CLIs still reject the exact commented skeleton placeholder and accept a real basis section, including when an earlier execution pack is retained. All 21 report-validator tests pass per plugin. I reviewed the newly introduced phase-status check separately; its handling of an existing corrupt state file has a new inline comment, rather than duplicating this already-fixed HTML finding.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/phases/workshop/workshop-refresh.md
leon1418: [🤖 AI review 🤖] **The reported workshop deletion issue is fixed at `ee6a59a98a76c06f7cb7afa23976dd715d2f7afe` in both mirrors.**

Entry and refresh now explicitly preserve the entire prior execution pack and contain no deletion list. I traced accepted Compare/Apply/Exit and declined re-entry, and checked synthetic copies containing customer-edited `baseline.tf`/`variables.tf`, tfvars, Terraform state, a custom `.tf`, a Kubernetes manifest, and prior reports/docs. All eleven execution-artifact hashes remained unchanged during the state/CLI probes. The decision CLI accepts those retained files with Generate pending. No live migration was executed and no real customer files were accessed or changed. The remaining report-consumer mismatch is tracked in the original decision-completion thread.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
```diff
@@ -68,23 +119,89 @@ def validate(html: str, migration_dir: Path | None) -> list[str]:
...
+        # description). Fail open on a missing/corrupt status file (an
+        # ambiguous case must not block a genuinely fresh decision run).
+        phase_path = migration_dir / ".phase-status.json"
+        generate_status: str | None = None
+        if phase_path.is_file():
+            try:
+                phase = json.loads(phase_path.read_text(encoding="utf-8"))
+                generate_status = (phase or {}).get("phases", {}).get("generate")
+            except (OSError, json.JSONDecodeError):
+                generate_status = None
```
leon1418: [🤖 AI review 🤖] **[P2] Do not treat an existing corrupt state file as verified pre-execution**

With `--mode decision --migration-dir`, an existing `.phase-status.json` containing truncated JSON such as `{"phases":{"generate":"completed"` now returns `REPORT_OK`; the same state with valid JSON correctly fails. An empty existing file also passes in both plugins. This exception handler turns an inability to read the sole current-cycle signal into success, so the validator can no longer establish the pre-execution invariant it claims to check. The interpreter's State-file validation contract requires invalid state to stop rather than be guessed.

Please return a `REPORT_FAIL` diagnostic for an existing unreadable/corrupt state file and validate the expected state shape/status. Keep any intentionally supported HTML-only or missing-state compatibility path distinct from corruption; a broken existing run is not evidence of a fresh cycle. Add a regression for the existing-corrupt-file case.

advisor/plugins/aws-startup-advisor/skills/heroku-to-aws/SKILL.md
```diff
@@ -228,56 +228,83 @@ contract). Both are `_kind: sidebar` — off-backbone, trigger-entered, never
...
+  `current_phase` to `"estimate"` and clear `run_mode` first, for every one of
+  those states, not only the terminal one. This is what lets the gate re-fire
+  on exit instead of the run falling through to a re-run of Estimate or a
+  re-selection of Generate. Knobs on the pilot sheet: region, HA, compute
+  target, cost optimization, CPU architecture (x86 vs Graviton). There is no
+  traffic-multiplier knob in v1.
 
 - **After Generate**: No prompt. If `phases.feedback` is still `"pending"`, set it to `"completed"` and mark the migration complete.
 
 **Critical constraint**: Follow each phase reference file's workflow exactly. If unable to complete a step, stop and report the specific issue. Do not fabricate or infer data.
```
leon1418, line 310: [🤖 AI review 🤖] **[P2] Preserve the advisor-only contextual-offers closing step**

This update also removes the entire `Contextual offers (final step)` block that exists in the reviewed base and the previous PR head. That block is the advisor Heroku flow's instruction to check the offers catalog and surface a qualifying offer after a final recommendation. There is no replacement here, so this PR silently disables that existing behavior while changing the decision gate.

The cross-plugin checker intentionally allowlists this SKILL.md because advisor has the offers catalog and migrate does not. Making the two files identical is therefore not sufficient validation; the green count moving to 283 identical / 26 allowlisted reflects the lost intentional difference. Please restore the advisor-only block while keeping the shared decision/re-entry changes.

leon1418 (changes requested):
[🤖 AI review 🤖]

Re-reviewed `ee6a59a98a76c06f7cb7afa23976dd715d2f7afe` against `ade57adaf9d776d18d316d28d3cee3f746582ced`, including all prior threads and both plugin trees.

Verified fixes: workshop no longer deletes customer execution artifacts; pending/in-progress Generate re-entry clears current phase and execution consent correctly; the commented decision-basis regression remains fixed. The real decision CLI now accepts retained execution files with Generate pending and rejects completed/in-progress Generate.

Requesting changes for three P2 concerns: the report instructions/fixture verifier still enforce the old file-absence contract (follow-up in the original thread); existing corrupt state silently bypasses the new pre-execution check; and the advisor-only contextual-offers closing step was removed. The latter two have new inline comments.

Local validation: 42 report-validator tests and 18 fixture checks passed, plus frontmatter, vendored-shared checks, cross-plugin drift (283 identical, 26 allowlisted), and `git diff --check`. Targeted real-CLI probes reproduced the remaining state-validation and fixture-boundary failures. Re-entry results are explicit traces of the documented DSL, with synthetic execution-file hashes preserved; no live migration or deployment was run. GitHub separately reports all nine check runs successful on this head.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/phases/estimate/estimate-assemble.md
herosjourney: Fixed and pushed at `e8cae60b`, both trees.

Confirmed: the fixture asserter still asserted `terraform/`/`generation-*.json` absence by raw file presence — the pre-state-based-fix contract — and `report-decision-core.md` still said those files "do NOT exist in decision mode" unconditionally. Both stale.

Removed the stale file-absence assertions from `check_expected_decide.py` and updated the doc to describe the actual contract (pre-execution is `phases.generate`-based, not filesystem-based). Also added a sibling golden fixture, `heroku-decision-gate/retained-execution-pack` (dispatched via `check_expected_decide_with_retained_pack.py`, which reuses the original asserter's exact logic), containing a valid decide-complete state alongside a retained execution pack with customer-edited `baseline.tf`, hand-authored `terraform.tfvars`, and a stale `generation-warnings.json` — so the CI-dispatched golden path now actually exercises the retained-pack scenario, not just the CLI checked by hand. 10 asserters now (up from 9), both green.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
herosjourney: Fixed and pushed at `e8cae60b`, both trees.

You're right — my fail-open path treated a missing file and a corrupt-but-existing file identically, and `INTERPRETER.md`'s own State-file validation contract is explicit that invalid JSON is a STOP condition, not something to guess past.

Distinguished the two cases: a genuinely missing `.phase-status.json` (no run has ever tracked state here) still fails open, but a file that exists and fails to parse — truncated, empty, or otherwise malformed — now returns `REPORT_FAIL` with a diagnostic pointing at `INTERPRETER.md`'s State-file validation section, rather than silently defaulting to "pending."

Verified against both your exact repros (truncated JSON, empty file) — both now correctly fail; confirmed they returned `REPORT_OK` on the pre-fix code, matching your report exactly. Added 2 regression tests; the existing missing-file-fails-open control is unchanged. 23/23 tests passing in both trees (up from 21).

advisor/plugins/aws-startup-advisor/skills/heroku-to-aws/SKILL.md
herosjourney: Fixed and pushed at `e8cae60b`. Thanks for catching this — a genuine regression, not something the drift checker could have caught given the intentional allowlist.

Confirmed: the "Contextual offers (final step)" block was dropped when I mirrored the `migrate/` tree's `SKILL.md` content over the `advisor/` copy during an earlier decision-gate edit — `migrate/` never had this block (no offers catalog there), so the mirror silently wiped it from `advisor/`.

Restored the block verbatim from `origin/main`, keeping my legitimate decision-gate changes earlier in the file intact. Verified the file's tail is now byte-identical to `origin/main`'s version beyond those changes, and confirmed `cross-plugin-drift`'s allowlisted-difference count is back to 27 (had silently dropped to 26 when the block was lost — thanks for flagging that specific signal, that's a good tell for this class of regression going forward).

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/phases/estimate/estimate-assemble.md
leon1418: [🤖 AI review 🤖] **Verified fixed at `fe5bb105e28f9e5660da1b804ca2e0b2ef757eb7` in both mirrors, including the change after `e8cae60b`.**

The state-based contract is now reflected in the renderer and shared fixture asserter. Both the ordinary and retained-execution-pack golden paths pass, and the same retained temporary run passes the real CLI and shared asserter. Customer-edited Terraform, tfvars/state, and prior reports remain unchanged. Both plugins pass all 10 fixture asserters and the subsequent fixture-integrity check. I also ran the retained-pack wrapper in an isolated plugin copy with `python -E`: it passes without creating `__pycache__` or changing any file.

Non-blocking wording cleanup: the parenthetical at `estimate-assemble.md` lines 205–207 still describes the old file-absence check. The canonical renderer and executable checks now use the correct state-based rule, so this does not block approval.

migrate/plugins/migration-to-aws/scripts/validate-heroku-migration-report.py
leon1418: [🤖 AI review 🤖] **Verified fixed at `fe5bb105e28f9e5660da1b804ca2e0b2ef757eb7` in both mirrors.**

Both original CLI repros—an existing truncated `.phase-status.json` and an existing empty file—now return `REPORT_FAIL` with the corruption diagnostic rather than `REPORT_OK`. Controls also behave as intended: Generate pending with retained execution artifacts passes; completed/in-progress Generate fails; the documented missing-file compatibility case remains supported. All 23 report-validator tests pass per plugin, and the earlier commented decision-basis regression remains fixed. No remaining blocker from this finding.

advisor/plugins/aws-startup-advisor/skills/heroku-to-aws/SKILL.md
leon1418: [🤖 AI review 🤖] **Verified fixed at `fe5bb105e28f9e5660da1b804ca2e0b2ef757eb7`.**

The advisor-only `Contextual offers (final step)` block is restored verbatim from the reviewed base, while migrate correctly has no corresponding block. The shared decision/re-entry edits remain in place. I checked the actual file tails and the other allowlisted differences, not only the green drift result; the checker is back to 282 identical / 27 allowlisted. This regression is addressed.

leon1418 (commented):
[🤖 AI review 🤖]

Completed the current-head re-review of `fe5bb105e28f9e5660da1b804ca2e0b2ef757eb7` against `ade57adaf9d776d18d316d28d3cee3f746582ced`, including the post-`e8cae60b` bytecode-cache fix. All previously reported substantive findings are verified fixed; I found no new substantive issue. Approval is recommended for the human reviewer to submit.

Verified in both plugin copies: retained execution packs pass the real decision CLI and golden asserters; the original truncated/empty state-file cases fail with diagnostics; the advisor-only closing step is restored; workshop re-entry preserves customer edits and state while correctly resetting Generate consent/current phase; commented basis markup is still rejected. The isolated retained-pack wrapper also passes with `python -E` without writing bytecode or changing fixture files.

Local validation: 46 report-validator tests, 20 fixture-asserter runs, both fixture-integrity checks, Heroku frontmatter, vendored-shared checks, cross-plugin drift (282 identical, 27 allowlisted), and `git diff --check` passed. Workflow transition evidence is from explicit DSL traces and isolated synthetic fixtures, not a live migration or deployment. GitHub separately reports all nine check runs successful on this head.

This COMMENT records verification; it does not submit approval or dismiss earlier blocking reviews. Required approval and unresolved review conversations remain outstanding merge requirements.

migrate/plugins/migration-to-aws/skills/heroku-to-aws/references/phases/estimate/estimate-assemble.md
herosjourney: Cleaned up and pushed at `7986b9c3`, both trees — thanks for flagging it even as non-blocking.

The parenthetical at `estimate-assemble.md` still described the old file-absence framing ("no `terraform/` / `generation-*.json` yet"). Updated it to describe the actual check (`.phase-status.json`'s `phases.generate` must be `pending`/absent for the current cycle) and pointed at `report-decision-core.md` for the full contract. Prose-only change, no behavior difference. Verified `dprint`, `markdownlint`, `lint:frontmatter`, `cross-plugin-drift` (282/27, unchanged), and `fixtures:assert` (10 asserters) all clean in both trees.

leon1418: [🤖 AI review 🤖] Verified at `7986b9c38f92f55a82c0d92facbf03786d80b44e`: the complete delta from `fe5bb105` is this same explanatory paragraph in the two plugin copies. It now matches the state-based current-cycle check and explicitly permits the retained execution pack. The validator invocation and all executable code, preservation/consent logic, fixtures, and advisor-only closing behavior are unchanged.

Fresh checks: both retained-pack CLI/asserter paths pass without changing fixture files; cross-plugin drift (282 identical / 27 allowlisted) and `git diff --check` pass. I reused the previous exact-code validation of 46 tests and 20 fixture checks rather than rerunning those unchanged suites. This addresses the final non-blocking wording note. No remaining review findings; approval remains recommended for the human reviewer to submit. No approval or dismissal was performed.

leon1418 (approved):

leon1418 (commented):
[🤖 AI review 🤖]

Completed the current-head review of `ec8ca7cd22e87a9b1a371f1d28de363b944d3593` against `ec681ba5d49b039dbf1ef9d87d3304205b824f3d`, including the complete delta from the previously reviewed `7986b9c3` and both queued merges. No substantive findings remain; the approval recommendation is unchanged.

The merged currency checks work in both Heroku report modes without bypassing the decision lifecycle checks. Retained execution packs pass with Generate pending; completed/in-progress Generate and the original corrupt-state cases fail as intended. Commented decision-basis markup remains rejected. The preservation, consent/reset instructions, schemas, and fixtures are unchanged from the prior verified source. Currency helpers match the merged base, and the base-only plan writer introduces no changed lifecycle consumer in this PR. Intentional advisor/migrate differences remain intact.

Local validation passed: 266 report tests across both plugin copies, 150 real-CLI currency/lifecycle probes, 20 fixture-asserter runs, Heroku frontmatter, vendored-shared checks, cross-plugin drift (282 identical, 27 allowlisted), and `git diff --check`. Probes covered rendered/inert markup, monthly suffixes, block boundaries, rate operands in both orders, retained artifacts, malformed state, and combined currency/lifecycle failures. No live migration or deployment was run; no customer outputs were touched. GitHub separately reports all nine checks successful on this head.

An existing human APPROVED review is recorded on this exact head and is preserved. This is a COMMENT only; no approval, dismissal, or conversation resolution was submitted. Merge readiness remains separate because review conversations are still unresolved.

Issue #295, opened 2026-09-14
RFC: Add contextual-offers-for-startups skill


medhatisac-eng:
## Summary

Add a `contextual-offers-for-startups` skill to `aws-startup-advisor` that runs **after** another recommendation skill has finalized its advice and, when genuinely relevant, appends a **single** AWS Activate partner offer as one quiet, dismissible line. Strictly merit-first, relevance-gated, one-or-none, capped, and muteable. Offer content is read live from `knowledge-base-for-startups/references/offers.md`.

## Motivation

AWS Activate partner offers are lookup-only today: a founder sees an offer only if they explicitly ask (a `knowledge-base-for-startups` keyword lookup) or navigate to a static entry point. Nothing surfaces a relevant offer in the moment, tied to what the founder is actually building or migrating — so a benefit that sits alongside Activate credits stays hard to discover. This adds the contextual layer without turning recommendations into advertising.

## Proposed Solution

- New folder: `advisor/plugins/aws-startup-advisor/skills/contextual-offers-for-startups/` — a flat `SKILL.md` owning the contextual behavior.
- Consulted as a final step by the recommendation-producing skills: `architect-for-startups`, `start-building-for-startups`, `agent-advisor`, `gcp-to-aws`, `heroku-to-aws`, `llm-to-bedrock`; a pointer added to `knowledge-base-for-startups`. `prompt-library-for-startups`, `tf-best-practices`, and `shared/` are intentionally unchanged.
- **Guardrails:**
  - **Merit-first** — offer availability is an annotation, never an input; lookup happens only after the recommendation is final and never re-ranks it.
  - **Transparent alternative** — when the merit pick is an AWS-native service and a competing partner offer exists, the AWS-native service is stated as the recommendation and the partner is presented as an optional alternative.
  - **Non-intrusive, answer-first** — one clearly-labeled optional line after the answer; never advertising, never proactive.
  - **Surfacing controls** — ≤1 offer per response, ≤1 per five messages, ≤2 per session, session dedupe, mute honored immediately.
- Offer value and eligibility are taken as written from the offer's detail file; nothing fabricated.
- **Click tracking:** the redeem link carries tracking parameters (a source identifier + offer slug + surface tag) appended to the offer's canonical `source_url` only — never changing the destination.

## Alternatives Considered

- **Keep offers lookup-only (status quo)** — rejected: misses the moment of intent.
[...]

PR #303, opened 2026-09-16
feat(contextual-offers-for-startups): add merit-first Activate offer layer

medhatisac-eng:
## Problem

The AWS Startup Advisor skills give strong technical recommendations, but relevant AWS Activate partner offers (for products a founder is actually building with or migrating to) are only discoverable if the founder goes looking in the knowledge base. There was no way to surface a genuinely relevant offer in context — and no disciplined, trust-preserving mechanism to do so without turning the assistant into an ad channel or letting commercial incentives bias the technical advice.

RFC: https://github.com/awslabs/startups/issues/295

## Solution

Add a new offer layer skill; each recommendation-producing skill runs the offer check inline as a mandatory closing step of its own output, after its advice is already final. The layer is merit-first, relevance-gated, quiet, and muteable — it appends at most one offer line, and often none.

Per-file summary:

- `advisor/plugins/aws-startup-advisor/skills/contextual-offers-for-startups/SKILL.md` (new) — the offer layer. Runs only after a calling skill's recommendation is final; never touches or re-ranks that recommendation. Surfaces a single AWS Activate offer only when a strict relevance rule holds (a named-product match, or a direct intent/component match to something the founder actually surfaced), with a high bar and show-one-or-none. Reads offer content live from `knowledge-base-for-startups/references/offers.md` (never snapshots, never fabricates values/slugs/eligibility). Renders one quiet, dismissible italic line whose redeem link is the offer's bare canonical `source_url`; documents frequency caps, session dedupe, mute, and honest eligibility.
- `advisor/plugins/aws-startup-advisor/skills/agent-advisor/SKILL.md`, `architect-for-startups/SKILL.md`, `gcp-to-aws/SKILL.md`, `heroku-to-aws/SKILL.md`, `llm-to-bedrock/SKILL.md`, `start-building-for-startups/SKILL.md` — each gains a "Contextual offers (final step)" section that runs the offer check inline as a mandatory closing step of its own output: it performs the relevance check itself and appends the single offer line if warranted, otherwise appends nothing, without depending on a separate skill being loaded. No change to how any of these skills decide their advice.
- `advisor/plugins/aws-startup-advisor/skills/knowledge-base-for-startups/SKILL.md` — adds a "Contextual offers companion" pointer noting the offers here can also be surfaced contextually by the new layer. This skill stays in lookup mode and remains the offer source of record; both read the same `references/offers.md`.
- `advisor/plugins/aws-startup-advisor/skills/knowledge-base-for-startups/references/offers/auth0.md` — adds an "Estimated offer value" heading so the contextual offer line can state the offer's value accurately.

Follow-up (out of scope here): offer-redeem-link tracking params + an "open in browser" prompt, in a separate change (tracking scheme pending analytics-team confirmation).

## Type of Change

- [x] New plugin/power/tool
- [ ] Bug fix
- [ ] Documentation update
- [ ] Guardrail/CI update

## Team Folder

- [x] `advisor/`
- [ ] `migrate/`
- [ ] Other: ___

## Checklist

- [ ] I have run `mise run build` locally and it passes


itsSauravK (commented):
considering all the skills have the same offers section at the end, do you think it is possible to somehow all of them to share a common source? If not, then this looks good

herosjourney:
**Scope concern: `transparent-alternative` clause is a poor fit for `gcp-to-aws` / `heroku-to-aws`**

Files: `advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/SKILL.md:442`, `advisor/plugins/aws-startup-advisor/skills/heroku-to-aws/SKILL.md:274`

> or the **transparent-alternative case** (your merit pick is an AWS-native service and a relevant partner offer exists for that same need)

Unlike the other two triggers (named-product match, intent/component match), this one doesn't require the founder to have raised the need at all — it only requires that the AWS-native service *appear somewhere in the recommendation*. For `architect-for-startups` / `start-building-for-startups`, that's a reasonable fit: those are open-ended advisory conversations where "here's an alternative you didn't ask about" is in-genre.

For the migration skills it's a worse fit, for two reasons specific to how they work:

1. **The "recommendation" is a whole multi-component infra plan** (compute, database, cache, queue, auth/IAM, etc.), not a single pick. A service like Cognito can appear as one incidental piece of a plan whose actual subject is something else entirely (e.g. Postgres → RDS, Kafka → MSK). Under this clause, that's still enough to justify surfacing an unrelated offer (e.g. Auth0) — with no requirement that the founder is migrating auth, mentioned auth, or cares about auth in this run.
2. **It can directly contradict the plan just delivered.** If the migration is genuinely moving Firebase Auth → Cognito, appending "if you prefer a managed alternative, Auth0 has an offer" undercuts the assistant's own recommendation in the same breath — worse than a generic upsell, because the plan and the footnote now disagree on what the founder should do.

Both cases produce founder confusion ("I'm not migrating my auth, why is this here?") and, in the contradiction case, active undermining of trust in the migration plan itself — which is a worse outcome than the "no offer shown" default this skill is designed to prefer.

**Suggested fix:** for `gcp-to-aws` and `heroku-to-aws` specifically, either drop the transparent-alternative clause entirely, or narrow it so it only fires when the AWS-native service is a component the founder is *actively migrating in this run* (i.e. it's the source-side workload identified in Discover/Clarify, not merely present anywhere in the output plan). The advisory skills (`architect-for-startups`, `start-building-for-startups`, `agent-advisor`) can reasonably keep the broader clause since there's no equivalent "plan vs. footnote" contradiction risk there.

herosjourney:
**Scope concern: `transparent-alternative` clause is a poor fit for the workflow skills (`gcp-to-aws`, `heroku-to-aws`, `llm-to-bedrock`, `agent-advisor`)**

Files: `advisor/plugins/aws-startup-advisor/skills/gcp-to-aws/SKILL.md:442`, `advisor/plugins/aws-startup-advisor/skills/heroku-to-aws/SKILL.md:274`, `advisor/plugins/aws-startup-advisor/skills/llm-to-bedrock/SKILL.md`, `advisor/plugins/aws-startup-advisor/skills/agent-advisor/SKILL.md`

> or the **transparent-alternative case** (your merit pick is an AWS-native service and a relevant partner offer exists for that same need)

Unlike the other two triggers (named-product match, intent/component match), this one doesn't require the founder to have raised the need at all — it only requires that the AWS-native service *appear somewhere in the recommendation*. For `architect-for-startups` / `start-building-for-startups`, that's a reasonable fit: those are open-ended advisory conversations where "here's an alternative you didn't ask about" is in-genre.

For the four workflow skills it's a worse fit, because each one delivers a single integrated, multi-part plan rather than a conversational pick — and the clause doesn't check whether the founder actually asked about the specific part it's matching against:

- **`gcp-to-aws` / `heroku-to-aws`**: the recommendation is a whole infra plan (compute, database, cache, queue, auth/IAM, etc.). A service like Cognito can appear as one incidental piece of a plan whose actual subject is something else (e.g. Postgres → RDS, Kafka → MSK) — enough to justify surfacing an unrelated offer (e.g. Auth0) with no signal the founder is migrating auth at all. Worse, if the plan genuinely *is* moving auth to Cognito, the appended "if you prefer a managed alternative, Auth0 has an offer" directly undercuts the recommendation it's attached to.
- **`llm-to-bedrock`**: this skill's entire purpose is migrating the founder off OpenAI/Gemini/Anthropic onto Bedrock — Bedrock isn't one component of the plan, it *is* the deliverable. The offers catalog has direct alternatives for that same need (Fireworks AI for open-model inference, CompactifAI for compressed LLM inference). Finishing a Bedrock rewrite and appending a footnote suggesting a different LLM provider undercuts the one thing the skill was asked to do.
- **`agent-advisor`**: the recommendation bundles a runtime pick plus AgentCore services (Memory, Gateway, Identity). Descope's "Agentic Identity Hub" is a direct alternative to AgentCore Identity, so recommending AgentCore Identity as part of an architecture could trigger a Descope footnote the founder never asked for. (Its Temporal offer is a different, fine case — Temporal workflow code is explicitly kept as-is per this skill's design, so a Temporal offer there is a legitimate named-product match, not this problem.)

In all four cases the failure mode is the same: founder confusion ("I didn't ask about auth/my LLM provider/identity, why is this here?"), or worse, the offer footnote contradicting the plan the assistant just delivered — a worse outcome than the "no offer shown" default this skill is designed to prefer.

**Suggested fix:** for these four skills, either drop the transparent-alternative clause, or narrow it so it only fires when the AWS-native service is a component the founder is *actively migrating/deciding on in this run* (i.e. it came from Discover/Clarify as the thing being worked on), not merely present anywhere in the output. `architect-for-startups`, `start-building-for-startups`, and general advisory use can reasonably keep the broader clause since there's no equivalent "plan vs. footnote" contradiction risk there.

medhatisac-eng:
@herosjourney good call — agreed. For `gcp-to-aws` and `heroku-to-aws` I've dropped the transparent-alternative trigger (and its line template) in 480d12e, so those skills now surface an offer only on a named-product match or a direct intent/component match to something the founder actually raised. The advisory skills (`architect-for-startups`, `start-building-for-startups`, `agent-advisor`) keep the broader clause as you suggested. Note: the "narrow it to the actively-migrated component" option wouldn't fully cover the contradiction case you gave (Firebase Auth → Cognito is itself an auth migration), so dropping it was the cleaner fix. One follow-up: `llm-to-bedrock` has a similar shape (recommend Bedrock, then a competing-LLM offer could contradict the migration) — want me to drop it there too, or is that different enough to keep?

@SauravK the repeated block is intentionally inlined in each skill rather than shared, for reliability. An earlier version delegated to a separate "consult the offers skill" step, and in testing the model frequently didn't load/run it, so the offer silently never appeared; making it a self-contained inline step is what made it fire dependably. The repo's vendored-shared tooling copies whole files/trees, not a fragment into the middle of each SKILL.md, so single-sourcing this short block would need new build machinery. Happy to revisit a synced-fragment approach later if we want to DRY it up.

herosjourney:
Thanks for the quick turnaround on 480d12e — that's the cleaner fix, agreed on the Cognito point.

On `llm-to-bedrock`: yes, drop it there too. It's actually the cleanest case of the four — this skill's only job is the Bedrock migration, so there's no benign version of the transparent-alternative trigger for it. A named-product match or intent/component match would require the founder to have actually raised Fireworks/CompactifAI or an open-model need; the transparent-alternative case, by contrast, only requires that Bedrock (the deliverable itself) be the AWS-native pick, which is always true here. Every firing of that clause on this skill would be suggesting the founder reconsider the migration they just asked for and got — there's no case analogous to `architect-for-startups` where "here's an AWS-native alternative to a partner pick" is genuinely useful context. Drop it.

medhatisac-eng:
@herosjourney thanks — agreed on the expanded scope. In f9555d7 I've dropped the transparent-alternative trigger (and its line template) from `llm-to-bedrock` and `agent-advisor` as well, so all four workflow skills now surface offers only on a named-product match or a direct intent/component match to something the founder actually raised. Only `architect-for-startups` and `start-building-for-startups` keep the broader transparent-alternative clause. And good point on `agent-advisor` — the Temporal offer still works there as a named-product match (Temporal is kept as-is per the skill's design), which this change leaves intact.

itsSauravK (approved):

herosjourney:
Approve.

PR #304, opened 2026-09-17
feat(azure-to-aws): add Azure→AWS migration skill (infra + AI/agentic tracks)

icarthick:
## What

Adds the **`azure-to-aws`** migration skill to both plugin trees (`aws-startup-advisor` and `migration-to-aws`), built as "gcp-to-aws's product surface on heroku-to-aws's DSL machinery." It takes an Azure estate (Terraform IaC and/or application code) through Discover → Clarify → Design → Estimate → (opt-in) Generate, producing an AWS design, a cost estimate, and — when the user opts into `decide_and_execute` — Terraform + a migration report.

Includes a self-contained **AI/agentic workload** track (Azure OpenAI / Cognitive Services → Amazon Bedrock, with traditional-AI capabilities routed to Textract/Rekognition/etc.), mirroring gcp's self-contained AI subsystem.

## Why

Azure is the remaining major source cloud without a first-class migration skill. This lands it on the shared, already-converged AWS-side product shape (decide-vs-execute gate, `run_mode`, Assess handoff via `DECISION.md`, tf-policy gate), so it plugs into the existing gate/report contracts rather than inventing new ones. Design posture: **mirror gcp; diverge only for a documented Azure fact** (ARM-ID edges, x86-default for Windows/.NET, App Service Plan fan-in), each recorded as a decision.

## Validation

- **First real end-to-end capability run:** exercised the skill headless (`claude -p`, real plugin) across **14 Azure sample repos** (8 infra, 2 AI iac+appcode, 4 AI app-code-only). **13/14 completed all phases** with `run_mode: decide_and_execute` and a full artifact set; the 1 miss was shared-git-tree corruption from concurrent runs, not a skill defect. Confirmed behaviors: App Service Plan fan-in preserved (1 EB env for N apps), Document Intelligence → Textract lands in `services_not_estimated[]`, Azure OpenAI gpt-4o → Bedrock Claude with honest cross-family caveats.
- **Gates green (both trees, by exit code):** frontmatter, cross-plugin drift (374 identical / 28 allowlisted), shared-vendored check, fixture asserters (13: 9 golden + 4 smoke), pricing-coverage.
[...]

PR #307, opened 2026-09-18
fix(bedrock): refresh model lifecycle statuses and adopt the 45-day Legacy policy

herosjourney:
## Problem

The Bedrock lifecycle table in both plugin trees is a snapshot dated **Aug 24, 2026**. Two things have gone stale since:

1. **Three models have passed EOL and three more crossed into the 90-day exclusion zone.** The plugin would still recommend, or price, models that are either already gone or will be unusable before a migration reaches production.
2. **AWS split the lifecycle policy on 2026-09-07.** Models launched on or after that date are governed by their model card and may get only a **45-day** Legacy period instead of six months, and they never appear in the pre-2026-09-07 Legacy table at all. Previously, a candidate's absence from the table was indistinguishable from "Active" — which is now an unsafe inference.

## Changes

Applied identically to `migrate/plugins/migration-to-aws/` and `advisor/plugins/aws-startup-advisor/`.

### `skills/gcp-to-aws/references/shared/ai-model-lifecycle.md`

Documents both policies and links the two authoritative pages (the original lifecycle page now covers only post-2026-09-07 models; pre-2026-09-07 models moved to `model-lifecycle-legacy.html`).

Table changes, recomputing `days_to_eol` as of 2026-09-17:

| Model | EOL | Was | Now |
| --- | --- | --- | --- |
| Claude 3 Haiku | 2026-09-10 | excluded | **Removed** (past EOL) |
| Nova Premier v1 | 2026-09-14 | excluded | **Removed** (past EOL) |
| Nova Sonic v1 | 2026-09-14 | excluded | **Removed** (past EOL) |
| Nova Canvas v1 | 2026-09-30 | excluded | excluded (13 days) |
| Nova Reel v1 | 2026-09-30 | excluded | excluded (13 days) |
| Claude Sonnet 4 | 2026-10-14 | excluded | excluded (27 days) |
| Jamba 1.5 Large / Mini | 2026-11-26 | legacy | **excluded** (70 days) |
| Marengo Embed v2.7 | 2026-11-30 | legacy | **excluded** (74 days) |
| Claude Opus 4.1 | 2027-01-08 | legacy | legacy (113 days) — the only row still outside the zone |

Two behavioral rules follow from this:

- The Nova Canvas carve-out that let it appear "as a Legacy fallback annotation" is dropped — it is inside the exclusion zone, so it must not appear at all.
- **Refresh Cadence** now requires a `GetFoundationModel` / `ListFoundationModels` check on `modelLifecycle.status` for any candidate absent from the table, and requires reporting the lifecycle as *unverified* when neither the API nor the model card is reachable, rather than inferring `active` from a pricing cache `Status` column.

### `skills/gcp-to-aws/references/shared/pricing-cache.md`

Restates the affected `Status` values, drops the EOL'd Nova Premier row from the web-grounding table, and adds a header note that the `Status` column is a dated snapshot rather than a substitute for the runtime lifecycle check.

Rates for EOL'd models are deliberately retained — they still explain the cost delta to a user already on one. `Last updated` is **not** bumped, because no prices were re-verified; only lifecycle statuses changed.

### `tools/model-id-lint.py`

Adds patterns for the three newly-EOL'd IDs so CI catches them as migration targets. Each appears only in the two catalog files today, which are allowlisted.

The other past-EOL IDs (Llama 3.2 / 3.1 405B, Command R / R+, Claude 3 Sonnet, Claude 3.5 Sonnet, Titan Image Generator v2) are deliberately **not** covered: `llm-to-bedrock` names them as migration *source* models in `resolve-source-model` and `run-source-model-baseline`, which is a valid use. Extending the lint to source-model references needs a source-vs-target distinction and belongs in a separate change.

## Verification

`mise run lint` and `mise run fmt:check` both pass, including:

- `lint:model-ids` — OK for both plugins
- `drift:check` — OK (281 identical, 27 allowlisted across 6 skill trees)
- `shared:check` — OK for both plugins
- `lint:md` — 0 errors across 896 files


migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/shared/ai-model-lifecycle.md
```diff
@@ -48,40 +56,40 @@ On each run, compute `days_to_eol = EOL date − today` for every model in the L
...
+- Titan Image Generator v2 (`amazon.titan-image-generator-v2:0`) — EOL 2026-06-30
+- Llama 3.2 all sizes (`meta.llama3-2-*-instruct-v1:0`) — EOL 2026-07-07
+- Llama 3.1 405B Instruct (`meta.llama3-1-405b-instruct-v1:0`) — EOL 2026-07-07
+- Claude 3 Sonnet (`anthropic.claude-3-sonnet-20240229-v1:0`) — EOL 2026-07-30
+- Claude 3.5 Sonnet v1 (`anthropic.claude-3-5-sonnet-20240620-v1:0`) — EOL 2026-07-30
+- Claude 3.5 Sonnet v2 (`anthropic.claude-3-5-sonnet-20241022-v2:0`) — EOL 2026-07-30
+- Command R / R+ (`cohere.command-r-v1:0` / `cohere.command-r-plus-v1:0`) — EOL 2026-08-19
+- Claude 3 Haiku (`anthropic.claude-3-haiku-20240307-v1:0`) — EOL 2026-09-10 (replacement: Claude Haiku 4.5)
+- Nova Premier v1 (`amazon.nova-premier-v1:0`) — EOL 2026-09-14 (replacement: Nova 2 Pro)
+- Nova Sonic v1 (`amazon.nova-sonic-v1:0`) — EOL 2026-09-14 (replacement: Nova 2 Sonic)
```
leon1418, line 86: [🤖 AI review 🤖] [P2] Update the lifecycle guard for the Removed section

Moving Nova Sonic (and the other two retired models) out of the pipe-delimited table leaves `agent-advisor/scripts/test_scoring.py::_legacy_or_excluded_rows()` unable to see them: it reads only table rows containing `legacy` or `excluded`. Consequently, `test_drift_mechanism_actually_fires_on_a_known_legacy_model` now fails in both plugin trees. I ran each complete Agent Advisor suite: the base `b30cb39b` passes 322 tests, while this head passes 321 and fails that test. An adversarial probe that adds a selectable pool key mapped to `nova-sonic-v1` also passes `test_no_pool_model_is_legacy_or_excluded` at this head, whereas the base rejects it. Please update the lifecycle matcher and its self-proof in both `test_scoring.py` copies to cover the new Removed representation, so retiring a model does not remove it from the guard. The green build does not run these Python suites.

leon1418 (changes requested):
[🤖 AI review 🤖]

Reviewed `a13a42beb1238662102a80addb83049e5c274353` against `b30cb39be02d2eae78cb8c482f4018435d4dc988`. Requesting changes for one P2 lifecycle-guard regression described inline.

The refreshed dates and launch-date/45-day policy match the current AWS lifecycle documentation. The three changed mirror pairs are byte-identical. Each full Agent Advisor Python suite reports 321 passed / 1 failed; both base suites pass all 322. Each scoped preflight/IAM/pricing suite passes 63 tests. Model-ID lint, cross-plugin drift, Markdown lint, formatting, and diff whitespace checks pass. Additional local probes reproduce the Removed-section guard gap and verify date arithmetic plus bare/profile lint and static-pricing behavior.

GitHub currently reports all nine checks successful; that result does not cover the failing Agent Advisor suites. No live AWS inference or deployment was performed.

migrate/plugins/migration-to-aws/skills/gcp-to-aws/references/shared/ai-model-lifecycle.md
herosjourney: Fixed and pushed at `856c6e81`, both trees.

Confirmed: `_legacy_or_excluded_rows()` only matched pipe-table rows containing "legacy"/"excluded", so a model that ages fully past EOL into the bulleted **Removed** list (like Nova Sonic v1 here) became invisible to the drift guard entirely — even though the file's own text calls Removed a stronger violation than "excluded".

Extended the matcher to also capture bulleted entries between the `**Removed` heading and the closing "AWS page lag" blockquote. Repointed the original table self-proof at Claude Opus 4.1 (still legacy in the table) and added a second, Removed-specific self-proof anchored on Nova Sonic v1, so each matching path (table row vs. Removed bullet) has its own proof it actually fires.

Reproduced your adversarial probe (a pool key mapped to a Removed model's id fragment) — confirmed it now gets rejected; it passed on the pre-fix code. Also checked the file's other bulleted lists (policy prose, exclusion-zone rules) aren't swallowed by the Removed-section scan. 323/323 tests passing in both trees (up from 322).

leon1418: [🤖 AI review 🤖] Verified fixed at `856c6e8101c00a3925445c92ea078bf9fda43265` in both plugin trees.

Both complete Agent Advisor suites now pass **323/323**. The matcher captures all 10 Removed entries while excluding all 14 unrelated bullets in the actual document; added before/after-section bullet controls are also ignored. The original adversarial pool cases for Nova Sonic, Nova Premier, and Claude 3 Haiku now reject, while the pre-fix matcher accepts them. Legacy Opus 4.1 and excluded Canvas/Jamba still reject, and the Active Sonnet control still passes. Disabling either the table scan or the Removed scan makes its corresponding self-proof fail.

The scoped preflight/IAM/pricing suites also pass **63/63 per tree**, and model-ID lint, drift, Markdown lint, formatting, and whitespace checks pass. This addresses the P2 finding; I found no remaining substantive issue in the current PR diff.

leon1418 (dismissed):

herosjourney:
## Recommendation: fold this fix into #304 and close this PR as superseded

Short version: **the 45-day-policy content here is correct and needs to land, but this PR edits a path that #304 is about to relocate** \u2014 so merging it on the current path would set up a silent clobber. The cleaner move is to port this content into #304 and close this.

**Why.** #304 introduces a canonical AI lifecycle file at `skills/shared/ai/ai-model-lifecycle.md` (vendored into each skill via `shared:sync`), and **deletes** the `gcp-to-aws/references/shared/ai-model-lifecycle.md` path this PR edits. Because the two PRs touch different paths, git raises **no merge conflict** \u2014 whichever lands second silently wins:\n- If #304 lands after this, #304's canonical file (which still carries the stale \u201cLegacy = minimum 6 months\u201d text, with no 45-day / 2026-09-07 policy) silently reappears on main and is fanned out byte-identically into 4 vendored copies, all blessed by `shared:check`.\n- If this lands after #304, its diff targets a path that no longer exists.\n\n**Timing makes the call easy.** #304 is ~7 days out; this EOL fix has a ~10-14 day window \u2014 so #304 carries the fix to main with slack, and there\u2019s no urgency reason to merge this early on the soon-to-be-obsolete path.\n\n**Plan:**\n1. Port this PR\u2019s corrected content \u2014 the 45-day / 2026-09-07 split policy, model-card governance, and the EOL\u2019d rows \u2014 into #304\u2019s canonical `skills/shared/ai/ai-model-lifecycle.md`, then `shared:sync` so the vendored copies inherit it. (#304 currently creates this file in both the advisor **and** migrate trees; if `shared:check` enforces cross-tree byte-identity, the port has to touch both to keep CI green, even though the migrate tree is being deprecated.)\n2. Verify #304\u2019s canonical file carries the 45-day policy \u2014 not the stale 6-month text \u2014 **before** #304 merges (tracked as a blocking checklist item on #304).\n3. Close this PR as superseded once #304\u2019s branch carries the content.\n\n**Do not close this PR until the port is committed on #304\u2019s branch** \u2014 closing early + #304 merging with its current stale file = the fix is lost silently, with CI defending the wrong content. The two actions must happen together.\n

leon1418 (commented):
[🤖 AI review 🤖]

Re-reviewed `d2473bb4418d588d94200d03ccb3e660c653e618` against `ec681ba5d49b039dbf1ef9d87d3304205b824f3d` using `startups-hybrid-v1`: OCR delegation, one fresh independent general reviewer, repository contract review, and the completed evidence gate. All eight changed files were covered, including the four Markdown files omitted by OCR selection. No remaining substantive findings.

The September 21 lifecycle snapshot and 45-day policy match the current primary AWS documentation. The bounded Removed-list guard still captures all ten entries and excludes unrelated bullets. Both Agent Advisor suites pass 323 tests; the scoped preflight/IAM/pricing suites pass 63 tests per tree. The independent reviewer also ran both 323-test suites. Lifecycle/model-ID probes, model-ID lint, drift, Markdown lint, formatting, and whitespace checks pass. GitHub reports nine successful checks; no live AWS invocation or deployment was performed.

The author's proposal to carry this change through #304 remains a separate landing decision. Before closing #307 as superseded, verify that the corrected policy and lifecycle guard are preserved in the replacement paths and consumers.

leon1418 (approved):

PR #309, opened 2026-09-18
feat(migrate): add plan.json writer for the plugin-to-web handoff

shrugninreddy-hash:
## What this adds

`scripts/emit-plan-json.py` — a writer that a **finished GCP or Heroku infra migration run** uses to emit a `plan.json` summary file. That file is what the AWS Startups Migrate **web import page** ingests so a customer can continue their migration on the web (view recommendations, cost estimates, and connect with AWS
account teams / partners).

Plus `tests/test_emit_plan_json.py` (unit tests).

## Requirements this satisfies

- **Copy validated values only — never generate.** Every field is copied from an already-validated run artifact or omitted; nothing is computed or invented.
- **Fail-open by design.** Any missing or unusable input leaves the migration successful, writes no file, prints the reason (`PLAN_SKIP | reason=…`), and stays re-runnable. A missed handoff must never cost a customer their result.
- **Infra-only for now.** Emits only pure-infra `INFRA_ONLY` runs. A run that also costed AI (`estimation-ai.json` present) is `FULL`/`AI_ONLY` and is skipped — see Follow-ups.
- **Strict-schema conformant.** Output field names, enum values, and types match the web import contract exactly (which rejects unknown keys / wrong types), and the file is written atomically (temp + rename) so a crash can't leave a partial.

## Fields emitted

`schemaVersion`, `sourcePlatform` (from `owning_skill`), `scope` (`INFRA_ONLY`), `runId` (from `.phase-status.json`), `producerVersion` (from the plugin manifest, optional), and `cost`: `awsMonthly` + `awsMonthlyBasis` (`BALANCED`), optional `sourceMonthly`, and `awsServiceItems[]` (per-service name + monthly cost from `projected_costs.breakdown`).

## Scope of this PR (please note)

This is **the writer only** — it is **not wired into any skill and prints no CTA link**, so nothing runs it during a customer migration yet. It ships dormant with **zero runtime effect** until a follow-up wires the invocation + the web CTA link. The split is deliberate: the CTA must not reach customers before the web import page is enabled.

## Follow-ups (separate PRs)

- Wire the invocation at run completion + print the web CTA link (ONLY after the web import page is enabled in PROD).
- OpenAI / LLM-to-Bedrock path (has no persisted cost artifact today).
- Richer optional fields (`expertConsultations`, `accuracy`, per-service `classification`/`category`) once the plugin persists them.

## Testing

- 35 unit tests: happy paths (GCP + Heroku), `breakdown` shape variance against real workshop fixtures, and every fail-open edge (missing/corrupt/wrong-type/ non-finite/non-UTF-8 input, AI-inclusive skip, stale-plan cleanup).
- `bandit` clean. Follows the existing root `scripts/` convention (stdlib-only, argparse, machine-readable status line) like`validate-migration-report.py`.


migrate/plugins/migration-to-aws/scripts/emit-plan-json.py
```diff
@@ -0,0 +1,305 @@
...
+        return []
+
+    items: list[dict] = []
+    for key, entry in breakdown.items():
+        # Drop the aggregate row (any casing) so the total is never shown as a service.
+        if not isinstance(key, str) or key.strip().lower() == "total" or not isinstance(entry, dict):
+            continue
+        monthly = entry.get("mid")
+        if not _is_amount(monthly):
+            continue
```
leon1418: [🤖 AI review 🤖] [P2] Preserve the core service rows from the GCP estimate contract

The GCP producer is instructed to write `references/shared/schema-estimate-infra.md`, whose `projected_costs.breakdown` stores the core Fargate, Aurora, S3 and networking figures under `monthly`, not `mid` (the same contract exists in both plugins). Running this writer against that exact JSON example returns `PLAN_OK` with `awsMonthly: 265` but no `awsServiceItems`; adding the required security/observability entries would retain those `mid` rows while still discarding the core services. This loses available per-service data on the supported GCP route. Handle the documented GCP shape with explicit scenario semantics, or align the producer contract first, and add a test using its actual example rather than only the Heroku fixture and synthetic `mid` entries.

migrate/plugins/migration-to-aws/scripts/emit-plan-json.py
```diff
@@ -0,0 +1,305 @@
...
+        },
+    }
+
+    source_monthly = _source_monthly(infra.get("current_costs"), source_platform)
+    if source_monthly is not None:
+        plan["cost"]["sourceMonthly"] = source_monthly
+
+    service_items = _service_items(projected)
+    if service_items:
+        plan["cost"]["awsServiceItems"] = service_items
```
leon1418, line 273: [🤖 AI review 🤖] [P2] Enforce the import limits before emitting service items

The strict web import contract permits at most 100 `awsServiceItems` and at most 128 characters in each `serviceName`, but this list is copied without either bound. I verified that a breakdown with 101 valid numeric rows, or one row whose label has 129 characters, produces `PLAN_OK` and an output outside that contract. The source estimate schema imposes neither limit, so passing its validation does not make these values importable; an optional breakdown can therefore make the entire handoff fail. Check these limits before attaching the optional field and omit unusable optional data or fail open with a reason, with boundary tests tied to the import contract.

migrate/plugins/migration-to-aws/scripts/emit-plan-json.py
```diff
@@ -0,0 +1,305 @@
...
+        print(f"PLAN_SKIP | reason=unreadable input: {err}")
+        return 0
+
+    # Write atomically: render to a sibling temp file, then rename into place. A
+    # crash/kill mid-write can then only leave the temp file, never a truncated
+    # plan.json the import might ingest.
+    tmp_path = out_path.parent / (out_path.name + ".tmp")
+    try:
+        tmp_path.write_text(json.dumps(plan, indent=2) + "\n", encoding="utf-8")
+        os.replace(tmp_path, out_path)
```
leon1418: [🤖 AI review 🤖] [P2] Give each invocation an exclusively created temporary file

The fixed `plan.json.tmp` path does not make overlapping reruns atomic. I reproduced this by pausing writer A halfway through its write, then letting writer B write a longer plan and rename that same temporary file into `plan.json`. When A resumes, its open descriptor overwrites the now-published file; A prints `PLAN_SKIP` because the temp path has gone, B has already printed `PLAN_OK`, and the final `plan.json` fails JSON parsing with extra data. A pre-existing temp symlink also lets `write_text` overwrite its target, including `estimation-infra.json`. Create a unique sibling temp file exclusively (for example with `tempfile.mkstemp`), replace from that path, and clean up only the file owned by that invocation.

leon1418 (changes requested):
[🤖 AI review 🤖]

Reviewed `55f522c2468e18ca72a23a7cbecf16fe0343c55c`. Requesting changes for three verified issues: the GCP producer shape loses core service items, optional service data can exceed the strict import bounds, and overlapping writers can corrupt the published JSON through the shared temporary path.

Validation: all 35 new tests passed locally; the cross-plugin drift check passed (281 identical, 27 allowlisted); `git diff --check` passed. Additional probes used the exact GCP schema example from both plugins, exercised import boundaries, and reproduced concurrent-write corruption. All 13 currently reported GitHub checks are successful. The writer-only scope and deferred skill/CTA wiring are intentional; no production import was exercised.

migrate/plugins/migration-to-aws/scripts/emit-plan-json.py
leon1418: [🤖 AI review 🤖] Verified fixed at `e682e1c306f7be39fd356e363075f2efee177bdd`. Replayed the exact GCP estimate example from both plugin trees: Fargate (71), Aurora PostgreSQL (269), S3 (86), and networking (53) now survive in `awsServiceItems`, while the selected Balanced headline remains 265. A mixed breakdown also preserves a security-baseline `mid` row alongside the core `monthly` rows; nested alternatives are not emitted. This resolves my service-row omission finding.

migrate/plugins/migration-to-aws/scripts/emit-plan-json.py
leon1418: [🤖 AI review 🤖] Verified fixed at `e682e1c306f7be39fd356e363075f2efee177bdd` against the current import bounds. Source-schema-valid inputs retain exactly 100 service items and omit the optional list at 101. Names at 128 UTF-16 units are retained; 129 ASCII units or 65 astral characters are dropped. I also verified the USD cap boundaries and that an invalid primary amount does not fall back to a different figure. This resolves my import-limit finding.

migrate/plugins/migration-to-aws/scripts/emit-plan-json.py
leon1418: [🤖 AI review 🤖] Verified fixed at `e682e1c306f7be39fd356e363075f2efee177bdd`. Replayed the controlled overlapping-write case: writer A paused after writing half its unique temp, writer B published a complete longer plan, then A resumed and published a complete plan. Both returned `PLAN_OK`; the final JSON remained valid and no write temps remained. A pre-existing `plan.json.tmp` symlink left the estimate artifact untouched, and an injected replace failure preserved the previous plan while cleaning up only the new temp. This resolves my shared-temp corruption finding.

leon1418 (commented):
[🤖 AI review 🤖]

Follow-up review of `e682e1c306f7be39fd356e363075f2efee177bdd`: all three findings from my previous review are verified fixed, and I found no remaining substantive issue in the current writer-only diff. I recommend approval for the user to submit.

Validation: 46 writer tests passed locally; 19 additional probes passed, covering the exact GCP examples from both plugin trees, import boundaries, concurrent writes, legacy-temp symlink preservation, and replace-failure recovery. The cross-plugin drift check and `git diff --check` passed. All 13 current GitHub checks are successful. Downstream compatibility was checked against current source contracts; no production import was performed.

This is a COMMENT review. The earlier CHANGES_REQUESTED review has not been dismissed or superseded, and the existing threads remain for resolution before merge.

leon1418 (approved):

